PR #28, opened 2019-10-09
Add the integration tests in the parsec repository

hug-dev:
As the integration tests really test the PARSEC service functionalities
it makes more sense to put them here and use a dependency on
parsec-client-test.
It also makes the CI script easier as there is no need to clone the Test
Client somewhere now.


hug-dev:
I will remove them from the test client repo when this gets merged 👍

PR #30, opened 2019-10-10
Use thread pool instead of new thread per request

### LINKED ISSUE #29: Implement a thread pool
Currently all requests generate a new thread - this can turn out to be very inefficient if a large number of requests hits the server at the same time, constantly. A better option would be to implement a thread pool and use it to allocate requests.

An idiomatic implementation is suggested in [The Rust Book](https://doc.rust-lang.org/book/ch20-00-final-project-a-web-server.html). 

This change will be tested as part of the stress tests, when those will be implemented.

ionut-arm:
Added a threadpool to handle our requests instead of spinning up a new thread per request which would allow a DoS.

Resolves #29.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -382,6 +382,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "num_cpus"
+version = "1.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "libc 0.2.62 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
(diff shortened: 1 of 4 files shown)


ionut-arm:
> It's so good how easy it is 😃
> I guess in the future the thread pool size will be set in the configuration?

Yes, I'd think so, the admin can judge by themselves and maybe we should stick with `ncpus` for the default value?

ionut-arm:
`threadpool` actually uses that crate under the hood and the Builder uses the number of CPUs as a default if you don't specify a value :)

PR #34, opened 2019-10-10
Add a GitHub Actions workflow for CI

hug-dev:
The workflow will perform all the checks described in the all.sh script
which has been moved to `tests`.


hug-dev:
I did not add the logic that was in the `Dockerfile` about downloading and compiling Mbed Crypto. We can see if that takes too much time or not. If it does we can still execute this workflow in the `Dockerfile` that we had.

hug-dev:
Would that be even faster?

Issue #35, opened 2019-10-10
Make PARSEC a daemon


hug-dev:
PARSEC should run as a daemon, running fully as a background process. Its logs should go at a convenient place and not on the standard output.
It should be possible to gracefully terminate it and wait for all of its threads to finish.

justincormack:
logging to stdout is the normal way for daemons to run now; systemd or whatever will handle redirecting logs to wherever, so it is not necessary to change that. We should have a clean shutdown.

hug-dev:
Agree to both points.
I have written some research I made about the subject and what I think we should be doing:

There seems to be two kinds of daemons:
* "traditional" SysV daemons which contain a double-forking mechanism to separate themselves from their parent process and the TTY they are linked to. They redirect themselves their standard file descriptors to a file or system wide logging. The crate [`daemonize`](https://docs.rs/daemonize/0.4.1/daemonize/) works like that.
* "modern" systemd daemons where systemd itself manages the daemons which do not have to do anything special apart from having a graceful shutdown mechanism. This can be done by **implementing a `SIGTERM` signal handler** (signal that is sent to the daemon when systemd wants to close it.) systemd needs a service file that describe the service type and its logging mechanism. Logs can be redirected easily so that they can be viewed with the `journalctl`, something like `journalctl -u parsecd.service`. There are also advanced features of systemd to automatically start-up the daemon whenever a connection is received on the socket, and not have it running all the time.

I am mostly inclined to follow the modern, systemd, process of daemonization and I see for that the following tasks:
1. Create a `.service` file and put it at the root of this repo (same way as [`containerd`](https://github.com/containerd/containerd/blob/master/containerd.service) does)
2. Implement a handler of the `SIGTERM` signal to cleanly shutdown PARSEC (send to all thread a message through a channel for them to return and `join` them)
3. Adds CI tests where PARSEC is run as a daemon, with `systemctl` commands.

Separate but linked tasks:
* With #36, replace all `println` calls with a real logging system (`log` + `env_logger` for example). (Possibly add log tests?)
* Investigate socket activation of PARSEC

**Note:** systemd is a Linux only daemon manager, which is not on MacOS. The service will need to have OS dependant (`#[cfg]`?) code and still work by just running `cargo run` on other OSes. We will need to investigate daemon manager on the other OSes we want PARSEC to run on.

[...]

PR #39, opened 2019-10-14
Add SIGTERM handler for a graceful shutdown

hug-dev:
Changes the wait_for_connection function in the Listen trait to be
non-blocking. That allows checking in the main loop if the SIGTERM
signal has been raised.

Testing:
```shell
[hugdev@localhost parsec]$ ./target/debug/parsec &
[1] 11951
[hugdev@localhost parsec]$ kill 11951
[hugdev@localhost parsec]$ SIGTERM signal received.
Shutting down PARSEC, waiting for all threads to finish.

[1]+  Done                    ./target/debug/parsec

```

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -25,9 +25,14 @@ use parsec::providers::{
 use parsec_interface::operations_protobuf::ProtobufConverter;
 use parsec_interface::requests::AuthType;
 use parsec_interface::requests::{BodyType, ProviderID};
+use signal_hook::flag;
+use signal_hook::SIGTERM;
+use std::io::Error;
 use std::path::PathBuf;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::sync::RwLock;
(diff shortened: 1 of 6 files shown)


hug-dev:
Implement point 2 of #35

src/bin/main.rs
```diff
@@ -91,23 +96,38 @@ fn build_components() -> (FrontEndHandler, impl Listen) {
...
             let front_end_handler = front_end_handler.clone();
             threadpool.execute(move || {
                 front_end_handler.handle_request(stream);
             });
-        } else {
-            println!("Error on establishing last connection, continuing...");
         }
+
+        //TODO: this value should come from the configuration.
+        thread::sleep(Duration::from_millis(10));
```
ionut-arm: You can probably put this in an else branch above to "simulate" a timeout on the socket, so that you only sleep if there's no one waiting for a connection.

hug-dev: Good idea!

PR #41, opened 2019-10-15
Add locking around key handle operations in mbed provider

### LINKED ISSUE #40: Key handle manipulation is not thread-safe in Mbed Crypto
As mentioned [here](https://github.com/ARMmbed/mbed-crypto/issues/266), Mbed Crypto is not yet thread safe when it comes to allocating or releasing key slots.
This came out of our own stress tests as well, with the service failing to handle requests or tripping over in a segmentation fault. The underlying problem was that all the threads processing requests were getting the same key slot allocated (number 32) and were modifying each other's keys. 

This issue should be addressed by hiding the calls to create, open or close a key behind a mutex.

ionut-arm:
Add a mutex around create, open and close key operations.
Add a semaphore to wrap all function call that use keys.

Fixes #40

### DIFF
--- a/src/providers/mbed_provider/mod.rs
+++ b/src/providers/mbed_provider/mod.rs
@@ -17,7 +17,7 @@ use crate::authenticators::ApplicationName;
 use crate::key_id_managers::{KeyTriple, ManageKeyIDs};
 use std::collections::HashSet;
 use std::convert::TryInto;
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Mutex, RwLock};
 
 use parsec_interface::operations::key_attributes::KeyLifetime;
 use parsec_interface::operations::ProviderInfo;
@@ -29,6 +29,7 @@ use parsec_interface::operations::{OpExportPublicKey, ResultExportPublicKey};
 use parsec_interface::operations::{OpImportKey, ResultImportKey};
 use parsec_interface::operations::{OpListOpcodes, ResultListOpcodes};
 use parsec_interface::requests::{Opcode, ProviderID, ResponseStatus, Result};
+use std_semaphore::Semaphore;
 use uuid::Uuid;
 
 #[allow(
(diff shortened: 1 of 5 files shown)


src/providers/mbed_provider/mod.rs
```diff
@@ -56,13 +57,17 @@ const SUPPORTED_OPCODES: [Opcode; 7] = [
     Opcode::ListOpcodes,
 ];
 
+const KEY_SLOT_COUNT: isize = 32;
```
hug-dev: Is there a way to read that value from the FFI?

src/providers/mbed_provider/mod.rs
```diff
@@ -56,13 +57,17 @@ const SUPPORTED_OPCODES: [Opcode; 7] = [
...
+
 pub struct MbedProvider {
     // When calling write on a reference of key_id_store, a type
     // std::sync::RwLockWriteGuard<dyn ManageKeyIDs + Send + Sync> is returned. We need to use the
     // dereference operator (*) to access the inner type dyn ManageKeyIDs + Send + Sync and then
     // reference it to match with the method prototypes.
     key_id_store: Arc<RwLock<dyn ManageKeyIDs + Send + Sync>>,
     local_ids: RwLock<LocalIdStore>,
+    key_handle_mutex: Mutex<()>,
```
hug-dev: Maybe it would be nice to add a comment on top of those lines to explain why the Mutex, Semaphore are needed. Just as an help for future code readers.

src/providers/mbed_provider/mod.rs
hug-dev: or maybe put it in the constants file instead?

hug-dev:
Should we merge the Stress Test first and then this one to show that it fixes the bug?

src/providers/mbed_provider/mod.rs
ionut-arm: Not through FFI, as it is not directly exposed in the header files, but I added it in the constants file.

src/providers/mbed_provider/mod.rs
ionut-arm: Done

PR #44, opened 2019-10-17
Add service builder and configuration

ionut-arm:
Adding configuration capability for the service.

See config.toml for details about currently supported params.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,0 +1,26 @@
+# PARSEC Service Configuration File
+
+# (Required) Core settings apply to the service as a whole rather than to individual components within it.
+[core_settings]
+# Size of the thread pool used for processing requests. Defaults to the number of processors on 
+# the machine.
+#thread_pool_size = 8
+
+# Duration of sleep when the connection pool is empty. This can limit the response
+# times for requests and so should be set to a low number. Default value is 10.
+#idle_listener_sleep_duration = 10 # in milliseconds
+
+# (Required) Configuration for the service IPC listener component.
+[listener]
+# (Required) Defines the type of IPC that the service will support. 
+listener_type = "DomainSocket"
+# (Required) Defines the timeout of the read and write operations on the IPC channel. After the
+# timeout expires, the connection is dropped.
+timeout = 200 # in milliseconds
+
+# (Required) Configuration for the component managing key IDs for providers.
+[key_manager]
+# (Required) Defines the type of key ID manager to be used.
+manager_type = "OnDisk"
+# Defines the path to the directory on disk where the mapping will be persisted. 
+#store_path = "./mappings"
--- a/src/key_id_managers/mod.rs
+++ b/src/key_id_managers/mod.rs
@@ -20,9 +20,21 @@
 
 use crate::authenticators::ApplicationName;
 use parsec_interface::requests::ProviderID;
+use serde::Deserialize;
 use std::fmt;
 
 pub mod on_disk_manager;
+
+#[derive(Deserialize)]
+pub enum KeyIdManagerType {
+    OnDisk,
+}
+
+#[derive(Deserialize)]
+pub struct KeyIdManagerConfig {
+    pub manager_type: KeyIdManagerType,
+    pub store_path: Option<String>,
+}
 
 /// This structure corresponds to a unique identifier of the key. It is used internally by the Key
 /// ID manager to refer to a key.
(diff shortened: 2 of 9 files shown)


src/utils/service_builder.rs
```diff
@@ -0,0 +1,136 @@
...
+pub struct ServiceConfig {
+    pub core_settings: CoreSettings,
+    pub listener: ListenerConfig,
+    pub key_manager: KeyIdManagerConfig,
+}
+
+pub struct ServiceBuilder;
+
+impl ServiceBuilder {
+    pub fn build_service(config: &ServiceConfig) -> FrontEndHandler {
```
hug-dev: Should it be `build_front_end_handler` instead of `build_service`?

src/utils/service_builder.rs
```diff
@@ -0,0 +1,136 @@
...
+
+        let simple_authenticator = Box::from(SimpleAuthenticator {});
+
+        FrontEndHandlerBuilder::new()
+            .with_dispatcher(dispatcher)
+            .with_authenticator(AuthType::Simple, simple_authenticator)
+            .build()
+    }
+
+    pub fn start_listener(config: &ListenerConfig) -> Box<dyn Listen> {
```
hug-dev: Why not `build_listener` 😛 ? Would be nice just for coherency with the other functions.

src/utils/service_builder.rs
hug-dev: I understand that it does build almost everything, but is it the whole service though?

config.toml
```diff
@@ -0,0 +1,26 @@
...
+# (Required) Defines the timeout of the read and write operations on the IPC channel. After the
+# timeout expires, the connection is dropped.
+timeout = 200
+
+# (Required) Configuration for the component managing key IDs for providers.
+[key_manager]
+# (Required) Defines the type of key ID manager to be used.
+manager_type = "OnDisk"
+# Defines the path to the directory on disk where the mapping will be persisted. 
+#store_path = "./mappings"
```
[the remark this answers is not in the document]
ionut-arm: `store_path` is an option in `KeyIdManagerConfig`
We do at the moment, this config will be expanded with other functionality, but by allowing more than one you'd have to also specify for each provider which one it uses and so on. I'd rather keep this a smaller change and limit it to one, it's easy to expand.

src/key_id_managers/mod.rs
```diff
@@ -20,10 +20,22 @@
...
 
+#[derive(Deserialize)]
+pub enum KeyIdManagerType {
+    OnDisk,
+}
+
+#[derive(Deserialize)]
+pub struct KeyIdManagerConfig {
+    pub manager_type: KeyIdManagerType,
+    pub store_path: Option<String>,
```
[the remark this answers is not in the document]
ionut-arm: I was thinking about that as well - I think so, unless it's an in-memory store in which case you can leave it empty. Otherwise, you have to store them _somewhere_

src/utils/service_builder.rs
ionut-arm: It felt more natural - you build a service, but you start listening (instead of build listening :P )

src/utils/service_builder.rs
ionut-arm: Well... it is, all the service components (maybe excluding the listener, if you want to consider that part of the service) are created here, so `build_front_end_handler` is way more misleading than `build_service`

config.toml
```diff
@@ -0,0 +1,26 @@
...
+# times for requests and so should be set to a low number. Default value is 10.
+#idle_listener_sleep_duration = 10
+
+# (Required) Configuration for the service IPC listener component.
+[listener]
+# (Required) Defines the type of IPC that the service will support. 
+listener_type = "DomainSocket"
+# (Required) Defines the timeout of the read and write operations on the IPC channel. After the
+# timeout expires, the connection is dropped.
+timeout = 200
```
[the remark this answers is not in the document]
ionut-arm: Agree, I specified it for the other one but forgot to update here

src/key_id_managers/mod.rs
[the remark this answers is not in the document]
hug-dev: We can leave that for now and expand later if needed!

src/utils/service_builder.rs
ionut-arm: (and as the configuration will evolve, I think almost every component will get its own `build` function that will be private to this module)

PR #46, opened 2019-10-17
Add a compile-time option for a daemon binary

### LINKED ISSUE #35: Make PARSEC a daemon

hug-dev:
The "systemd-daemon" feature will compile PARSEC to be used as a
systemd, socket activated daemon.
Add systemd unit files and instruction on how to install the service.

Closes #35 

A `Dockerfile` could be added later to install the daemon from a local checkout of `parsec` inside a container and only expose the socket to the host.

@ionut-arm FYI, the systemd unit files contain two hardcoded paths:
* `/home/parsec/.cargo/bin/parsec` for the binary
* `/home/parsec/parsec.sock` for the socket

### DIFF
--- a/src/front/domain_socket.rs
+++ b/src/front/domain_socket.rs
@@ -22,6 +22,7 @@ use super::listener;
 
 use listener::Listen;
 use listener::ReadWrite;
+use std::os::unix::io::FromRawFd;
 
 static SOCKET_PATH: &str = "/tmp/security-daemon-socket";
 
@@ -43,23 +44,35 @@ impl DomainSocketListener {
     /// fails
     /// - if binding to the socket path fails
     fn init(&mut self) {
-        let socket = Path::new(SOCKET_PATH);
+        if cfg!(feature = "systemd-daemon") {
+            // The PARSEC service is socket activated (see parsec.socket file).
+            // systemd creates the PARSEC service giving it an initialised socket as the file
+            // descriptor number 3 (see sd_listen_fds(3) man page).
+            // If an instance of PARSEC compiled with the "systemd-daemon" feature is run directly
+            // instead of by systemd, this call will still work but the next accept call on the
+            // UnixListener will generate a Linux error 9 (Bad file number), as checked below.
+            unsafe {
+                self.listener = Some(UnixListener::from_raw_fd(3));
+            }
+        } else {
+            let socket = Path::new(SOCKET_PATH);
 
-        if socket.exists() {
-            fs::remove_file(&socket).unwrap();
+            if socket.exists() {
+                fs::remove_file(&socket).unwrap();
+            }
+
+            let listener_val = match UnixListener::bind(SOCKET_PATH) {
+                Ok(listener) => listener,
+                Err(err) => panic!(err),
+            };
+
+            // Set the socket as non-blocking.
+            listener_val
+                .set_nonblocking(true)
+                .expect("Could not set the socket as non-blocking");
+
+            self.listener = Some(listener_val);
         }
-
-        let listener_val = match UnixListener::bind(SOCKET_PATH) {
-            Ok(listener) => listener,
-            Err(err) => panic!(err),
-        };
-
-        // Set the socket as non-blocking.
-        listener_val
-            .set_nonblocking(true)
-            .expect("Could not set the socket as non-blocking");
-
-        self.listener = Some(listener_val);
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/front/domain_socket.rs
```diff
@@ -84,6 +97,16 @@ impl Listen for DomainSocketListener {
                     }
                 }
                 Err(err) => {
+                    if cfg!(feature = "systemd-daemon") {
+                        // When run as a systemd daemon, a file descriptor mapping to the Domain Socket
+                        // should have been passed to this process.
+                        if let Some(os_error) = err.raw_os_error() {
+                            // On Linux, 9 is EBADF (Bad file number)
```
ionut-arm: Is it reasonable that we'll reach this point? if not => `unreachable!();`

lnicola: It's also possible that the user wants to run the application as a service instead of socket activation. TBH, using fd 3 directly feels a bit dirty.

According to http://0pointer.de/blog/projects/socket-activation.html, `sd_listen_fds` would be a better solution, but my `sd-notify` crate doesn't support that (yet?).

hug-dev: Hi @lnicola!
Currently we only have the compile-time flag `systemd-daemon` to differentiate between a binary to be launched by systemd or by the command line. If compiled as a daemon and launched on the command line by the user, it will trigger this error.

I agree that a better way would be to do it at runtime, the way shown in your link below and assuming that we can also do runtime checks on other daemon-dependant parts of our application and remove the compile-time flag.
For example, what would be the effect of `sd-notify` if not run as a daemon?
The crate `systemd` has Rust wrappers around this function and other very useful functions that would allow us to do that. It is licensed under LGPL-2.1+ so we would have to check for compatibility with our Apache 2.0 license.

lnicola: > Currently we only have the compile-time flag systemd-daemon to differentiate between a binary to be launched by systemd or by the command line. If compiled as a daemon and launched on the command line by the user, it will trigger this error.

That shouldn't be needed in general, IMHO.

> For example, what would be the effect of sd-notify if not run as a daemon?

Do you mean `sd_notify::notify`? I haven't checked the implementation of `sd_listen_fds`, but readiness notification works like this:

 - if `systemd` or a compatible service manager starts the daemon, it sets an well-known environment variable (`NOTIFY_SOCKET`) with a Unix Domain Socket address
 - applications can check whether the variable is set, and can send messages and control `systemd` by writing to that socket
 - applications should unset the variable before spawning child processes

In the case of `notify`, if the program is not launched by `systemd`, the variable won't be set, and the function will return `Ok(())`, so the whole thing is a no-op.

> The crate systemd has Rust wrappers around this function and other very useful functions that would allow us to do that. It is licensed under LGPL-2.1+ so we would have to check for compatibility with our Apache 2.0 license.

I don't think they're compatible. I can look into adding more functionality to the `sd-notify` crate. If you want an alternative to `sd-notify`, there's https://github.com/lucab/libsystemd-rs, which is licensed under MIT/Apache 2.0. The reason I started `sd-notify` is https://github.com/lucab/libsystemd-rs/issues/24.

hug-dev: > That shouldn't be needed in general, IMHO.

I agree now with this philosophy to make one binary that works both when run as an user or by `systemd` but something I forgot as well: ideally (for now) we would like `parsec` to work on as many OSes as possible. The compile-time flag is a way to make it work also on OSes that do not have `systemd`. And just having conditional compilation and target Linux only might not be enough if some Linux distributions do not have `systemd`?

> In the case of notify, if the program is not launched by systemd, the variable won't be set, and the function will return Ok(()), so the whole thing is a no-op.

Thanks for the explanation, that is very good to know!

> I can look into adding more functionality to the sd-notify crate. If you want an alternative to sd-notify, there's https://github.com/lucab/libsystemd-rs, which is licensed under MIT/Apache 2.0. The reason I started sd-notify is lucab/libsystemd-rs#24.

Even if we end up keeping the compile-time flag, it would be more robust and less dirty to call `sd_listen_fds` as done on the blog post. Like this, even if the flag is set the binary can be launched both ways.
I checked the docs of `libsystemd-rs` and it does not seem to have the function.
If you were to add a `sd_listen_fds` function to your crate, we would use it for sure 😋 !!

lnicola: > And just having conditional compilation and target Linux only might not be enough if some Linux distributions do not have systemd?

It's totally fine to run such a binary on a non-`systemd` distro -- unless you're linking to the C `libsystemd`, which might prove problematic (and you also want to avoid it because of the licensing issue?).

> If you were to add a sd_listen_fds function to your crate, we would use it for sure yum !!

No promises, but I'll take a look to see what it would imply. I didn't implement FD passing in `sd-notify` because it's quite complex, and it's probably using the same mechanism.

hug-dev: > It's totally fine to run such a binary on a non-systemd distro -- unless you're linking to the C libsystemd, which might prove problematic (and you also want to avoid it because of the licensing issue?).

Oh, just got it, I did not know you could implement the function in Rust only without any `systemd` dependency, as you did for `notify`. In that case, the only conditional compilation checks would be on the OS which is good!

> No promises, but I'll take a look to see what it would imply. I didn't implement FD passing in sd-notify because it's quite complex, and it's probably using the same mechanism.

Great, please ping me if you ever do 😃 !

Issue #36, opened 2019-10-11
Improve logging message structure


ionut-arm:
Currently our log messages specify the error that occurred, however it would be useful to also include the location in code. 
Example:
Currently: "Failed to read request; status: {}" 
Enhanced: "FrontEndHandler handle_request: Failed to read request; status: {}"

Alternatively, when implementing actual logging, this might be sorted out for us by the logging framework (?)

hug-dev:
`env_logger` put in the code at #48 has the following output:
```
[2019-10-18T15:39:51Z INFO  parsec::providers::mbed_provider] Mbed Provider - Destroy Key
```
Is that what you had in mind?

ionut-arm:
> `env_logger` put in the code at #48 has the following output:
> 
> ```
> [2019-10-18T15:39:51Z INFO  parsec::providers::mbed_provider] Mbed Provider - Destroy Key
> ```
> 
> Is that what you had in mind?

Yes! Though I'm not sure it'd be that helpful to admins, but better than nothing

PR #48, opened 2019-10-18
Replace println calls with log crate

hug-dev:
Pulls in new changes from test client that already contains the switch
to log commit.
Adds an INFO log level in the test script, in the CI and in the systemd
unit file.


.github/workflows/ci.yml
```diff
@@ -38,7 +38,7 @@ jobs:
     - uses: actions/checkout@v1
     - run: cargo build --verbose
     - run: |
-        cargo run &
+        RUST_LOG=info cargo run &
```
ionut-arm: Could you also add `RUST_BACKTRACE=1`?

hug-dev: Of course! That is, Ionut, an excellent suggestion.

hug-dev:
Closes #36

PR #51, opened 2019-10-21
Add provider configuration

ionut-arm:
Add ability to configure what providers are initialised and with what key ID managers; add logging-related configuration

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ sd-notify = { version = "0.1.1" }
 toml = "0.4.2"
 serde = { version = "1.0", features = ["derive"] }
 env_logger = "0.7.1"
-log = "0.4.8"
+log = { version = "0.4.8", features = ["serde"] }
 
 [dev-dependencies]
 parsec-client-test = { git = "https://github.com/parallaxsecond/parsec-client-test", tag = "0.1.3" }
@@ -40,6 +40,3 @@ mbed-crypto-version = "mbedcrypto-1.1.0"
 default = ["mbed"]
 mbed = []
 
-# Feature to compile the PARSEC binary to be executed as a systemd daemon.
-# This feature is only available on Linux.
-systemd-daemon = []
--- a/config.toml
+++ b/config.toml
@@ -2,7 +2,7 @@
 
 # (Required) Core settings apply to the service as a whole rather than to individual components within it.
 [core_settings]
-# Size of the thread pool used for processing requests. Defaults to the number of processors on 
+# Size of the thread pool used for processing requests. Defaults to the number of processors on
 # the machine.
 #thread_pool_size = 8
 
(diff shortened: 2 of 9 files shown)


src/back/dispatcher.rs
```diff
@@ -81,6 +81,25 @@ impl DispatcherBuilder {
...
+                for (provider_id, backend_handler) in new_backends.drain() {
+                    backends.insert(provider_id, backend_handler);
+                }
+            }
+            None => {
+                let mut map = HashMap::new();
+                for (provider_id, backend_handler) in new_backends.drain() {
+                    map.insert(provider_id, backend_handler);
+                }
+                self.backends = Some(map);
```
hug-dev: Could it just be:
```rust
self.backends = Some(new_backends);
```
?

src/back/dispatcher.rs
```diff
@@ -81,6 +81,25 @@ impl DispatcherBuilder {
         self
     }
 
+    pub fn with_backends(mut self, mut new_backends: HashMap<ProviderID, BackEndHandler>) -> Self {
+        match &mut self.backends {
+            Some(backends) => {
+                for (provider_id, backend_handler) in new_backends.drain() {
+                    backends.insert(provider_id, backend_handler);
+                }
```
hug-dev: Could you use:
```rust
backends.extend(new_backends);
```
instead?

src/bin/main.rs
```diff
@@ -86,3 +81,22 @@ fn main() -> Result<(), Error> {
...
+fn log_setup(config: &ServiceConfig) {
+    let mut env_log_builder = env_logger::builder();
+
+    if let Some(level) = config.core_settings.log_level {
+        env_log_builder.filter_level(level);
+    }
+    if cfg!(feature = "systemd-daemon") {
+        // Logs viewed with journalctl already contain timestamps.
+        env_log_builder.format_timestamp(None);
+    } else {
```
hug-dev: I think you can just remove this case and only keep the `else` part, that way we can completely remove this flag 😃

hug-dev: and then we can also remove it from `Cargo.toml`

src/back/dispatcher.rs
hug-dev: Maybe there is some way to call `extend` on an `Option` to handle both cases

config.toml
```diff
@@ -2,25 +2,47 @@
...
+# (Required) Timeout of the read and write operations on the IPC channel. After the
 # timeout expires, the connection is dropped.
 timeout = 200 # in milliseconds
 
-# (Required) Configuration for the component managing key IDs for providers.
-[key_manager]
-# (Required) Defines the type of key ID manager to be used.
+# (Required) Configuration for the components managing key IDs for providers.
+# Defined as an array of tables: https://github.com/toml-lang/toml#user-content-array-of-tables
+[[key_manager]]
```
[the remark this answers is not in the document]
ionut-arm: Was thinking about that. The thing is that despite this being an array, the contents of the array are added as: 

```
[[key_manager]]
...first key manager...

[[key_manager]]
...second key manager...
```

So in a way it looks more sensible to keep the name as singular, since each "header" identifies only one entry.

src/back/dispatcher.rs
ionut-arm: Done

src/bin/main.rs
ionut-arm: Done

src/back/dispatcher.rs
ionut-arm: See update

src/back/dispatcher.rs
```diff
@@ -67,16 +67,17 @@ impl DispatcherBuilder {
...
-            Some(backends) => {
-                backends.insert(provider_id, backend_handler);
-            }
-            None => {
-                let mut map = HashMap::new();
-                map.insert(provider_id, backend_handler);
-                self.backends = Some(map);
-            }
-        }
+        let mut backends = self.backends.unwrap_or_default();
```
hug-dev: `unwrap_or_default` is a good idea!!

Cargo.toml
```diff
@@ -40,6 +40,3 @@ mbed-crypto-version = "mbedcrypto-1.1.0"
 default = ["mbed"]
 mbed = []
 
-# Feature to compile the PARSEC binary to be executed as a systemd daemon.
-# This feature is only available on Linux.
-systemd-daemon = []
```
hug-dev: 🎊

Cargo.toml
```diff
@@ -21,7 +21,7 @@ sd-notify = { version = "0.1.1" }
 toml = "0.4.2"
 serde = { version = "1.0", features = ["derive"] }
 env_logger = "0.7.1"
-log = "0.4.8"
+log = { version = "0.4.8", features = ["serde"] }
```
hug-dev: Just for my curiosity: why is the `serde` feature needed here?

ionut-arm: https://github.com/parallaxsecond/parsec/pull/51/files#diff-5c0f3ed7e332f3aead779609ed270a50R53

The config key that we're deserializing from the config file is essentially a log level filter (which has to have the serde feature enabled for that)

PR #52, opened 2019-10-22
Remove references to key lifetime

hug-dev:
Changes needed after they were removed from
parallaxsecond/parsec-interface-rs#5.

Will need changes in `Cargo.toml` file to pull the newer version of `parsec-interface-rs`.
@ionut-arm Should it be `0.2.0` as well?


ionut-arm:
> Should it be `0.2.0` as well?

I think it makes sense to keep this consistently versioned so yeah, `0.2.0` sounds good

src/providers/mbed_provider/constants.rs
```diff
@@ -157,7 +157,6 @@ pub const PSA_ALG_KEY_DERIVATION_MASK: psa_algorithm_t = 0x010f_ffff;
 pub const PSA_ALG_SELECT_RAW: psa_algorithm_t = 0x3100_0001;
 pub const PSA_ALG_FFDH_BASE: psa_algorithm_t = 0x2210_0000;
 pub const PSA_ALG_ECDH_BASE: psa_algorithm_t = 0x2220_0000;
-pub const PSA_KEY_LIFETIME_VOLATILE: psa_key_lifetime_t = 0x0000_0000;
```
ionut-arm: Just as a sidenote, this could've been kept - definitely not all the constants defined here are actually used; maybe in the future we'll end up supporting volatile keys somehow, but this can go as is.

hug-dev: I'll revert that.

PR #56, opened 2019-10-24
Upgrade dependency on Mbed Crypto to v2.0.0

### LINKED ISSUE #38: Update to Mbed Crypto v2.0.0
We're currently using Mbed Crypto 1.1.0 which exposed an old version of PSA Crypto API. 

After the [interface work](https://github.com/parallaxsecond/parsec-interface-rs/issues/4) is done, we should update the version of mbed we pull to [2.0.0](https://github.com/ARMmbed/mbed-crypto/releases/tag/mbedcrypto-2.0.0) and modify the way we create and handle keys (where necessary).

Changes that will probably be needed:
* Updating the way Mbed Provider handles key creation (as per the [new PSA spec](https://github.com/ARMmbed/mbed-crypto/raw/psa-crypto-api/docs/PSA_Cryptography_API_Specification.pdf))
* Updating the way we deal with key lifetimes (deciding how this will be handled is WIP)
* Adding `permit_copy` as a flag for key attributes

If providers will be forced to only support one key lifetime, a new issue shall be raised.

ionut-arm:
Bumping dependency version of Mbed Crypto to [2.0.0](https://github.com/ARMmbed/mbed-crypto/releases/tag/mbedcrypto-2.0.0). Also making all changes required to keep the service functional. This included a change in the client library, bumping that to `0.1.5`.

This resolves #38

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ env_logger = "0.7.1"
 log = { version = "0.4.8", features = ["serde"] }
 
 [dev-dependencies]
-parsec-client-test = { git = "https://github.com/parallaxsecond/parsec-client-test", tag = "0.1.4" }
+parsec-client-test = { git = "https://github.com/parallaxsecond/parsec-client-test", tag = "0.1.5" }
 num_cpus = "1.10.1"
 
 [build-dependencies]
@@ -34,7 +34,7 @@ toml = "0.4.2"
 serde = { version = "1.0", features = ["derive"] }
 
 [package.metadata.config]
-mbed-crypto-version = "mbedcrypto-1.1.0"
+mbed-crypto-version = "mbedcrypto-2.0.0"
 
 [features]
 default = ["mbed"]
--- a/src/providers/mbed_provider/mod.rs
+++ b/src/providers/mbed_provider/mod.rs
@@ -29,6 +29,8 @@ use parsec_interface::operations::{OpExportPublicKey, ResultExportPublicKey};
 use parsec_interface::operations::{OpImportKey, ResultImportKey};
 use parsec_interface::operations::{OpListOpcodes, ResultListOpcodes};
 use parsec_interface::requests::{Opcode, ProviderID, ResponseStatus, Result};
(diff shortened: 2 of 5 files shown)


ionut-arm:
One thing to note is that because of bindgen being unable to generate bindings for `static inline` functions, a lot of the functions that operate on `psa_key_attributes_t` do not receive a binding. At the moment we're cutting corners, setting the parameters "by hand".

Also, we'll need, at some point, to implement what Mbed Crypto calls `domain_parameters` - extra crypto parameters for specific algorithms (RSA, DSA...). Or at least decide how we want to handle them

Cargo.toml
```diff
@@ -34,7 +34,7 @@ toml = "0.4.2"
 serde = { version = "1.0", features = ["derive"] }
 
 [package.metadata.config]
-mbed-crypto-version = "mbedcrypto-1.1.0"
+mbed-crypto-version = "mbedcrypto-2.0.0"
```
hug-dev: 😎

src/providers/mbed_provider/mod.rs
```diff
@@ -105,13 +104,13 @@ fn create_key_id(
...
     local_ids_handle: &mut LocalIdStore,
-) -> Result<psa_crypto_binding::psa_key_id_t> {
-    let mut key_id = rand::random::<psa_crypto_binding::psa_key_id_t>();
+) -> Result<KeyId> {
+    let mut key_id = rand::random::<KeyId>();
     while local_ids_handle.contains(&key_id)
-        && key_id != 0
-        && key_id < constants::PSA_MAX_PERSISTENT_KEY_IDENTIFIER
+        || key_id == 0
+        || key_id > constants::PSA_MAX_PERSISTENT_KEY_IDENTIFIER
```
hug-dev: Good catch :P

ionut-arm: I'm still horrified I made that mistake in the first place.

hug-dev: It happens to the best of us.

PR #58, opened 2019-10-29
Make buffers dynamically sized in Mbed Provider

### LINKED ISSUE #27: Use dynamically-sized buffers in Mbed provider
At the moment the Mbed Crypto provider uses fixed-size (1024 bytes) buffers for operations where data is returned from Mbed Crypto, without checking if the operation failed for lack of space. 

This could be improved by either computing a sufficient size before the operation and/or by checking the result value in case the failure had to do with the buffer size. I would lean towards just checking the result, although allocating too much memory might become costly in a busy environment (do we care for the Mbed provider?). So, a fix would be to allocate 1024 bytes, check the result and if it was not enough allocate more and try again (up to a configurable threshold). 

The two operations that currently export any data are "sign" and "export public key". When we get to encrypting data we'll definitely have to pay more attention to buffer sizes.

ionut-arm:
Buffer sizes are now dynamically computed based on the macros provided in `crypto_sizes.h`.

Resolves #27

### DIFF
--- a/src/providers/mbed_provider/utils.rs
+++ b/src/providers/mbed_provider/utils.rs
@@ -18,7 +18,7 @@ use super::psa_crypto_binding::{
     psa_key_policy_s, psa_key_type_t, psa_key_usage_t, psa_status_t,
 };
 use parsec_interface::operations::key_attributes::*;
-use parsec_interface::requests::ResponseStatus;
+use parsec_interface::requests::{ResponseStatus, Result};
 use std::convert::TryFrom;
 
 /// Converts between native PARSEC key attributes and ID and the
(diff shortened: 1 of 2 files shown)


src/providers/mbed_provider/utils.rs
```diff
@@ -186,3 +186,34 @@ pub fn convert_status(psa_status: psa_status_t) -> ResponseStatus {
...
 }
+
+macro_rules! bits_to_bytes {
+    ($size:expr) => {
+        ($size + 7) / 8
+    };
+}
+
+/// Compute the size of the asymmetric signature, given the key attributes of the signing key.
+pub fn psa_asymmetric_sign_output_size(key_attrs: &psa_key_attributes_t) -> Result<usize> {
```
hug-dev: Maybe it would be nice to add where do those functions come from? Something like which C file, which macro are they implementing? It would help for maintaining in the future I think.

ionut-arm: Done

PR #59, opened 2019-10-29
Make key slot release implicit

### LINKED ISSUE #57: Drop key handles implicitly
We should remove the mental strain of remembering to "release" key slots whenever we finish an operation by using a [finally](https://crates.io/crates/finally-block) closure to do it for us whenever the key handle goes out of scope.

it'd probably work something like:

```
let key_handle = open_key(...);
let _finally = finally( || release_key(key_handle, &self.key_handle_mutex) );
```

ionut-arm:
Key slots are now ✨automagically✨ released when the `Key` object is dropped. 

Resolves #57

### DIFF
--- a/src/providers/mbed_provider/utils.rs
+++ b/src/providers/mbed_provider/utils.rs
@@ -14,12 +14,15 @@
 // limitations under the License.
 use super::constants::*;
 use super::psa_crypto_binding::{
-    psa_algorithm_t, psa_core_key_attributes_t, psa_key_attributes_t, psa_key_bits_t, psa_key_id_t,
-    psa_key_policy_s, psa_key_type_t, psa_key_usage_t, psa_status_t,
+    self, psa_algorithm_t, psa_core_key_attributes_t, psa_key_attributes_t, psa_key_bits_t,
+    psa_key_handle_t, psa_key_id_t, psa_key_policy_s, psa_key_type_t, psa_key_usage_t,
+    psa_status_t,
 };
+use log::error;
 use parsec_interface::operations::key_attributes::*;
 use parsec_interface::requests::{ResponseStatus, Result};
 use std::convert::TryFrom;
+use std::sync::Mutex;
 
 /// Converts between native PARSEC key attributes and ID and the
 /// `psa_key_attributes_t` structure required by Mbed Crypto.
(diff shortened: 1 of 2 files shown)


src/providers/mbed_provider/utils.rs
```diff
@@ -219,3 +222,79 @@ pub fn psa_export_public_key_size(key_attrs: &psa_key_attributes_t) -> Result<us
         _ => Err(ResponseStatus::PsaErrorInvalidArgument),
     }
 }
+
+/// Wrapper around raw `psa_key_handle_t` which allows for easier manipulation of
+/// handles and the attributes associated with them.
+pub struct Key<'a>(psa_key_handle_t, &'a Mutex<()>);
```
hug-dev: Maybe it should be `KeyHandle`?

src/providers/mbed_provider/utils.rs
```diff
@@ -219,3 +222,79 @@ pub fn psa_export_public_key_size(key_attrs: &psa_key_attributes_t) -> Result<us
...
+            Err(convert_status(get_attrs_status))
+        } else {
+            Ok(key_attrs)
+        }
+    }
+
+    /// Release the key stored under this handle.
+    pub fn release_key(&mut self) {
+        unsafe {
+            let _guard = self.1.lock().expect("Grabbing key handle mutex failed");
```
hug-dev: I am wondering if there are any implications/safety concerns of putting a safe statement inside an `unsafe` block? Maybe that could be outside of it and still work?

src/providers/mbed_provider/utils.rs
```diff
@@ -219,3 +222,79 @@ pub fn psa_export_public_key_size(key_attrs: &psa_key_attributes_t) -> Result<us
...
+
+    /// Extract the raw handle value.
+    pub fn raw_handle(&self) -> psa_key_handle_t {
+        self.0
+    }
+}
+
+impl Drop for Key<'_> {
+    fn drop(&mut self) {
+        if self.0 != 0 {
```
hug-dev: Will the `psa_close_key` fail if the handle is `0`? Maybe that check could be put in the `release_key` function if so.

src/providers/mbed_provider/utils.rs
ionut-arm: That's how it was named initially, but I renamed it because I thought an operation called `get_key_attributes` sounds a bit weird for a key handle.

src/providers/mbed_provider/utils.rs
ionut-arm: I'll have a look through the Nomicon again, see what they think

src/providers/mbed_provider/utils.rs
ionut-arm: Not sure it does - we don't check the return status for `close_key` anyway! But we could put a check in `release_key` too.

src/providers/mbed_provider/utils.rs
hug-dev: Fair enough!

src/providers/mbed_provider/utils.rs
hug-dev: I was thinking that if one of those safe function that we put in an `unsafe` block suddendly become `unsafe`, then we will still execute them without any warning 😕 
But I do not know if that can happen

src/providers/mbed_provider/utils.rs
ionut-arm: I've updated - now the check is done in `release_key` which is called every time a `Key` is dropped

PR #61, opened 2019-10-29
Add cross-compilation logic for Mbed Crypto

hug-dev:
When compiling the service with
`cargo build --target aarch64-unknown-linux-gnu`
Mbed Crypto should also be compiled for the same target and all object
files should be linked together for that target. This commit adds the
necessary logic to achieve that.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -23,6 +23,11 @@ const MBED_CRYPTO_VERSION_KEY: &str = "mbed-crypto-version";
 const SETUP_MBED_SCRIPT_PATH: &str = "./setup_mbed_crypto.sh";
 const BUILD_CONFIG_FILE_PATH: &str = "./build-conf.toml";
 
+const DEFAULT_NATIVE_MBED_COMPILER: &str = "clang";
+const DEFAULT_NATIVE_MBED_ARCHIVER: &str = "ar";
+const DEFAULT_ARM64_MBED_COMPILER: &str = "aarch64-linux-gnu-gcc";
+const DEFAULT_ARM64_MBED_ARCHIVER: &str = "aarch64-linux-gnu-ar";
+
 #[derive(Debug, Deserialize)]
 struct Configuration {
     mbed_config: Option<MbedConfig>,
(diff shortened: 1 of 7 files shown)


build.rs
```diff
@@ -70,45 +81,66 @@ fn setup_mbed_crypto(mbed_config: &MbedConfig) {
     let metadata = package
         .metadata
         .expect("Cargo.toml does not contain package metadata.");
-    let plasma_config = get_value_from_table(&metadata, CONFIG_TABLE_NAME);
```
ionut-arm: Wow, 'plasma' was still a thing. 😮

build.rs
```diff
@@ -70,45 +81,66 @@ fn setup_mbed_crypto(mbed_config: &MbedConfig) {
...
-    let project_dir = env::var("CARGO_MANIFEST_DIR")
-        .expect("CARGO_MANIFEST_DIR environment variable was not found");
+    let header = mbed_include_dir.clone() + "/psa/crypto.h";
 
-    println!(
-        "cargo:rustc-link-search=native={}/target/mbed-crypto/library/",
-        project_dir
-    );
-    println!("cargo:rustc-link-lib=static=mbedcrypto");
+    println!("cargo:rerun-if-changed={}", header);
```
ionut-arm: 👍 thanks for this

build.rs
```diff
@@ -70,45 +81,66 @@ fn setup_mbed_crypto(mbed_config: &MbedConfig) {
...
-    if let Some(mbed_compiler) = &mbed_config.mbed_compiler {
-        run_script.arg(format!("CC={}", mbed_compiler));
+    let toolchain;
+    let mbed_compiler;
+    let mbed_archiver;
+    if std::env::var("TARGET").unwrap() == "aarch64-unknown-linux-gnu" {
+        toolchain = mbed_config.aarch64_unknown_linux_gnu.as_ref().unwrap();
+        mbed_compiler = toolchain
+            .mbed_compiler
+            .clone()
```
ionut-arm: Do you need to clone here? Since you only use the value in an if-else, you should be able to just extract it out of `toolchain`

hug-dev: If I remove the `clone` I have the following error:
```
error[E0507]: cannot move out of `toolchain.mbed_compiler` which is behind a shared reference
  --> build.rs:95:25
   |
95 |           mbed_compiler = toolchain
   |  _________________________^
96 | |             .mbed_compiler
   | |__________________________^ move occurs because `toolchain.mbed_compiler` has type `std::option::Option<std::string::String>`, which does not implement the `Copy` trait
help: consider borrowing the `Option`'s content
   |
95 |         mbed_compiler = toolchain
96 |             .mbed_compiler.as_ref()
   |
```
If I use `as_ref` however then in the `unwrap_or` I have on one side a `&String` and one the other a `&str`. I did not find a good way to convert both to the same type so I just assumed using `clone` was acceptable here because of that 😃

PR #64, opened 2019-10-30
Add cross compilation tests to the CI with cross

hug-dev:
Uses rust-embedded/cross tool to cross-compile and execute the unit
tests on QEMU for the Arm64 on Linux target.
Makes the necessary changes to make it work.


hug-dev:
Importing the changes to the interface that should fix the issue.

hug-dev:
And it did 🎇

PR #71, opened 2019-11-19
Add SIGHUP signal handling to reload configuration

hug-dev:
Following daemon(7), it is recommended that systemd daemons reload the
configuration files when SIGHUP is received. This commit implements that
by dropping the Parsec structures and re-instanciating them with the new
configuration.
Adds a Drop trait implementation on the providers for them to
unitialize/free the memory structures of their backends when they are
dropped.
Modifies the test script and CI workflow to use reloading.

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -14,7 +14,7 @@
 // limitations under the License.
 use log::info;
 use parsec::utils::{ServiceBuilder, ServiceConfig};
-use signal_hook::{flag, SIGTERM};
+use signal_hook::{flag, SIGHUP, SIGTERM};
 use std::io::Error;
 use std::sync::{
     atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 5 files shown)


src/bin/main.rs
```diff
@@ -26,39 +26,57 @@ const CONFIG_FILE_PATH: &str = "./config.toml";
...
+            threadpool.join();
+
+            // Explicitely call drop now because otherwise Rust will drop these variables only
+            // after they have been overwritten, in which case some values/libraries might be
+            // initialized twice.
+            drop(front_end_handler);
+            drop(listener);
+            drop(threadpool);
+
+            config_file = ::std::fs::read_to_string(CONFIG_FILE_PATH)
```
ionut-arm: You can probably pull this out into a method that returns a tuple with 3 elements (`front_end_handler`, `listener`, `threadpool`) to remove the duplication.

hug-dev: I agree that code duplication is not the best but here that does not annoy me that much as I think those three components are too different to be returned by the same method?

Issue #72, opened 2019-12-05
Review response codes returned by providers


hug-dev:
The response codes returned by the `Provide` methods should be reviewed and documented in our book. We are currently not consistent amoung all of our providers and the response codes should be conform to what PSA describes.

hug-dev:
For example there is a mix between `KeyAlreadyExists` and `PsaErrorAlreadyExists`

hug-dev:
As part of this issue we should also make sure that we have a correct and exhaustive mapping between provider errors codes and PSA ResponseStatus. We currently always return `PsaErrorHardwareFailure` when a provider fails.

PR #75, opened 2019-12-06
Add a TPM provider

hug-dev:
This provider uses the tss-esapi crate to interface with a TSS 2.0
Enhanced System library.
Adds new configuration entries to select the TPM provider and which
TCTI device to use.
The TPM provider currently only supports RSA key, signing and verifying
operations.

Co-authored-by: Ionut Mihalcea <ionut.mihalcea@arm.com>

### DIFF
--- a/src/providers/tpm_provider/mod.rs
+++ b/src/providers/tpm_provider/mod.rs
@@ -1,0 +1,461 @@
+// Copyright (c) 2019, Arm Limited, All Rights Reserved
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//          http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+use super::Provide;
+use crate::authenticators::ApplicationName;
+use crate::key_id_managers::{KeyTriple, ManageKeyIDs};
+use log::{error, info};
+use parsec_interface::operations::key_attributes::*;
+use parsec_interface::operations::ProviderInfo;
+use parsec_interface::operations::{OpAsymSign, ResultAsymSign};
+use parsec_interface::operations::{OpAsymVerify, ResultAsymVerify};
+use parsec_interface::operations::{OpCreateKey, ResultCreateKey};
+use parsec_interface::operations::{OpDestroyKey, ResultDestroyKey};
+use parsec_interface::operations::{OpExportPublicKey, ResultExportPublicKey};
+use parsec_interface::operations::{OpImportKey, ResultImportKey};
+use parsec_interface::operations::{OpListOpcodes, ResultListOpcodes};
+use parsec_interface::requests::{Opcode, ProviderID, ResponseStatus, Result};
+use serde::{Deserialize, Serialize};
+use serde_asn1_der::asn1_wrapper::*;
+use std::sync::{Arc, Mutex, RwLock};
+use tss_esapi::{
+    constants::TPM2_ALG_SHA256, response_code::Tss2ResponseCodeKind, utils::AsymSchemeUnion,
+    utils::Signature, utils::TpmsContext, Tcti,
+};
+use uuid::Uuid;
+extern crate num_bigint_dig as num_bigint;
+use num_bigint::{BigInt, Sign};
+
+const SUPPORTED_OPCODES: [Opcode; 7] = [
+    Opcode::CreateKey,
+    Opcode::DestroyKey,
+    Opcode::AsymSign,
+    Opcode::AsymVerify,
+    Opcode::ImportKey,
+    Opcode::ExportPublicKey,
+    Opcode::ListOpcodes,
+];
+
+const ROOT_KEY_SIZE: usize = 2048;
+const ROOT_KEY_AUTH_SIZE: usize = 32;
+
+pub struct TpmProvider {
+    // The Mutex is needed both because interior mutability is needed to the ESAPI Context
+    // structure that is shared between threads and because two threads are not allowed the same
+    // ESAPI context simultaneously.
+    esapi_context: Mutex<tss_esapi::TransientObjectContext>,
+    // The Key ID Manager stores the key context and its associated authValue (a PasswordContext
+    // structure).
+    key_id_store: Arc<RwLock<dyn ManageKeyIDs + Send + Sync>>,
+}
+
+// Public exponent value for all RSA keys.
+const PUBLIC_EXPONENT: [u8; 3] = [0x01, 0x00, 0x01];
+const AUTH_VAL_LEN: usize = 32;
+
+// The RSA Public Key data are DER encoded with the following representation:
+// RSAPublicKey ::= SEQUENCE {
+//     modulus            INTEGER,  -- n
+//     publicExponent     INTEGER   -- e
+// }
+#[derive(Serialize, Deserialize, Debug)]
+struct RsaPublicKey {
+    modulus: IntegerAsn1,
+    public_exponent: IntegerAsn1,
+}
+
+// The PasswordContext is what is stored by the Key ID Manager.
+#[derive(Serialize, Deserialize)]
+struct PasswordContext {
+    context: TpmsContext,
+    auth_value: Vec<u8>,
+}
+
+/// Inserts a new mapping in the Key ID manager that stores the PasswordContext.
+fn insert_password_context(
+    store_handle: &mut dyn ManageKeyIDs,
+    key_triple: KeyTriple,
+    password_context: PasswordContext,
+) -> Result<()> {
+    let error_storing = |e| {
+        error!("Error storing a mapping: {}.", e);
+        Err(ResponseStatus::KeyIDManagerError)
+    };
+    let error_serializing = |e| {
+        error!("Error serializing the PasswordContext: {}.", e);
+        Err(ResponseStatus::KeyIDManagerError)
+    };
+
+    if store_handle
+        .insert(
+            key_triple,
+            bincode::serialize(&password_context).or_else(error_serializing)?,
+        )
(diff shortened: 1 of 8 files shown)


hug-dev:
The `tss-esapi` crate that is built by default, expect the TSS 2.0 to be installed on the system at a fixed location. For example the headers at `/usr/local/include/tss2/`, as per our build script.
That's why this PR fails on the CI, as TSS is not installed. It will also fail for users when they try to do a `cargo build` as the TPM Provider is a default feature.

Possible solutions that I see:

1. remove the TPM providers as default features and make sure they all compile when they are selected individually. That would be a short term solution and is definitely needed (the second part).
2. instead of linking TSS 2.0 at link time, we could dynamically load the library using libloading [as it is done for PKCS 11](https://github.com/mheese/rust-pkcs11/blob/master/src/lib.rs). We would then have to write the interface by hand or maybe there is a way to use bindgen for that as well. The library path would be in the configuration as well.
3. Execute the CI inside a Dockerfile that contains the needed file. We can have one Dockerfile that contains Mbed Crypto, PKCS 11 simulation and the TPM server.

I think 1. and 3. are definitely needed for now and we can look at 2. later.

hug-dev:
parallaxsecond/parsec#76 addresses point 1

hug-dev:
Check rust-lang/rust-bindgen#1541 for point 2.

hug-dev:
Executed `./all.sh` locally and all tests passed 😄 We will need to put tests on the CI as well!

src/providers/tpm_provider/mod.rs
```diff
@@ -0,0 +1,462 @@
...
+        let mut esapi_context = self
+            .esapi_context
+            .lock()
+            .expect("ESAPI Context lock poisoned");
+
+        let (key_context, auth_value) = esapi_context
+            .create_rsa_signing_key(key_size, AUTH_VAL_LEN)
+            .or_else(|e| {
+                error!("Error creating a RSA signing key: {}.", e);
+                dbg!(e);
```
ionut-arm: Probably not needed

src/providers/tpm_provider/mod.rs
```diff
@@ -0,0 +1,462 @@
...
+            .esapi_context
+            .lock()
+            .expect("ESAPI Context lock poisoned");
+
+        let public_key: RsaPublicKey = serde_asn1_der::from_bytes(&key_data).or_else(|err| {
+            error!("Could not deserialise key elements: {}.", err);
+            Err(ResponseStatus::PsaErrorCommunicationFailure)
+        })?;
+        if public_key.public_exponent.to_bytes_be().1 != PUBLIC_EXPONENT {
+            error!("The TPM Provider only supports 0x101 as public exponent for RSA public keys, {:?} given.", public_key.public_exponent.to_bytes_be());
```
ionut-arm: Maybe we should have an issue to change that

src/providers/tpm_provider/mod.rs
```diff
@@ -0,0 +1,462 @@
...
+            return Err(ResponseStatus::UnsupportedOperation);
+        }
+        let key_data = public_key.modulus.to_bytes_be().1;
+
+        let len = key_data.len();
+        if len != 128 && len != 256 {
```
ionut-arm: I might've mislead you, not sure that's actually true, will look into it

ionut-arm: I think the TPM can support whatever sizes it wants (larger than 1024)

src/providers/tpm_provider/mod.rs
```diff
@@ -0,0 +1,462 @@
...
+        let key_triple = KeyTriple::new(app_name, ProviderID::TpmProvider, key_name);
+
+        let store_handle = self.key_id_store.read().expect("Key store lock poisoned");
+        let mut esapi_context = self
+            .esapi_context
+            .lock()
+            .expect("ESAPI Context lock poisoned");
+
+        let len = hash.len();
+        if len > 64 {
```
ionut-arm: Ideally this should take into account the digest size of the hash algorithm that the key is created with - I think we actually have hash algorithms assigned to keys, maybe we can do something between the TPM crate and the provider to be able to get the public definition of the key and deduce from that the correct length

ionut-arm: Same for sign, of course

ionut-arm: Alternatively, these checks could be left to the crate to do by itself (which it actually is doing atm)

src/providers/tpm_provider/mod.rs
```diff
@@ -0,0 +1,462 @@
...
+            .expect("Conversion to usize failed.");
+
+        let mut store_handle = self.key_id_store.write().expect("Key store lock poisoned");
+        let mut esapi_context = self
+            .esapi_context
+            .lock()
+            .expect("ESAPI Context lock poisoned");
+
+        let (key_context, auth_value) = esapi_context
+            .create_rsa_signing_key(key_size, AUTH_VAL_LEN)
```
ionut-arm: I don't know if creating an "RSA key definition structure" and passing that here is more sensible. We should be setting the hash algorithm, for example.

Maybe we could have a method like `create_signing_key` which takes in an enum (`RSA` or `ECC`) and the enum contains the other params as well... Don't know, really

src/providers/tpm_provider/mod.rs
hug-dev: It is a limitation given by the TSS 2.0 specs, check https://trustedcomputinggroup.org/wp-content/uploads/TPM-Rev-2.0-Part-2-Structures-01.38.pdf section 12.2.3.5 "`TPMS_RSA_PARMS`", page 133

> A TPM compatible with this specification and supporting RSA shall support two primes and an exponentof  zero.  Support  for  other  values  is  optional.  Use  of  other  exponents  in  duplicated  keys  is  not  recommended because the resulting keys would not be interoperable with other TPMs

And an exponent of zero maps to `0x101`:

> When  zero,  indicates  that  the  exponent  is  the  default  of 2^16 + 1

src/providers/tpm_provider/mod.rs
hug-dev: You have a good eye 🦅 
Will remove

src/providers/tpm_provider/mod.rs
hug-dev: I will only keep the lower bound check then!

src/providers/tpm_provider/mod.rs
hug-dev: Is not it fine at the moment as we are supporting only one kind of hash algorithm?

src/providers/tpm_provider/mod.rs
hug-dev: I think it is fine for now as we are limiting the scope of things that we support.

src/providers/tpm_provider/mod.rs
ionut-arm: Not sure about that! We actually support any hash algorithm, at least with Mbed Crypto; but it depends on what people want from us

src/providers/tpm_provider/mod.rs
ionut-arm: my ELF eyes see everything

ionut-arm: apart from bugs in my own code :'(

src/providers/tpm_provider/mod.rs
ionut-arm: Given that the feature is optional, we could introduce it (as an option :D ) - default it to 0, but give ppl the possibility to change it

I'll raise an issue but not set the v1.0 Milestone on it

src/providers/tpm_provider/mod.rs
ionut-arm: True. We'll have to see how to manage that for the TPM crate, though

src/providers/tpm_provider/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+    auth_value: Vec<u8>,
+}
+
+/// Inserts a new mapping in the Key ID manager that stores the PasswordContext.
+fn insert_password_context(
+    store_handle: &mut dyn ManageKeyIDs,
+    key_triple: KeyTriple,
+    password_context: PasswordContext,
+) -> Result<()> {
+    let error_storing = |e| {
```
ionut-arm: Nice!!! I actually believe this could be moved to be something surfaced by the KeyIdManager - apart from the text everything seems general, maybe the KIM could do this before returning to the providers?

ionut-arm: Same for many of the `or_else` handlers in here

src/providers/tpm_provider/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+        let signature = Signature {
+            scheme: AsymSchemeUnion::RSASSA(TPM2_ALG_SHA256),
+            signature,
+        };
+
+        let password_context = get_password_context(&*store_handle, key_triple)?;
+
+        esapi_context
+            .verify_signature(password_context.context, &hash, signature)
+            .or_else(|e| {
```
ionut-arm: I suggest doing a mapping between `Tss2ResponseCode` or the kind and the Parsec `ResponseStatus` - an `impl From<Tss2......` would let you just do `?` - but maybe have this as another issue

src/providers/tpm_provider/mod.rs
hug-dev: The thing is that the `KeyIDManager` returns a `String` in its `Err` variant, so we would still need an `or_else` to convert it to `ResponseStatus` :/

src/providers/tpm_provider/mod.rs
hug-dev: Yes that is a good idea and something we should also do for the PKCS 11 provider. Currently both provider just return the generic "hardware failure" error. We will need that for the 1.0 milestone. Will raise an issue.

hug-dev: This is actually under existing issue #72

src/providers/tpm_provider/mod.rs
ionut-arm: Yes, but we could do that in the manager itself

hug-dev: Ok, got it now, that is a good idea!

hug-dev: Will create an issue.

hug-dev: In #77

PR #76, opened 2019-12-06
Make sure Cargo features work

hug-dev:
Makes sure that Parsec builds with commands like:
```
cargo build --no-default-features --features "mbed-crypto-provider"
cargo build --no-default-features --features "tpm-provider"
```

Also renames the "mbed" features to "mbed-crypto-provider" for
consistency.

Exit the service with an error code if no provider could be instanciated.

### DIFF
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -15,7 +15,7 @@
 use log::info;
 use parsec::utils::{ServiceBuilder, ServiceConfig};
 use signal_hook::{flag, SIGHUP, SIGTERM};
-use std::io::Error;
+use std::io::{Error, ErrorKind};
 use std::sync::{
     atomic::{AtomicBool, Ordering},
     Arc,
@@ -42,10 +42,12 @@ fn main() -> Result<(), Error> {
 
     info!("Parsec started. Configuring the service...");
 
+    let front_end_handler = ServiceBuilder::build_service(&config)
+        .ok_or_else(|| Error::new(ErrorKind::Other, "Parsec can not be configured."))?;
     // Multiple threads can not just have a reference of the front end handler because they could
     // outlive the run function. It is needed to give them all ownership of the front end handler
     // through an Arc.
-    let mut front_end_handler = Arc::from(ServiceBuilder::build_service(&config));
+    let mut front_end_handler = Arc::from(front_end_handler);
     let mut listener = ServiceBuilder::start_listener(&config.listener);
     let mut threadpool = ServiceBuilder::build_threadpool(config.core_settings.thread_pool_size);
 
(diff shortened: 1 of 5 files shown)


src/bin/main.rs
```diff
@@ -71,7 +78,13 @@ fn main() -> Result<(), Error> {
             config_file = ::std::fs::read_to_string(CONFIG_FILE_PATH)
                 .expect("Failed to read configuration file");
             config = toml::from_str(&config_file).expect("Failed to parse service configuration");
-            front_end_handler = Arc::from(ServiceBuilder::build_service(&config));
+            front_end_handler = Arc::from(match ServiceBuilder::build_service(&config) {
```
ionut-arm: `...build_service(&config).ok_or_else(|| Err(Error::new... )?...`

src/utils/service_builder.rs
```diff
@@ -185,7 +203,15 @@ fn get_provider(config: &ProviderConfig, key_id_manager: KeyIdManager) -> Provid
                         ))
                 .with_user_pin(config.user_pin.clone())
                 .build()
-                )
+                ))
+        }
+        #[cfg(not(all(feature = "mbed-crypto-provider", feature = "pkcs11-provider")))]
+        _ => {
+            warn!(
```
ionut-arm: I know we're not ready for a release yet, but this sounds like a case where we should panic instead of just printing a warning - if the config file defines a provider but that provider isn't enabled, the admin might be expecting some functionality to be available which will not be there

src/bin/main.rs
hug-dev: Nice one! Will change.

src/utils/service_builder.rs
hug-dev: > we should panic instead of just printing a warning

That is what I did first but having it return an `Option` allows us to do the following things:

* even if one of the providers could not be instantiated, Parsec can still run with the other ones
* an admin might reload the configuration with a bad one, adding a provider that is not in the binary: in that case, it will print a warning without the whole service shutting down
* I added a feature in this commit that checks if no providers could be instantiated in which case the Parsec binary will exit with an error code: I think that is still a better way to exit than panicking which can be confusing.

IMO returning always seems cleaner than panicking when it is possible.

> the admin might be expecting some functionality to be available which will not be there

You have a point there: I think it is better to solve this with the admin reading the logs and executing tests after Parsec has been installed than just panicking. I guess both options are possible and we will have to make a decision about this.

ionut-arm: I'm ok with this going in like this, we'll have to see if we get any feedback about it

PR #82, opened 2019-12-18
Allow optional providers and key ID managers

hug-dev:
That allows the Parsec initialization to return more meaningfull errors
than panicking inside the TOML parser.

### DIFF
--- a/src/utils/service_builder.rs
+++ b/src/utils/service_builder.rs
@@ -74,17 +74,21 @@ pub struct CoreSettings {
 pub struct ServiceConfig {
     pub core_settings: CoreSettings,
     pub listener: ListenerConfig,
-    pub key_manager: Vec<KeyIdManagerConfig>,
-    pub provider: Vec<ProviderConfig>,
+    pub key_manager: Option<Vec<KeyIdManagerConfig>>,
+    pub provider: Option<Vec<ProviderConfig>>,
 }
 
 pub struct ServiceBuilder;
 
 impl ServiceBuilder {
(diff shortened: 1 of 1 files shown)


src/utils/service_builder.rs
```diff
@@ -74,17 +74,26 @@ pub struct CoreSettings {
...
 
 impl ServiceBuilder {
     pub fn build_service(config: &ServiceConfig) -> Option<FrontEndHandler> {
-        let key_id_managers = build_key_id_managers(&config.key_manager);
+        let key_id_managers =
+            build_key_id_managers(config.key_manager.as_ref().unwrap_or(&Vec::new()));
 
-        let providers = build_providers(&config.provider, key_id_managers);
+        if key_id_managers.is_empty() {
+            error!("At least one key ID manager is needed by the providers. No valid key id manager could be created from the configuration.");
```
ionut-arm: I feel like requiring a key ID manager even when there's no provider is a bit weird. 

Also, what if in the future there'll be a provider that doesn't need a key ID manager?

hug-dev: > I feel like requiring a key ID manager even when there's no provider is a bit weird.

The code will not compile if there are no providers though (as per line 98).
But your second point makes sense and I can remove that check. If the KIM hashmap is empty the following error will be triggered when trying to build a provider:
```rust
  let key_id_manager = match key_id_managers.get(&config.key_id_manager) {
            Some(key_id_manager) => key_id_manager,
            None => {
                error!(
                    "Key ID manager with specified name was not found ({})",
                    config.key_id_manager
                );
                continue;
            }
```

Issue #69, opened 2019-11-15
Test strategy for our providers on the CI


hug-dev:
I think it could be tested by changing nothing but having a different configuration file which points to the SoftHSM library. There is an available [`Dockerfile`](https://github.com/psmiraglia/docker-softhsm) that installs the library that we can use.

I think we also have to think how we are going to scale testing of multiple providers on the CI.

I was thinking of having single provider tests (our current tests) and multiple providers tests that test more the configuration as opposed to a provider implementation. For that purpose we should have multiple configuration files for all of those cases: one per provider and another one including them all.

hug-dev:
We could add an optional argument when launching Parsec to specify a configuration file path and use this one. That way we can have in-tree one configuration file per provider and test them independently 😄 

And one more with all of them 😈

ionut-arm:
I think we can/should have a CLI option to set the path of the config file for the service and that could also help with tests.

A good way to deal with tests for different provider configurations would be to organize them into folders, place a

hug-dev:
We would have the following test structure (in `tests`):
```
all.sh -> Executing all of the tests below + rustfmt + clippy (will work only if everything is installed)
per_provider/ -> Test executed only with a single provider
    normal/
    stress/
    fuzz/
    persistence/
    PROVIDER_NAME/ -> Folder containing the Dockerfile and config.toml file per provider
        Dockerfile
        config.toml
all_providers/ -> Tests using all providers
    Dockerfile -> Would contain all providers (not published)
    config.toml
    mod.rs
cross_compilation/ -> Testing cross compilation of Parsec
    Dockerfile
    Cross.toml
```

The following `Dockerfile` would be needed:
1. one for the TPM provider + tss-esapi crate (there is one in parallaxsecond/rust-tss-esapi repo already)
2. one for the PKCS 11 provider
3. one for the Mbed Crypto provider + crate (cf #62)
4. one combining 1, 2 and 3 for the `all_providers` tests
5. one for cross-compilation, that already exists

Exactly the same tests (except `cross_compilation`) could be done on an Arm platform, on the CI, check #55 to also have coverage for Arm.

PR #83, opened 2019-12-20
Modify tests directory structure

hug-dev:
In order to execute the tests under different single providers, under
all providers and for cross-compilation, split the tests directory into
sub-directories. Each one contains a specific configuration file to run
the tests under.
Please check issue #69 for details.


tests/cross_compilation/config.toml
```diff
@@ -0,0 +1,10 @@
+[core_settings]
+# Using the default options for the core settings as Parsec will not be run for the
+# cross-compilation test.
+
+[listener]
+listener_type = "DomainSocket"
+timeout = 200 # in milliseconds
+
+# No providers are compiled in for the cross-compilation example.
```
ionut-arm: Not sure this is the case - the providers are compiled, this config file is just for actually running the service, right?

tests/per_provider/provider_cfg/pkcs11/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM ubuntu:latest
+
+RUN apt-get update && \
+    apt-get install -y wget automake autoconf libtool pkg-config && \
+    apt-get install -y curl libssl-dev
+	# These libraries are needed for bindgen as it uses libclang.so
```
ionut-arm: ? Needed?

tests/per_provider/ci.sh
```diff
@@ -0,0 +1,96 @@
...
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+# Per provider CI test script
+#
+# This script will execute integration tests targeting a platform with a single provider included.
+# It is meant to be executed inside one of the container which Dockerfiles
+# are in tests/per_provider/provider_cfg/*/.
+#
+# Usage: ./tests/per_provider/ci.sh PROVIDER_NAME
```
ionut-arm: Might be worth adding a check at the beginning of scripts that *require* an argument and print something like this if that argument wasn't provided

tests/all_providers/ci.sh
```diff
@@ -17,52 +17,51 @@
...
+# Unit tests and doc tests #
+############################
+RUST_BACKTRACE=1 cargo test --lib $FEATURES
+RUST_BACKTRACE=1 cargo test --doc $FEATURES
 
 #####################
 # Integration tests #
 #####################
-RUST_BACKTRACE=1 RUST_LOG=info cargo run &
+RUST_BACKTRACE=1 cargo run $FEATURES \
```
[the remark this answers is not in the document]
hug-dev: No, because we do a `cargo build` before so that is alsmost instantaneous 😸

tests/cross_compilation/config.toml
hug-dev: Well as of now, the cross-compilation test only compiles the service for Aarch64 and does not even execute any test, so this file is useless. I will remove it.

tests/per_provider/ci.sh
hug-dev: Good idea! Will add.

tests/per_provider/provider_cfg/pkcs11/Dockerfile
hug-dev: Will remove that

PR #90, opened 2020-01-21
Modify Travis CI test script

hug-dev:
Modifies the Dockerfile and removes the all-providers tests so that the
tests pass on Aarch64.


tests/all_providers/config.toml
```diff
@@ -1,5 +1,5 @@
 [core_settings]
-log_level = "debug"
+log_level = "error"
```
hug-dev: I had to do this to limit the log size, otherwise the build was failing on Travis CI.

PR #92, opened 2020-01-27
Replace most panicking behaviours with Result

hug-dev:
This commit looks at:
* unwrap
* expect
* panic
* unreachable
* unimplemented
functions/macros and tries to replace most of them with returning errors
instead.

Exceptions are:
* inside tests
* when using with read/write locks or mutexes
* env::var method
* converting from u32 to usize
* parsing Uuid

I took the liberty to align all the non-`ResponseStatus` errors to `std::io::Result` errors. I had to choose one of the `ErrorKind` variant and in some cases this choice is maybe bad. Feel free to propose changes. Using `std::io::Result` makes it easy to integrate with other libraries that we use that also return `std::io::Result` but we could also define our own.

### DIFF
--- a/src/back/backend_handler.rs
+++ b/src/back/backend_handler.rs
@@ -21,6 +21,7 @@ use parsec_interface::requests::{
     request::RequestHeader, Request, Response, ResponseStatus, Result,
 };
 use parsec_interface::requests::{BodyType, ProviderID};
+use std::io::{Error, ErrorKind};
 
 /// Component responsible for unmarshalling requests, passing the operation
 /// to the provider and marshalling the result.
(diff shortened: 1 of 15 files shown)


src/front/domain_socket.rs
```diff
@@ -43,39 +43,46 @@ impl DomainSocketListener {
     /// - if a file/socket exists at the path specified for the socket and `remove_file`
     /// fails
     /// - if binding to the socket path fails
-    pub fn new(timeout: Duration) -> Self {
-        // If this PARSEC instance was socket activated (see the `parsec.socket`
+    pub fn new(timeout: Duration) -> Result<Self> {
+        // If this Parsec instance was socket activated (see the `parsec.socket`
```
ionut-arm: 👍

src/front/domain_socket.rs
```diff
@@ -128,8 +135,17 @@ impl DomainSocketListenerBuilder {
         self
     }
 
-    pub fn build(self) -> DomainSocketListener {
-        let timeout = self.timeout.expect("The listener timeout was not set");
+    pub fn build(self) -> Result<DomainSocketListener> {
+        let timeout = match self.timeout {
```
ionut-arm: `ok_or_else` instead of `match`?

src/back/backend_handler.rs
```diff
@@ -222,15 +223,29 @@ impl BackEndHandlerBuilder {
...
-            content_type: self.content_type.expect("Content type missing"),
-            accept_type: self.accept_type.expect("Accept type missing"),
-            version_min: self.version_min.expect("Version min missing"),
-            version_maj: self.version_maj.expect("Version maj missing"),
-        }
+    pub fn build(self) -> std::io::Result<BackEndHandler> {
+        Ok(BackEndHandler {
+            provider: self
+                .provider
+                .ok_or_else(|| Error::new(ErrorKind::InvalidData, "provider is missing"))?,
```
ionut-arm: If you don't also log something in the lambda, `ok_or(Error::new(ErrorKind::InvalidData, "provider is missing"))?` is simpler

src/providers/mbed_provider/utils.rs
```diff
@@ -81,13 +85,11 @@ pub fn convert_key_bits(key_size: u32) -> psa_key_bits_t {
...
+pub fn convert_key_type(key_type: KeyType) -> Result<psa_key_type_t> {
     match key_type {
-        KeyType::RsaKeypair => PSA_KEY_TYPE_RSA_KEYPAIR,
-        KeyType::RsaPublicKey => PSA_KEY_TYPE_RSA_PUBLIC_KEY,
-        _ => {
-            unimplemented!();
-        }
+        KeyType::RsaKeypair => Ok(PSA_KEY_TYPE_RSA_KEYPAIR),
+        KeyType::RsaPublicKey => Ok(PSA_KEY_TYPE_RSA_PUBLIC_KEY),
+        _ => Err(ResponseStatus::PsaErrorInvalidArgument),
```
ionut-arm: Wouldn't this be `UnsupportedParameters` as well?

src/front/domain_socket.rs
hug-dev: Yes good point, will change.

src/providers/mbed_provider/utils.rs
hug-dev: True, will change.

src/back/backend_handler.rs
hug-dev: I tried that but clippy complains:
```rust
error: use of `ok_or` followed by a function call
   --> src/front/domain_socket.rs:148:48
    |
148 |           DomainSocketListener::new(self.timeout.ok_or(
    |  ________________________________________________^
149 | |             Error::new(
150 | |                 ErrorKind::InvalidInput,
151 | |                 "listener timeout missing",
152 | |                 )
153 | |         )?)
    | |_________^
    |
    = note: `-D clippy::or-fun-call` implied by `-D clippy::all`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call
help: try this
    |
148 |         DomainSocketListener::new(self.timeout.ok_or_else(|| Error::new(
149 |                 ErrorKind::InvalidInput,
150 |                 "listener timeout missing",
151 |                 ))?)
    |
```
It also makes it easier to maintain and add a log in teh future :)

ionut-arm: Fair enough, this makes sense! Thanks for the explanation!

PR #97, opened 2020-01-31
Implement fuzz testing

ionut-arm:
This commit implements fuzz testing through libFuzzer and the framework around it
to continuously run the fuzzer in a Docker container, along with a service that
checks for updates on the repo.
The service builds the Docker image and launches the fuzzer, after which it waits
for any update to the service. Given that the fuzzing corpus will remain in-place,
whenever fuzzing is restarted for another service version, it will take off from the
already existing cases.
The Docker image has all the required components to run all the current providers -
Mbed, PKCS11 and TPM - and to continuously fuzz the service through a stub version of
the frontend handler. If a crash is detected, the fuzzer will notify the team and
continue fuzzing.


fuzz/run_fuzz.sh
```diff
@@ -0,0 +1,45 @@
...
+
+set +e
+
+while [ true ]
+do
+    # Run fuzzer
+    cargo +nightly fuzz run fuzz_service
+
+    # Notify about crash
+    echo "Here we'd ping the webhook to notify"
```
hug-dev: As we currently do not notify the crash, would it be easy to read the logs to find the crashes that happened? Could it be easier, for now, to remove the `while true` loop and exit at the first crash?

Cargo.toml
```diff
@@ -49,3 +50,4 @@ default = ["mbed-crypto-provider", "pkcs11-provider"]
 mbed-crypto-provider = []
 pkcs11-provider = ["pkcs11", "picky-asn1-der", "picky-asn1"]
 tpm-provider = ["tss-esapi", "picky-asn1-der", "picky-asn1"]
+fuzz = ["parsec-interface/fuzz", "arbitrary"]
```
hug-dev: Is the `parsec-interface/fuzz` feature still needed now? As we only pass a stream to the fuzzer and now longer a `Request` + `ApplicationName`?
I guess as discussed in #54 it will be used when fuzzing individual components, but for now should we remove it?

fuzz/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+
+[package]
+name = "parsec-fuzz"
+version = "0.0.0"
+authors = ["Automatically generated"]
+publish = false
+edition = "2018"
```
hug-dev: Do we care about filling those with the same values as in our other crates?

fuzz.sh
```diff
@@ -0,0 +1,60 @@
...
+    docker run -d --rm -v $(pwd):/parsec -w /parsec/fuzz --name $FUZZ_CONTAINER_NAME parsec/fuzz ./run_fuzz.sh
+elif [[ "$1" == "stop" ]]
+then
+    docker kill $FUZZ_CONTAINER_NAME
+elif [[ "$1" == "follow" ]]
+then
+    docker logs -f --tail 100 $FUZZ_CONTAINER_NAME
+elif [[ "$1" == "clean" ]]
+then
+    docker run -d --rm -v $(pwd):/parsec -w /parsec/fuzz --name $CLEANUP_CONTAINER_NAME parsec/fuzz ./cleanup.sh
```
hug-dev: The `cleanup.sh` script seems to only have a dependency on the `rm` command. Do we need to execute it inside the `parsec/fuzz` container? We could also write the commands directly here and remove the script.

fuzz/fuzz_targets/fuzz_service.rs
```diff
@@ -0,0 +1,88 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#![no_main]
+
+use parsec::utils::{ServiceBuilder, ServiceConfig};
+use parsec::authenticators::ApplicationName;
+use parsec::fuzz_utils::requests::Request;
```
hug-dev: I don't think these two `use` are needed anymore. cf the comment on the `parsec-interface-rs` `fuzz` feature :)

fuzz/config.toml
```diff
@@ -0,0 +1,28 @@
+[core_settings]
+log_level = "error"
```
hug-dev: If I am not wrong, this will be ignored right? Maybe better to remove it to not add confusion.

fuzz/fuzz_targets/fuzz_service.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    fn flush(&mut self) -> Result<()> {
+        Ok(())
+    }
+}
+
+fuzz_target!(|stream: MockStream| {
+    FRONT_END_HANDLER.handle_request(stream);
+});
```
hug-dev: That is really awesome 💯 
Because of the `MockStream` `Read + Write` interface and the modularity of the design this is amazing to see how easy it is to re-use almost everything in Parsec!

fuzz/run_fuzz.sh
```diff
@@ -0,0 +1,45 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ------------------------------------------------------------------------------
+
+set -e
+set -x
```
hug-dev: As we were discussing it and just FYI, I was reading this [interesting article](https://sipb.mit.edu/doc/safe-shell/) about safe shell scripting and the recommended variables to set on top of shell scripts. Not requesting any changes here.

src/authenticators/mod.rs
```diff
@@ -21,9 +21,12 @@
 
 pub mod simple_authenticator;
 
+#[cfg(feature = "fuzz")]
+use arbitrary::Arbitrary;
 use parsec_interface::requests::request::RequestAuth;
 use parsec_interface::requests::Result;
 
+#[cfg_attr(feature = "fuzz", derive(Arbitrary))]
```
hug-dev: Are those still needed?

src/lib.rs
```diff
@@ -50,3 +50,8 @@ pub mod front;
 pub mod key_id_managers;
 pub mod providers;
 pub mod utils;
+
+#[cfg(feature = "fuzz")]
+pub mod fuzz_utils {
+    pub use parsec_interface::*;
+}
```
hug-dev: Same here: is it still needed?

fuzz.sh
```diff
@@ -0,0 +1,60 @@
...
+
+FUZZ_CONTAINER_NAME=parsec_fuzzer
+CLEANUP_CONTAINER_NAME=parsec_fuzzer_cleanup
+
+set -e
+
+if [[ "$1" == "run" ]]
+then
+    # Set up fuzz folder
+    cp fuzz/config.toml fuzz/run_config.toml
```
hug-dev: Why is it needed to copy the configuration file?

fuzz/run_fuzz.sh
ionut-arm: would be, but why not continue? the crash data (i.e. the input that triggered the crash) is saved by the fuzzer anyway. So far the only crash that happened was because the fuzzer kept using up more and more memory and died when it reached its limit - not exactly a meaningful crash

Cargo.toml
ionut-arm: yeah, i guess I can remove it and reintroduce when adding the other fuzzing stuff

fuzz.sh
ionut-arm: well yes, but no. if you try that on your server, in the parsec account, you'll see why :) when docker runs, it runs as root, so removing new files isn't possible from a non-sudoer account, unless you do something from a container again

fuzz.sh
ionut-arm: so that you don't have to remove the PKCS11 slot number after each use

src/authenticators/mod.rs
ionut-arm: Nope!

fuzz/config.toml
ionut-arm: true!

fuzz/Cargo.toml
ionut-arm: wasn't sure about that... it was obviously automatically generated, I can update it to be something similar to the parsec `Cargo.toml`

fuzz.sh
hug-dev: Ah makes sense! Maybe a comment would help here then

fuzz.sh
ionut-arm: I'll add a comment!

hug-dev: Thanks for the info!

fuzz/Cargo.toml
hug-dev: I think it would be good

fuzz/run_fuzz.sh
hug-dev: That's good then! As long as it is easy to know when a crash happenened and with what data.

ionut-arm: It doesn't only record crashes, it also records slow runs

fuzz/Cargo.toml
ionut-arm: I'll only update the authors, will leave version as "0.0.0"

Issue #91, opened 2020-01-24
Check for more Clippy lints


hug-dev:
The following command:
```bash
cargo clippy --all-targets --all-features -- -D clippy::all -D clippy::pedantic -D clippy::cargo
```
will enable *all* Clippy lints in all Parsec's code.
We should investigate which ones of those should be fixed.

hug-dev:
The same should be done in our other crates.

hug-dev:
All of our operation structures (example: `OpPing`) are declared in their `src/operations/...` (example: `ping.rs`) file.
Clippy, with the `pedantic` lint group enabled complains about that as it is a duplication of name:
```rust
error: item name ends with its containing module's name
  --> src/operations/ping.rs:18:1
   |
18 | pub struct OpPing;
   | ^^^^^^^^^^^^^^^^^^
   |
   = note: `-D clippy::module-name-repetitions` implied by `-D clippy::pedantic`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#module_name_repetitions

error: item name ends with its containing module's name
  --> src/operations/ping.rs:26:1
   |
26 | / pub struct ResultPing {
27 | |     pub supp_version_maj: u8,
28 | |     pub supp_version_min: u8,
29 | | }
   | |_^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#module_name_repetitions
```
The absolute name of the structure is `operations::ping::OpPing` but using namespacing we could rename this structure to be `Operation` only and so its absolute name would be `operations::ping::Operation`
In context where we need to use all the `OpXXX` or all the `ResultXXX`, we can import `operations::{ping, etc}` and use the different ones with `ping::Operation`, `ping::Result`, etc...

What are you thoughts about that?

ionut-arm:
That sounds good! We could eiither use `ping::Operation` or simply re-export them with an alias at the root of the library, i.e. `pub use ...::ping::Operation as OpPing`

hug-dev:
Following the same logic we should also rename all of our Protobuf operations to remove the `PingProto` from `OpPingProto` and `ResultPingProto` (and all of them) as it will be namespaced correctly (to be checked in other languages, like C).
Otherwise we have the same issue:
```rust
use super::generated_ops::ping::{OpPingProto, ResultPingProto};
```
which seems a bit weird if we do not fix as well.
That would incur a lot of changes to leaving that issue for now.

[...]

Issue #62, opened 2019-10-30
Create a PSA Crypto Rust wrapper crate


ionut-arm:
Our Mbed Provider is closely tied to the FFI layer to Mbed Crypto, which lead to questions being raised about the safety of the code, especially around `unsafe` blocks and how such blocks should be handled in the larger context. 

A sensible solution is to split out the code directly interfacing with Mbed Crypto into its own module/crate, keeping the `unsafe` calls contained.

hug-dev:
As part of that, we should also make the Mbed Crypto crate dynamically load the C shared library to remove the need of building it ourselves.

hug-dev:
I propose the following design/ideas for the Mbed Crypto crate.
It will have three goals:

1. Link the available Mbed Crypto library on the system (in the default installation location) to the crate. Thtat means that we will no longer compile it ourselves.
2. Produce `bindgen` Rust FFI to call the C library from Rust + write the needed preprocessor constants in a separate file (the equivalent of our current `constants.rs`). That would be the low-level, unsafe, layer.
3. On top of that layer produce a safe Rust wrapper that uses idomatic Rust features and abstract C types.

There seems to be a lot of duplication between things we use in the Parsec Rust interface and things we would abstract in this crate. For example, the `KeyAttributes` structure and the way we convert it to the `psa_key_attributes_t` C structure. Should we re-use the same types?
As in, should we have a `parsec-interface-rs` dependency in the Mbed Crypto crate? Should those structures be only defined once in the Mbed Crypto crate?

ionut-arm:
> should we have a `parsec-interface-rs` dependency in the Mbed Crypto crate?

I'm not sure that adds much value - we can have a dependency on the Mbed crypto crate in the interface crate and implement the conversions there. The Mbed-specific structures and some abstractions for the usage patterns should be defined in the Mbed crate, but nothing more

egrimley-arm:
If I understand correctly, `bindgen` generates Rust code that lets you directly access structures that are supposed to be opaque (such as `psa_key_attributes_t`), while it doesn't give you access to parts of the official API (such as `psa_set_key_lifetime`) which are defined as "static" functions or perhaps even as macros. So you might still have to write and compile a static C library consisting of shims/stubs. Or is there a better way?

hug-dev:
[...]

PR #99, opened 2020-02-04
Add cargo clippy lints to the CI

hug-dev:
This commit fixes the clippy cargo group lints and adds the check on the
CI script.
Also fixes some lints in the build script.

See #91 


src/lib.rs
```diff
@@ -17,15 +17,12 @@
     const_err,
     dead_code,
     improper_ctypes,
-    legacy_directory_ownership,
     non_shorthand_field_patterns,
     no_mangle_generic_items,
     overflowing_literals,
     path_statements,
     patterns_in_fns_without_body,
-    plugin_as_library,
```
ionut-arm: What's with these ones?

build.rs
```diff
@@ -12,6 +12,36 @@
 // WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+
+#![deny(
```
ionut-arm: No way to just have these in a "global" file somewhere, something like a `./clippy.toml`? :(

build.rs
```diff
@@ -75,63 +105,67 @@ fn get_value_from_table<'a>(table: &'a Value, key: &str) -> Result<&'a Value> {
...
+                    .clone()
+                    .unwrap_or_else(|| DEFAULT_ARM64_MBED_COMPILER.to_string()),
+                toolchain
+                    .mbed_archiver
+                    .clone()
+                    .unwrap_or_else(|| DEFAULT_ARM64_MBED_ARCHIVER.to_string()),
+            )
+        } else {
+            let toolchain;
+            toolchain = mbed_config.native.as_ref().ok_or_else(|| {
```
ionut-arm: Do you think we should keep the subtable as required but the keys in it as optional? The other way around sounds more sensible - put a subtable in your `build-config.toml` with the correct keys if you want to change smth

src/lib.rs
hug-dev: They have been removed as rustc lints and are hard errors now. When adding them there is a warning.

build.rs
hug-dev: I don't think it is available at the moment, the `clippy.toml` file seems to be for [configuring existing lints](https://github.com/rust-lang/rust-clippy#configuration).

build.rs
hug-dev: I don't really mind, maybe that could be discussed as part of #62 ?

ionut-arm: Sure!

Issue #105, opened 2020-02-13
Rectify and document our versioning system


hug-dev:
We use three different versions in Parsec:

1. The wire protocol version
2. The Parsec service version
3. The provider version

# The wire protocol version

It describes the version of the wire protocol itself: what are the different fields of requests and responses. It is needed to parse request and response headers.
The wire protocol might change in the future.
The `Ping` operation returns the highest wire protocol version supported by the service.
Clients need to perform a bootstrapping sequence if they want to switch to a higher wire protocol version:

1. Client send `Ping` using wire protocol version `1.0`
2. Service answer with highest wire protocol version available `x.y`
3. Client now knows that it can use wire protocol version `x.y` for further request.

The response are made by the service using the same wire protocol version than the request it was sent so that client can parse them.

The four first fields of the wire protocol: magic number, header size, version major and version minor **will never change** for requests and responses.

If a client sends a request with a wire protocol version strictly higher than the highest one supported by the service, a response status `VersionTooBig` will be sent back as an error.

## Actions to take

* Make sure that the wire protocol version system is documented properly and is consistent across all of our documentation
* Add in the [client library guide](https://parallaxsecond.github.io/parsec-book/parsec_client/writing_library.html) explanation about how the bootstrapping sequence should be performed.
* Modify the front-end handler to first parse the first four fields, check if the version sent is lower than the highest supported (send back `VersionTooBig` otherwise) and pass the rest to the appropriate wire protocol version deserializer. We should add the reverse logic on the return journey to serialize the response using the same wire protocol version than the request.

# The Parsec service version

This version is returned by the `ListProviders` operation in the `ProviderInfoProto` structure of the Core Provider. It represents the implementation version of the whole Parsec service. Everytime something is changed on the Parsec service, this number should be modified using the semantic versioning rules the best way possible. It is a way for clients to check if the Parsec version they are using contains the bug/security vulnerabilities fixes that they need or contains the new features that they want.
To make it easy, this version should be the same as the `version` field in the `Cargo.toml` file of the `parsec` crate.

## Actions to take

* Document this in the [ListProviders](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/list_providers.html) section.
* Should probably add a CHANGELOG page in the documentation describing the changes that each version brings so clients are aware.
* Add the logic in the Core Provider to directly fetch its version from the `Cargo.toml` file.

# The Provider version

[...]

PR #110, opened 2020-02-24
Fix the handling of versioning

hug-dev:
Make it clear, in the Core provider, that the ping operation is about
the wire format protocol version and not the provider version.
For the Core provider, its version is the one of the whole service.

Implements #105 


hug-dev:
Thanks! The persistent tests were badly named in the test script so I haev changed that. Added some `echo` as well so we can better see what is hapepning if the PKCS 11 provider tests fail again.

Issue #103, opened 2020-02-11
Split out ProviderConfig


ionut-arm:
Configuration for providers is currently being passed using a common `ProviderConfig`. Ideally, we would replace this with a configuration structure that contains some common parts and then an enum to cover the provider-specific options. This would make it easier to specify which ones are `Optional` and which ones aren't straight in the Config definition.

hug-dev:
Maybe complementary of #70

Issue #70, opened 2019-11-15
Modify configuration to have provider-specific table


hug-dev:
Currently we have one generic `ProviderConfig` structure for all of our providers. It might be better to have one configuration structure per provider type as most of them do not share the same fields.

PR #114, opened 2020-02-26
Modify configuration to have provider-specific structs

anta5010:
Use  Enums of structs with provider specific options instead of a common struct.

  Checks for missing parameters are not required in get_provider any more because it's done while parsing the config file into enums.
  
  This change covers https://github.com/parallaxsecond/parsec/issues/70 and https://github.com/parallaxsecond/parsec/issues/103

### DIFF
--- a/src/providers/mod.rs
+++ b/src/providers/mod.rs
@@ -26,32 +26,47 @@ pub mod mbed_provider;
 #[cfg(feature = "tpm-provider")]
 pub mod tpm_provider;
 
-#[derive(Copy, Clone, Deserialize, Debug)]
-pub enum ProviderType {
-    MbedProvider,
-    Pkcs11Provider,
-    TpmProvider,
+#[derive(Deserialize, Debug)]
+#[serde(tag = "provider_type")]
+pub enum ProviderConfig {
+    MbedProvider {
+        key_id_manager: String,
+    },
+    Pkcs11Provider {
+        key_id_manager: String,
+        library_path: String,
+        slot_number: usize,
+        user_pin: Option<String>,
+    },
+    TpmProvider {
+        key_id_manager: String,
+        tcti: String,
+        owner_hierarchy_auth: String,
+    },
 }
 
-impl ProviderType {
-    pub fn to_provider_id(self) -> ProviderID {
-        match self {
-            ProviderType::MbedProvider => ProviderID::MbedProvider,
-            ProviderType::Pkcs11Provider => ProviderID::Pkcs11Provider,
-            ProviderType::TpmProvider => ProviderID::TpmProvider,
+use self::ProviderConfig::*;
+impl ProviderConfig {
+    pub fn key_id_manager(&self) -> &String {
+        match *self {
+            MbedProvider {
+                ref key_id_manager, ..
+            } => key_id_manager,
+            Pkcs11Provider {
+                ref key_id_manager, ..
+            } => key_id_manager,
+            TpmProvider {
+                ref key_id_manager, ..
+            } => key_id_manager,
         }
     }
-}
-
-#[derive(Deserialize, Debug)]
-pub struct ProviderConfig {
-    pub provider_type: ProviderType,
-    pub key_id_manager: String,
-    pub library_path: Option<String>,
-    pub slot_number: Option<usize>,
-    pub user_pin: Option<String>,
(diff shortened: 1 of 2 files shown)


src/providers/mod.rs
```diff
@@ -26,32 +26,47 @@ pub mod mbed_provider;
 #[cfg(feature = "tpm-provider")]
 pub mod tpm_provider;
 
-#[derive(Copy, Clone, Deserialize, Debug)]
-pub enum ProviderType {
-    MbedProvider,
-    Pkcs11Provider,
-    TpmProvider,
+#[derive(Deserialize, Debug)]
+#[serde(tag = "provider_type")]
```
hug-dev: I don't understand what this `serde` attribute is for 😕

src/utils/service_builder.rs
```diff
@@ -232,37 +232,34 @@ unsafe fn get_provider(config: &ProviderConfig, key_id_manager: KeyIdManager) ->
...
-                        Error::new(ErrorKind::InvalidData, "missing slot number")
-                })?)
-                .with_user_pin(config.user_pin.clone())
-                .build()?,
-                ))
+                    .with_key_id_store(key_id_manager)
+                    .with_pkcs11_library_path(library_path.clone())
+                    .with_slot_number(*slot_number)
+                    .with_user_pin(user_pin.clone())
+                    .build()?,
```
hug-dev: So much clearer!

src/providers/mod.rs
anta5010: We need that for correctly parsing the parsec config file into the enum of structs. See "Internally tagged" on this page:
https://serde.rs/enum-representations.html

I though about adding a comment with the link, but it's the first link in google search results for "serde tag". I can add it if useful.

ionut-arm: I don't think it needs a comment, but @hug-dev might disagree. 

As for the feature use itself, I think it's very appropriate, makes it a lot more explicit what you need for each Provider. Thanks! :)

src/providers/mod.rs
```diff
@@ -26,32 +26,47 @@ pub mod mbed_provider;
...
+    },
 }
 
-impl ProviderType {
-    pub fn to_provider_id(self) -> ProviderID {
-        match self {
-            ProviderType::MbedProvider => ProviderID::MbedProvider,
-            ProviderType::Pkcs11Provider => ProviderID::Pkcs11Provider,
-            ProviderType::TpmProvider => ProviderID::TpmProvider,
+use self::ProviderConfig::*;
```
ionut-arm: Just as an FYI, really - we tend not to do wildcard exports (unless it's for tests), either for enum variants or for module contents. We had to do that for some stuff around the TPM because there were so many things that we had to import... Generally in Rust world it's seen as an anti-pattern because it pollutes the namespace and makes it less obvious where identifiers are coming from.

For example, this is a lint that was introduced specifically for this case (glob import of enum variants) in clippy: https://github.com/rust-lang/rust-clippy/issues/541 - it is only a pedantic lint, though, which we don't enable (for now?)

src/providers/mod.rs
hug-dev: Oh I see, with this tag the deserialization into the enum of structs maps exactly our `provider` format in the config file, because we already had the `provider_type` field. That way we do not have to change the configuration, that is neat!

I always think that a one line comment never hurt, even if it is superfluous. But only do it if you need to make other changes to the PR 😃

anta5010: done

src/providers/mod.rs
anta5010: fixed

src/providers/mod.rs
```diff
@@ -26,32 +26,52 @@ pub mod mbed_provider;
...
-#[derive(Copy, Clone, Deserialize, Debug)]
-pub enum ProviderType {
-    MbedProvider,
-    Pkcs11Provider,
-    TpmProvider,
+#[derive(Deserialize, Debug)]
+// For providers configs in parsec config.toml we use a format similar
+// to the one described in the Internally Tagged Enum representation
+// where "provider_type" is the tag field. For details see:
+// https://serde.rs/enum-representations.html
```
hug-dev: Great explanation, thank you!

PR #117, opened 2020-02-28
Improvments for tests/ci.sh

anta5010:
With this change the tests/ci.sh can be used for manual testing.

### DIFF
--- a/tests/ci.sh
+++ b/tests/ci.sh
@@ -19,8 +19,23 @@
 
 set -e
 
+# The clean up procedure is called when the script finished or is interrupted
+cleanup () {
+    echo "Shutdown Parsec and clean up"
+    # Stop Parsec if running
+    if [ -n "$PARSEC_PID" ]; then kill $PARSEC_PID || true ; fi
+    # Stop tpm_server if running
+    if [ -n "$TPM_SRV_PID" ]; then kill $TPM_SRV_PID || true; fi
+    # Remove the slot_number line added by find_slot_number.sh
+    sed -i '/^slot_number =.*/d' $CONFIG_PATH
+    # Remove fake mapping files
+    if [ -d "mappings" ]; then rm -rf -- "mappings"; fi
+
+    cargo clean
+}
+
 usage () {
-	printf "
+    printf "
 Continuous Integration test script
 
 This script will execute various tests targeting a platform with a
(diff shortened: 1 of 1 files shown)


tests/ci.sh
```diff
@@ -111,62 +118,61 @@ RUST_BACKTRACE=1 cargo test --doc $FEATURES
...
+    echo "Trigger a configuration reload to load the new mappings"
+    kill -s SIGHUP $PARSEC_PID
+    # Sleep time needed to make sure Parsec is ready before launching the tests.
+    sleep 5
+
+    echo "Execute persistent test, after the reload"
+    RUST_BACKTRACE=1 cargo test $FEATURES persistent_after
+
+    echo "Shutdown Parsec"
+    kill $PARSEC_PID
```
ionut-arm: Don't know why I haven't asked until now, but why do we kill the server here? Would a reload suffice? @hug-dev

tests/ci.sh
```diff
@@ -19,8 +19,23 @@
...
+    # Stop Parsec if running
+    if [ -n "$PARSEC_PID" ]; then kill $PARSEC_PID || true ; fi
+    # Stop tpm_server if running
+    if [ -n "$TPM_SRV_PID" ]; then kill $TPM_SRV_PID || true; fi
+    # Remove the slot_number line added by find_slot_number.sh
+    sed -i '/^slot_number =.*/d' $CONFIG_PATH
+    # Remove fake mapping files
+    if [ -d "mappings" ]; then rm -rf -- "mappings"; fi
+
+    cargo clean
```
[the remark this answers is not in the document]
anta5010: Made it optional

tests/ci.sh
```diff
@@ -39,55 +54,47 @@ where PROVIDER_NAME can be one of:
...
-	sleep 5
-	tpm2_startup -c -T mssim
-	tpm2_changeauth -c owner tpm_pass
-elif [[ $1 = "pkcs11" ]]
-then
-	# PKCS11 provider
-	FEATURES="--features=$1-provider"
-	CONFIG_PATH="tests/per_provider/provider_cfg/$1/config.toml"
+    # Start and configure TPM server
+    rm -f NVChip
```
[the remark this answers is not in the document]
anta5010: done

tests/ci.sh
```diff
@@ -19,8 +19,24 @@
...
+    if [ -n "$PARSEC_PID" ]; then kill $PARSEC_PID || true ; fi
+    # Stop tpm_server if running
+    if [ -n "$TPM_SRV_PID" ]; then kill $TPM_SRV_PID || true; fi
+    # Remove the slot_number line added by find_slot_number.sh
+    sed -i '/^slot_number =.*/d' $CONFIG_PATH
+    # Remove fake mapping and temp files
+    if [ -d "mappings" ]; then rm -rf -- "mappings"; fi
+    if [ -f "NVChip" ]; then rm "NVChip" ; fi
+
+    if [ -z "$NO_CARGO_CLEAN" ]; then cargo clean; fi
```
hug-dev: Thanks for adding the option!

tests/ci.sh
```diff
@@ -111,62 +128,58 @@ RUST_BACKTRACE=1 cargo test --doc $FEATURES
...
+PARSEC_PID=$!
 # Sleep time needed to make sure Parsec is ready before launching the tests.
 sleep 5
 
-if [[ $1 = "all" ]]
-then
-	echo "Execute all-providers tests"
-	RUST_BACKTRACE=1 cargo test $FEATURES all_providers
+# Check that Parsec successfully started and is running
+pgrep -f target/debug/parsec >/dev/null
```
hug-dev: Good idea! So that if the service does not start successfully we don't even launch the first test.

tests/ci.sh
hug-dev: The shutdown is needed because the logging level is changed for the stress tests. I guess we could modify the config file in place before that step to change the log level in the file and then reload it but I think that is simpler.

If we had the same logging level for stress tests and other tests, then we would not even have to reload it for the stress tests!

PR #130, opened 2020-03-16
Allow TPM provider to use TestParms

### LINKED ISSUE #121: TPM provider should establish most-secure primitives for itself
Following the RaspberryPi tests, it's quite apparent that not all TPMs support the primitives that we'd want to use by default in, for example, securing our authentication sessions.

Based on work to be done in https://github.com/parallaxsecond/rust-tss-esapi/issues/38 we can pre-select a list of primitives and their parameters, and find the most desirable one that's supported, at runtime.

ionut-arm:
This commit improves the usage of the TSS crate by the TPM provider. The
provider can now go through a limited list of predefined ciphers that it
can use for securing sessions. It also uses the new transient context
builder.

Closes #121

# ON HOLD

Test on RPi failed because only part of the usage of hardcoded ciphers was removed. Removing all of it would imply a bigger refactor in the TSS crate, which I'll do after the interface updates get done

### DIFF
--- a/src/providers/tpm_provider/mod.rs
+++ b/src/providers/tpm_provider/mod.rs
@@ -37,7 +37,13 @@ use serde::{Deserialize, Serialize};
 use std::io::ErrorKind;
 use std::sync::{Arc, Mutex, RwLock};
 use tss_esapi::{
-    constants::TPM2_ALG_SHA256, utils::AsymSchemeUnion, utils::Signature, utils::TpmsContext, Tcti,
+    abstraction::transient::TransientKeyContextBuilder,
+    constants::TPM2_ALG_SHA256,
+    utils::primitives::{Cipher, CipherKeyLength, CipherMode},
+    utils::AsymSchemeUnion,
+    utils::Signature,
+    utils::TpmsContext,
+    Tcti,
 };
 use uuid::Uuid;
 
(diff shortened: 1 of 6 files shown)


ionut-arm:
Shall I not assume that the underlying TPM can support 2048 bit RSA keys? From the spec:

> A TPM that supports RSA should support a public modulus size of at least 2,048 bits. Support for  other key sizes is permitted

So theoretically we can check if the TPM supports RSA keys. Don't know how critical this is since interest seems to be focusing on RSA keys anyway, so presumably any clients would already have RSA-compatible TPMs.

src/providers/tpm_provider/mod.rs
```diff
@@ -408,6 +410,15 @@ impl Drop for TpmProvider {
...
 
+lazy_static! {
+    static ref CIPHERS: Vec<Cipher> = {
+        let mut ciphers = Vec::new();
+        ciphers.push(Cipher::aes_256_cfb());
+        ciphers.push(Cipher::aes_128_cfb());
+        ciphers
+    };
+}
+
```
hug-dev: I like the clarity of this block but is it worth it to have it instead of doing the operation inside the `find_session_cipher` function? It will be called anyway only each time the provider is built.
I am just being cautious about the number of dependencies that we use 😃

hug-dev:
> 
> 
> Shall I not assume that the underlying TPM can support 2048 bit RSA keys? From the spec:
> 
> > A TPM that supports RSA should support a public modulus size of at least 2,048 bits. Support for  other key sizes is permitted
> 
> So theoretically we can check if the TPM supports RSA keys. Don't know how critical this is since interest seems to be focusing on RSA keys anyway, so presumably any clients would already have RSA-compatible TPMs.

Where do you assume that the underlying TPM supports 2048 bit RSA keys? I do not see it 😕

src/providers/tpm_provider/mod.rs
ionut-arm: Sure, can change! Just wanted one clear place where these are defined so you can clearly see what the values are, but I think I'll actually try and define it as an array and manually define the contents. Clearer, but taking a lot more space

PR #135, opened 2020-03-31
Add tests checking if key attributes are respected

hug-dev:
These tests check:
* asymmetric signature operations can only be done if they is of correct
type
* the specific algorithm used for those operations needs to be permitted
* the usage flags of the key need to allow the operation

The tests are disabled for now until the checking feature is implemented.

### DIFF
--- a/tests/per_provider/normal_tests/key_attributes.rs
+++ b/tests/per_provider/normal_tests/key_attributes.rs
@@ -1,0 +1,131 @@
+// Copyright (c) 2020, Arm Limited, All Rights Reserved
+// SPDX-License-Identifier: Apache-2.0
+use parsec_client_test::TestClient;
+use parsec_interface::operations::psa_algorithm::{Algorithm, AsymmetricSignature, Cipher, Hash};
+use parsec_interface::operations::psa_key_attributes::{
+    KeyAttributes, KeyPolicy, KeyType, UsageFlags,
+};
+use parsec_interface::requests::ResponseStatus;
+
+#[ignore]
+#[test]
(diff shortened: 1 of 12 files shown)


tests/per_provider/normal_tests/key_attributes.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    client
+        .generate_key(key_name.clone(), key_attributes)
+        .unwrap();
+    let status = client
+        .sign_with_rsa_sha256(key_name, vec![0xDE, 0xAD, 0xBE, 0xEF])
+        .unwrap_err();
+
+    assert_eq!(status, ResponseStatus::PsaErrorNotPermitted);
+}
```
ionut-arm: Nice 👍

tests/per_provider/normal_tests/key_attributes.rs
```diff
@@ -0,0 +1,131 @@
...
+                export: false,
+                encrypt: false,
+                decrypt: false,
+                cache: false,
+                copy: false,
+                derive: false,
+            },
+            key_algorithm: permitted_algorithm,
+        },
+    };
```
ionut-arm: I'm wondering if this should even be permitted to exist. Feels like the kind of thing we can enforce, because we know what is allowed and what isn't. We can't use type checking directly, but maybe we can use constructors (e.g. for `KeyAttributes`)

hug-dev: True! Those things could be done as part of parallaxsecond/parsec-interface-rs#32

PR #138, opened 2020-04-03
Add KeyInfo to the KeyIDManager

hug-dev:
The KeyIDManager also needs to store the KeyAttributes of the key. It
will eventually be renamed KeyInfoManager in a later commit.
Uses methods on KeyAttributes to check if an operation is permitted and
compatible with the type of the key.

### DIFF
--- a/src/providers/pkcs11_provider/mod.rs
+++ b/src/providers/pkcs11_provider/mod.rs
@@ -19,7 +19,7 @@
 use super::Provide;
 use crate::authenticators::ApplicationName;
 use crate::key_id_managers;
-use crate::key_id_managers::{KeyTriple, ManageKeyIDs};
+use crate::key_id_managers::{KeyInfo, KeyTriple, ManageKeyIDs};
 use derivative::Derivative;
 use log::{error, info, warn};
 use parsec_interface::operations::list_providers::ProviderInfo;
@@ -269,13 +269,16 @@ impl Drop for Session<'_> {
 }
 
 /// Gets a key identifier from the Key ID Manager.
-fn get_key_id(key_triple: &KeyTriple, store_handle: &dyn ManageKeyIDs) -> Result<[u8; 4]> {
+fn get_key_id(
+    key_triple: &KeyTriple,
+    store_handle: &dyn ManageKeyIDs,
+) -> Result<([u8; 4], KeyAttributes)> {
     match store_handle.get(key_triple) {
-        Ok(Some(key_id)) => {
-            if key_id.len() == 4 {
+        Ok(Some(key_info)) => {
+            if key_info.id.len() == 4 {
                 let mut dst = [0; 4];
-                dst.copy_from_slice(key_id);
-                Ok(dst)
+                dst.copy_from_slice(&key_info.id);
+                Ok((dst, key_info.attributes))
             } else {
                 error!("Stored Key ID is not valid.");
                 Err(ResponseStatus::KeyIDManagerError)
@@ -288,6 +291,7 @@ fn get_key_id(key_triple: &KeyTriple, store_handle: &dyn ManageKeyIDs) -> Result
 
 fn create_key_id(
     key_triple: KeyTriple,
+    key_attributes: KeyAttributes,
     store_handle: &mut dyn ManageKeyIDs,
     local_ids_handle: &mut LocalIdStore,
 ) -> Result<[u8; 4]> {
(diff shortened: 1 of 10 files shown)


hug-dev:
I will rename the structures and files `KeyIDManager` -> `KeyInfoManager` in a subsequent PR to make this one easier to review.

src/providers/pkcs11_provider/mod.rs
```diff
@@ -493,18 +501,13 @@ impl Provide for Pkcs11Provider {
...
 
-        if op.attributes.key_type != KeyType::RsaKeyPair
-            || op.attributes.key_policy.key_algorithm
-                != Algorithm::AsymmetricSignature(AsymmetricSignature::RsaPkcs1v15Sign {
-                    hash_alg: Hash::Sha256,
-                })
-        {
-            error!(
-                "The PKCS11 provider currently only supports creating RSA key pairs for signing and verifying. The signature algorithm needs to be RSA PKCS#1 v1.5 and the text hashed with SHA-256.");
+        if op.attributes.key_type != KeyType::RsaKeyPair {
```
ionut-arm, line 504: This might annoy you, but could we move this check somewhere else, so that it's "global"? Maybe in the interface or in `providers/mod.rs`, where we can just have a list of supported algos that we can add to as we go. Or does that not work because we'd need to implement them everywhere at the same time

ionut-arm: I'm just thinking that we'll keep adding stuff, this `if` will keep growing

src/providers/pkcs11_provider/mod.rs
```diff
@@ -928,11 +963,41 @@ impl Provide for Pkcs11Provider {
...
         let signature = op.signature;
         let key_triple = KeyTriple::new(app_name, ProviderID::Pkcs11, key_name);
         let store_handle = self.key_id_store.read().expect("Key store lock poisoned");
-        let key_id = get_key_id(&key_triple, &*store_handle)?;
+        let (key_id, key_attributes) = get_key_id(&key_triple, &*store_handle)?;
+
+        if !key_attributes.can_verify_hash() {
+            error!("Key {} does not permit verifying hashes.", key_triple);
+            return Err(ResponseStatus::PsaErrorNotPermitted);
+        }
```
ionut-arm, line 976: Do you reckon this could be pushed inside the `can_verify_hash` so it returns a `Result<()>` instead of a `bool`? The error code and log message seem general enough, you could then just do `key_attributes.can_verify_hash()?`.

Same for the `permits_alg` and `is_compatible_with_alg`

src/providers/pkcs11_provider/mod.rs
hug-dev: I was thinking about putting those checks in common put it is not an easy problem:

* we probably want indepentant checks per provider as we might implement key attributes/algs support differently in each provider
* those checks can be different in each functions: the supported key attributes can be different in generate and import_key for example. For operations where an algorithm is specified, different operations might support different algorithms in the same provider...

I see your point though that those conditions everywhere might be messy.

src/providers/pkcs11_provider/mod.rs
hug-dev: I think it is nice and clear for those methods to return `bool`, making them return a specific value of `ResponseStatus` feels a bit weird, just to allow this particular code to be simpler.
I could put the log message inside the method but I am not sure it would still be an `error!`: how to be sure that the caller calls it in a context where it is an error? They could intentionally expect the method to return false in a valid case.

hug-dev: Will add twin methods for those ones in the interface that return a `Result` and contain the log message.

hug-dev: Created parallaxsecond/parsec-interface-rs#39 for that

src/providers/pkcs11_provider/mod.rs
```diff
@@ -269,13 +269,16 @@ impl Drop for Session<'_> {
 }
 
 /// Gets a key identifier from the Key ID Manager.
-fn get_key_id(key_triple: &KeyTriple, store_handle: &dyn ManageKeyIDs) -> Result<[u8; 4]> {
+fn get_key_id(
```
ionut-arm, line 272: Shouldn't this be `get_key_info`

src/providers/pkcs11_provider/mod.rs
```diff
@@ -867,10 +869,26 @@ impl Provide for Pkcs11Provider {
...
+        let alg = op.alg;
         let mut hash = op.hash;
         let key_triple = KeyTriple::new(app_name, ProviderID::Pkcs11, key_name);
         let store_handle = self.key_id_store.read().expect("Key store lock poisoned");
-        let key_id = get_key_id(&key_triple, &*store_handle)?;
+        let (key_id, key_attributes) = get_key_id(&key_triple, &*store_handle)?;
+
+        key_attributes.can_sign_hash()?;
+        key_attributes.permits_alg(alg.into())?;
+        key_attributes.compatible_with_alg(alg.into())?;
```
ionut-arm, line 880: 😎

src/providers/pkcs11_provider/mod.rs
hug-dev: Yes it should! I will do the renaming of all text and filenames from key ID manager to key info manager in a subsequent PR.

PR #141, opened 2020-04-07
A few more README updates including fixes for broken doc links

paulhowardarm:
- A few adjustments to wording
- Fix a couple of broken doc links
- Minor heading/sectional adjustments
- Stylistic consistency fixes (embolden all hyperlinks)

Signed-off-by: Paul Howard <paul.howard@arm.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,16 +25,16 @@
 
 # Welcome To PARSEC
 
-**PARSEC** is the **P**latform **A**bst**R**action for **SEC**urity, a new open-source initiative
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -70,7 +73,7 @@ Contributions from the developer community are welcome. Please refer to the cont
 
 # Example
```
hug-dev, line 74: As a note: when the Rust client is ready it would be nice to put it here in the example section instead of the Test client 😃

Issue #129, opened 2020-03-16
Update with PSA Crypto 1.0.0 interface


hug-dev:
Parsec service needs to be updated to use parallaxsecond/parsec-interface-rs#4.

The new interface introduced the `alg` parameter for sign/verify operations. The algorithm used for signing/verifying needs to be checked against the key policy (which has to be stored somewhere).

ionut-arm:
If the underlying API doesn't offer the option to retrieve said algorithm, we can probably repurpose the `KeyIDManager` to cover more key-related information (and potentially change it to `KeyInfoManager`)

hug-dev:
About modification in the Key ID Manager and checking for policy I am thinking of the following things to do:
* add integration tests, per provider, checking that `PsaErrorNotPermitted` is returned if:
  * the algorithm is not compatible with the type of the key
  * the operation is not permitted by the usage flags
  * the algorithm is not the permitted algorithm in the policy
* modify the Key ID Manager trait and implementors to use a `KeyInfo` structure (instead of just the key ID):
```rust
/// Information stored about a key
#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
pub struct KeyInfo {
    /// Reference to a key in the Provider
    pub id: Vec<u8>,
    /// Attributes of a key
    pub attributes: KeyAttributes,
}
```
* for each operation, providers need to check if it is permitted and return `PsaErrorNotPermitted` if not. For that implements the following method on the `KeyAttributes` structure:
```rust
fn permits(&self, op: NativeOperation) -> bool
```
to check if the key permits the operation, ie the conditions on the first point are respected. That way, providers get the key attributes from the KIM, directly call this method and return `PsaErrorNotPermitted` if it's `false`.

hug-dev:
I will also add here two things to check as part of this task:

- [ ] `ResponseStatus`: check when it is adequate to return service ones or PSA ones
- [ ] For each operation: check if the behaviour is as described in the API and update the book

hug-dev:
Tests have been merged in #135

hug-dev:
All updates are done! [Last PR](https://github.com/parallaxsecond/parsec/pull/144)

PR #142, opened 2020-04-09
Modify error translation coming from providers

hug-dev:
Replaces most error translations with PsaErrorCommunicationFailure. The
reason for that is that Parsec controls which inputs are given to the
Provider library and should not give it data that triggers an error.
Defines some genuine errors that it is hard to detect before.
This is part of #129 to make sure that PSA response status are retruned correctly. Will add more test to check for the specific error defined in the operations.

### DIFF
--- a/src/providers/pkcs11_provider/utils.rs
+++ b/src/providers/pkcs11_provider/utils.rs
@@ -18,16 +18,21 @@ use parsec_interface::requests::ResponseStatus;
 use pkcs11::errors::Error;
 use pkcs11::types::*;
 
+/// Convert the PKCS 11 library specific error values to ResponseStatus values that are returned on
+/// the wire protocol
+///
+/// Most of them are PsaErrorCommunicationFailure as, in the general case, the calls to the PKCS11
+/// library should suceed with the values crafted by the provider.
+/// If an error happens in the PKCS11 library, it means that it was badly used by the provider or
+/// that it failed in an unexpected way and hence the PsaErrorCommunicationFailure error.
+/// The errors translated to response status are related with signature verification failure, lack
+/// of memory, hardware failure, corruption detection, lack of entropy and unsupported operations.
 pub fn to_response_status(error: Error) -> ResponseStatus {
     match error {
         Error::Io(e) => ResponseStatus::from(e),
-        Error::Module(e) => {
+        Error::Module(e) | Error::InvalidInput(e) => {
             error!("Conversion of error \"{}\"", e);
-            ResponseStatus::PsaErrorGenericError
-        }
-        Error::InvalidInput(e) => {
-            error!("Conversion of error \"{}\"", e);
-            ResponseStatus::PsaErrorInvalidArgument
+            ResponseStatus::PsaErrorCommunicationFailure
         }
         Error::Pkcs11(ck_rv) => rv_to_response_status(ck_rv),
     }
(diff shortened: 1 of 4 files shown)


src/providers/pkcs11_provider/utils.rs
```diff
@@ -18,16 +18,21 @@ use parsec_interface::requests::ResponseStatus;
 use pkcs11::errors::Error;
 use pkcs11::types::*;
 
+/// Convert the PKCS 11 library specific error values to ResponseStatus values that are returned on
+/// the wire protocol
+///
+/// Most of them are PsaErrorCommunicationFailure as, in the general case, the calls to the PKCS 11
+/// library should never fail. Parsec is totally in control of the input it passes to the library,
+/// if an error happens in the PKCS 11 library, it means that it was badly used by Parsec and hence
+/// the PsaErrorCommunicationFailure error.
```
ionut-arm, line 27: I think that's an overstatement. Sure, some errors would be due to Parsec and how it handles data, but there's a large stack underneath us that could fail in weird ways - other code can have bugs too, not just ours; I think we saw that quite clearly for the TPM stack.

I'm not against this change, but this doc here make it sound as if it's impossible for errors to be returned if we just did everything correctly, and I don't think that's the case. It's probably better if clients don't see the bulk of these errors anyway :) a bit like how servers just return `InternalServerError` as a catch-all for a failure on their part.

src/providers/tpm_provider/utils.rs
```diff
@@ -15,127 +15,43 @@
...
-                    Tss2ResponseCodeKind::Type => ResponseStatus::PsaErrorInvalidArgument,
-                    Tss2ResponseCodeKind::Handle => ResponseStatus::PsaErrorInvalidHandle,
-                    Tss2ResponseCodeKind::Kdf => ResponseStatus::PsaErrorNotSupported,
-                    Tss2ResponseCodeKind::Range => ResponseStatus::PsaErrorInvalidArgument,
-                    Tss2ResponseCodeKind::AuthFail => ResponseStatus::PsaErrorNotPermitted,
-                    Tss2ResponseCodeKind::Nonce => ResponseStatus::PsaErrorInvalidArgument,
-                    Tss2ResponseCodeKind::Pp => ResponseStatus::PsaErrorBadState,
-                    Tss2ResponseCodeKind::Scheme => ResponseStatus::PsaErrorNotSupported,
-                    Tss2ResponseCodeKind::Size => ResponseStatus::PsaErrorInvalidArgument,
-                    Tss2ResponseCodeKind::Symmetric => ResponseStatus::PsaErrorNotSupported,
```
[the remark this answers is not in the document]
ionut-arm: The same applies to the PKCS11 stuff - I don't even know if for that one there's a way for us to verify what is supported and what isn't.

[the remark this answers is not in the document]
hug-dev: Agreed, will let those one go through.

src/providers/pkcs11_provider/utils.rs
hug-dev: Will moderate this statement to be more humble.

src/providers/pkcs11_provider/utils.rs
```diff
@@ -36,106 +41,25 @@ pub fn to_response_status(error: Error) -> ResponseStatus {
...
-        CKR_PIN_TOO_WEAK => ResponseStatus::PsaErrorInvalidArgument,
-        CKR_PUBLIC_KEY_INVALID => ResponseStatus::PsaErrorInvalidArgument,
-        CKR_FUNCTION_REJECTED => ResponseStatus::PsaErrorGenericError,
-        CKR_VENDOR_DEFINED => ResponseStatus::PsaErrorGenericError,
+        s @ CKR_CURVE_NOT_SUPPORTED
+        | s @ CKR_DOMAIN_PARAMS_INVALID
+        | s @ CKR_FUNCTION_NOT_SUPPORTED => {
+            error!("Not supported value ({:?})", s);
+            ResponseStatus::PsaErrorNotSupported
+        }
```
ionut-arm, line 59: 💯

PR #144, opened 2020-04-16
Add tests for operations edge cases

### LINKED ISSUE #139: Key should be deleted from the KIM if generation/import fails
If a key is generated or imported but execution fails during generation/import, the key should not be present in the Key ID Manager anymore.
An integration test should be created out of this.

hug-dev:
Adds many tests to make sure that the operations API is respected. Fixes
in the service the parts that are not compliant.
Also fixes #139

### DIFF
--- a/src/providers/mbed_provider/utils.rs
+++ b/src/providers/mbed_provider/utils.rs
@@ -114,11 +114,11 @@ pub fn convert_key_usage(operation: &psa_key_attributes::UsageFlags) -> psa_key_
         usage |= PSA_KEY_USAGE_EXPORT;
     }
 
-    if operation.sign_message || operation.sign_hash {
+    if operation.sign_message && operation.sign_hash {
         usage |= PSA_KEY_USAGE_SIGN;
     }
 
-    if operation.verify_message || operation.verify_hash {
+    if operation.verify_message && operation.verify_hash {
         usage |= PSA_KEY_USAGE_VERIFY;
     }
 
--- a/tests/all_providers/mod.rs
+++ b/tests/all_providers/mod.rs
@@ -13,29 +13,71 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 use parsec_client_test::TestClient;
-use parsec_interface::requests::ProviderID;
 use parsec_interface::requests::Result;
+use parsec_interface::requests::{Opcode, ProviderID};
 use std::collections::HashSet;
+use uuid::Uuid;
 
 #[test]
 fn list_providers() {
     let mut client = TestClient::new();
     let providers = client.list_providers().expect("list providers failed");
     assert_eq!(providers.len(), 4);
-    let ids: HashSet<ProviderID> = providers.iter().map(|p| p.id).collect();
-    assert!(ids.contains(&ProviderID::Core));
-    assert!(ids.contains(&ProviderID::MbedCrypto));
-    assert!(ids.contains(&ProviderID::Pkcs11));
-    assert!(ids.contains(&ProviderID::Tpm));
+    let uuids: HashSet<Uuid> = providers.iter().map(|p| p.uuid).collect();
+    // Core provider
+    assert!(uuids.contains(&Uuid::parse_str("47049873-2a43-4845-9d72-831eab668784").unwrap()));
+    // Mbed Crypto provider
+    assert!(uuids.contains(&Uuid::parse_str("1c1139dc-ad7c-47dc-ad6b-db6fdb466552").unwrap()));
+    // PKCS 11 provider
+    assert!(uuids.contains(&Uuid::parse_str("30e39502-eba6-4d60-a4af-c518b7f5e38f").unwrap()));
+    // TPM provider
+    assert!(uuids.contains(&Uuid::parse_str("1e4954a4-ff21-46d3-ab0c-661eeb667e1d").unwrap()));
 }
 
 #[test]
+// ListOpcodes is not currently supported only by the Core Provider, see
+// parallaxsecond/parsec-operations#9
+#[ignore]
 fn list_opcodes() {
     let mut client = TestClient::new();
-    let opcodes = client
-        .list_opcodes(ProviderID::MbedCrypto)
-        .expect("list providers failed");
-    assert_eq!(opcodes.len(), 7);
+    let mut crypto_providers_opcodes = HashSet::new();
+    let mut core_provider_opcodes = HashSet::new();
+
+    let _ = crypto_providers_opcodes.insert(Opcode::PsaGenerateKey);
+    let _ = crypto_providers_opcodes.insert(Opcode::PsaDestroyKey);
+    let _ = crypto_providers_opcodes.insert(Opcode::PsaSignHash);
+    let _ = crypto_providers_opcodes.insert(Opcode::PsaVerifyHash);
+    let _ = crypto_providers_opcodes.insert(Opcode::PsaImportKey);
+    let _ = crypto_providers_opcodes.insert(Opcode::PsaExportPublicKey);
+
+    let _ = core_provider_opcodes.insert(Opcode::Ping);
+    let _ = core_provider_opcodes.insert(Opcode::ListProviders);
+    let _ = core_provider_opcodes.insert(Opcode::ListOpcodes);
+
+    assert_eq!(
+        client
+            .list_opcodes(ProviderID::Core)
+            .expect("list providers failed"),
+        core_provider_opcodes
+    );
+    assert_eq!(
+        client
+            .list_opcodes(ProviderID::Tpm)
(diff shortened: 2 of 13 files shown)


tests/per_provider/normal_tests/export_public_key.rs
```diff
@@ -53,16 +66,71 @@ fn import_and_export_public_key() -> Result<()> {
...
+    client.generate_rsa_sign_key(key_name.clone())?;
+    let public_key = client.export_public_key(key_name)?;
+
+    // That should not fail if the bytes are in the expected format.
+    let _public_key: RsaPublicKey = picky_asn1_der::from_bytes(&public_key).unwrap();
+    Ok(())
+}
+
+#[test]
+fn check_public_rsa_export_format2() -> Result<()> {
```
ionut-arm, line 89: I'm not sure I understand what the difference is between this test and the one above

tests/per_provider/normal_tests/import_key.rs
```diff
@@ -69,25 +82,162 @@ fn import_key_twice() -> Result<()> {
...
+        .unwrap_err();
+
+    assert_eq!(status, ResponseStatus::PsaErrorInvalidArgument);
+
+    Ok(())
+}
+
+#[test]
+fn failed_imported_key_should_be_removed() -> Result<()> {
+    // If the key_bits field of the key attributes is zero, the operation should still work.
```
ionut-arm, line 201: Copy-pasta?

tests/per_provider/normal_tests/import_key.rs
```diff
@@ -69,25 +82,162 @@ fn import_key_twice() -> Result<()> {
...
-            KEY_DATA.to_vec(),
-        )
-        .expect_err("The key with the same name has already been created.");
-    assert_eq!(status, ResponseStatus::PsaErrorAlreadyExists);
+        },
+    };
+
+    if client
+        .import_key(key_name.clone(), attributes, Vec::new())
+        .is_err()
```
ionut-arm, line 236: This is kinda wrong - if `import_key` isn't an error, the test still succeeds. You should probably `?` it and just carry on with the next import.

ionut-arm: There are others like this - `failed_created_key_should_be_removed` is one example.

tests/per_provider/normal_tests/export_public_key.rs
```diff
@@ -53,16 +66,71 @@ fn import_and_export_public_key() -> Result<()> {
...
+                hash_alg: Hash::Sha256,
+            }),
+        },
+    };
+
+    client.generate_key(key_name.clone(), key_attributes)?;
+
+    let _public_key = client.export_public_key(key_name)?;
+
+    Ok(())
```
ionut-arm, line 135: you could set the return value of the test to `Result<Vec<u8>>` or whatever `export_public_key` returns, saves you of some needless code!

tests/per_provider/normal_tests/create_destroy_key.rs
```diff
@@ -80,3 +97,56 @@ fn create_destroy_twice() -> Result<()> {
     client.destroy_key(key_name)?;
     client.destroy_key(key_name_2)
 }
+
```
ionut-arm, line 100: Do we have a test that a short RSA key, say, 512 bits, isn't allowed?

src/providers/mbed_provider/utils.rs
```diff
@@ -114,11 +114,11 @@ pub fn convert_key_usage(operation: &psa_key_attributes::UsageFlags) -> psa_key_
         usage |= PSA_KEY_USAGE_EXPORT;
     }
 
-    if operation.sign_message || operation.sign_hash {
+    if operation.sign_message && operation.sign_hash {
         usage |= PSA_KEY_USAGE_SIGN;
```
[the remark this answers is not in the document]
hug-dev: Yes exactly, there isn't the difference between `hash` and `message` in the one we use.

tests/all_providers/mod.rs
```diff
@@ -13,29 +13,71 @@
...
+    assert!(uuids.contains(&Uuid::parse_str("30e39502-eba6-4d60-a4af-c518b7f5e38f").unwrap()));
+    // TPM provider
+    assert!(uuids.contains(&Uuid::parse_str("1e4954a4-ff21-46d3-ab0c-661eeb667e1d").unwrap()));
 }
 
 #[test]
+// ListOpcodes is not currently supported only by the Core Provider, see
+// parallaxsecond/parsec-operations#9
+#[ignore]
 fn list_opcodes() {
```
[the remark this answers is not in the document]
hug-dev: This could be a good idea in the configuration, yes! What do you think would be the use case?

tests/per_provider/normal_tests/asym_sign_verify.rs
```diff
@@ -115,15 +116,204 @@ fn only_verify_from_internet() -> Result<()> {
...
+    let mut hasher = Sha256::new();
+    hasher.input(b"Bob wrote this message.");
+    let hash = hasher.result().to_vec();
+
+    client.generate_rsa_sign_key(key_name.clone())?;
+
+    let _ = client.sign_with_rsa_sha256(key_name, hash)?;
+
+    Ok(())
+}
```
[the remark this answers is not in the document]
hug-dev: That's definitely a good idea for a new test. We can create an issue for that one!

tests/per_provider/normal_tests/export_public_key.rs
hug-dev: Oh, my bad! In this one test I wanted to use `generate_key` with the key type set to `RsaPublicKey` to check that it exports a key with the good format for both `RsaKeyPair` and `RsaPublicKey`. will update.

tests/per_provider/normal_tests/import_key.rs
hug-dev: Oops! Will remove those comments.

tests/per_provider/normal_tests/import_key.rs
hug-dev: I did it like this because the way I trigger an error here is by giving a key type of `Aes` which is not currently supported. However, we might not want this test to fail the moment `Aes` is supported, or does not trigger an error.

tests/per_provider/normal_tests/export_public_key.rs
hug-dev: True. I did intentionally not write good quality code as this is a test 😛

tests/per_provider/normal_tests/import_key.rs
ionut-arm: > However, we might not want this test to fail the moment Aes is supported

The import will fail when AES is supported anyway, because AES keys have at most 256 bits

> or does not trigger an error.

I'd say it should fail, because otherwise what's the point of the test? You want it to fail on the first import to show that you can import again succesfully after - imagine if we do something which leads to the first import succeeding, but now if you fail an import you can't reuse the name - we wouldn't be testing for that anymore so we wouldn't know.

tests/per_provider/normal_tests/create_destroy_key.rs
hug-dev: I looked the API for `psa_generate_key`, I don't think it limits a specific size.

tests/per_provider/normal_tests/import_key.rs
hug-dev: > The import will fail when AES is supported anyway, because AES keys have at most 256 bits

Good point!
I think I confused myself anyway: this test works in two steps:
1. We **intentionally** make an operation that will fail when generating/importing a key.
2. We try to generate/import a key with the same name to check that the key from 1 has been destroyed.

If I used `?` instead of `is_err`, on the first import, then the test will fail without us actually testing anything 😋

tests/per_provider/normal_tests/create_destroy_key.rs
ionut-arm: Indeed. We might want to ask them for a suggestion for a minimum, for potentially marking `Deprecated` sizes in the future - doesn't make much sense to me for us to mark small elliptic curves as deprecated because of size but not doing the same for RSA - but maybe there's a reason for that!

tests/per_provider/normal_tests/import_key.rs
ionut-arm: > If I used ? instead of is_err, on the first import, then the test will fail without us actually testing anything 😋

hug-dev: Yes good point, so we also check that the first one is definitely an error.

hug-dev:
Adresses the comments in a new commit for easier review.

Issue #116, opened 2020-02-27
PKCS 11 provider stress tests sometimes fail


hug-dev:
This issue only happens once in a while on the Travis CI (Arm64 tests) only.
[Logs](https://api.travis-ci.com/v3/job/291713886/log.txt) of an example of failed test.
The configuration is: `DOCKER_IMAGE_NAME=pkcs11-provider DOCKER_IMAGE_PATH=tests/per_provider/provider_cfg/pkcs11 SCRIPT="tests/ci.sh pkcs11"`

Relevent part of the logs:
```
Start Parsec for stress tests
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/parsec --config tests/per_provider/provider_cfg/pkcs11/config.toml`
Execute stress tests
    Finished test [unoptimized + debuginfo] target(s) in 0.18s
     Running target/debug/deps/parsec_service-9799ccc9d5a72007

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 12 filtered out

     Running target/debug/deps/parsec-8c1e7cba57c8e590
     Running target/debug/deps/mod-c4e7c83f5440ad54

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out


running 1 test
thread '<unnamed>' panicked at 'Failed to import key: ConnectionError', src/libcore/result.rs:1188:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:84
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:61
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1025
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1426
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:65
   7: std::sys_common::backtrace::print
[...]

PR #145, opened 2020-04-16
Remove stress test on Travis CI for PKCS 11

hug-dev:
Adds a command line option on the CI script to be able to not execute
the stress tests.
On Travis CI, the stress test fails for the PKCS 11 provider, see #116.
This commit disables them until the reason is found.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,12 +10,14 @@ jobs:
       env:
       - DOCKER_IMAGE_NAME=mbed-crypto-provider
       - DOCKER_IMAGE_PATH=tests/per_provider/provider_cfg/mbed-crypto
-      - SCRIPT="tests/ci.sh mbed-crypto"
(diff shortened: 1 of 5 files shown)


ionut-arm:
Does travis not pick up the changes from the branch? it seems to have failed

hug-dev:
> 
> 
> Does travis not pick up the changes from the branch? it seems to have failed

Yes you are right, it was failing during the persistence tests for the PKCS 11 provider (Broken pipe) and the stress tests for the Mbed Crypto provider.
I am curious to see if this problem is because the timeout on the service side for the tests is too small. I am trying to put it to 5 seconds to match the timeout on the test client.

hug-dev:
When trying to write the response on the stream to the client, the service sometimes has a `Broken Pipe` error. This might be due because, for some reason, the client disconnects from the socket.

PR #150, opened 2020-04-20
Move test client back in the Parsec repo

ionut-arm:
This commit moves the test client back in the main Parsec repo and moves
it on top of the `BasicClient` - the main core client, thus making the
most out of functionality re-use.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ name = "parsec"
 path = "src/bin/main.rs"
 
 [dependencies]
-parsec-interface = "0.12.0"
+parsec-interface = "0.13.0"
 rand = "0.7.2"
 base64 = "0.10.1"
 uuid = "0.7.4"
@@ -40,12 +40,13 @@ derivative = "1.0.3"
 version = "3.0.0"
 
 [dev-dependencies]
-parsec-client-test = { git = "https://github.com/parallaxsecond/parsec-client-test", tag = "0.3.0" }
 num_cpus = "1.10.1"
 picky-asn1-der = "0.2.2"
 picky-asn1 = "0.2.1"
 serde = { version = "1.0", features = ["derive"] }
 sha2 = "0.8.1"
+parsec-client = "0.1.0"
+parsec-interface = { version = "0.13.0", features = ["testing"] }
 
 [build-dependencies]
 bindgen = "0.50.0"
--- a/README.md
+++ b/README.md
@@ -73,25 +73,7 @@ Contributions from the developer community are welcome. Please refer to the cont
 
 # Example
 
-Launch the Parsec service with a single software-based provider (using the default configuration):
-```bash
-$ git clone https://github.com/parallaxsecond/parsec.git
-$ cd parsec
-$ RUST_LOG=info cargo run
-```
-
-Parsec Client Libraries can now communicate with the service. For example using the Rust Test client,
-RSA signatures can be done as follows:
-```rust
-use parsec_client_test::TestClient;
-
-let mut client = TestClient::new();
-let key_name = String::from("🔑 What shall I sign? 🔑");
-client.generate_rsa_sign_key(key_name.clone()).unwrap();
-let signature = client.sign(key_name,
-                            String::from("Platform AbstRaction for SECurity").into_bytes())
-                      .unwrap();
-```
+For examples of how to access PARSEC as a client application, check [this Rust client documentation](https://docs.rs/parsec-client/*/parsec_client/core/basic_client/struct.BasicClient.html).
 
 Check the [**user**](https://parallaxsecond.github.io/parsec-book/parsec_users.html), [**client developer**](https://parallaxsecond.github.io/parsec-book/parsec_client/index.html) and [**service developer**](https://parallaxsecond.github.io/parsec-book/parsec_service/index.html) guides for more information on building, installing, testing and using Parsec!
 
--- a/tests/per_provider/normal_tests/ping.rs
+++ b/tests/per_provider/normal_tests/ping.rs
@@ -12,17 +12,18 @@
 // WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-use parsec_client_test::RequestTestClient;
-use parsec_client_test::TestClient;
+use crate::test_clients::RequestClient;
+use crate::test_clients::TestClient;
 use parsec_interface::requests::request::{Request, RequestAuth, RequestBody};
 use parsec_interface::requests::Opcode;
 use parsec_interface::requests::ProviderID;
-use parsec_interface::requests::{ResponseStatus, Result};
+use parsec_interface::requests::ResponseStatus;
+use parsec_interface::requests::Result;
 
 #[test]
 fn test_ping() -> Result<()> {
     let mut client = TestClient::new();
-    let version = client.ping(ProviderID::Core)?;
+    let version = client.ping()?;
     assert_eq!(version.0, 1);
     assert_eq!(version.1, 0);
 
@@ -31,15 +32,16 @@ fn test_ping() -> Result<()> {
 
 #[test]
 fn mangled_ping() {
-    let mut client = RequestTestClient::new();
+    let client = RequestClient::default();
     let mut req = Request::new();
-    req.header.version_maj = 1;
     req.header.provider = ProviderID::Core;
     req.header.opcode = Opcode::Ping;
     req.auth = RequestAuth::from_bytes(Vec::from("root"));
 
     req.body = RequestBody::_from_bytes(vec![0x11, 0x22, 0x33, 0x44, 0x55]);
 
-    let resp = client.send_request(req).expect("Failed to read Response");
+    let resp = client
+        .process_request(req)
+        .expect("Failed to read Response");
     assert_eq!(resp.header.status, ResponseStatus::DeserializingBodyFailed);
 }
--- a/tests/per_provider/persistent_after.rs
+++ b/tests/per_provider/persistent_after.rs
@@ -15,8 +15,9 @@
 
 // These functions test for the service persistency to shutdown. They will be executed after the
 // service is shutdown, after the persistent_before tests are executed.
-use parsec_client_test::TestClient;
-use parsec_interface::requests::{Opcode, ProviderID, ResponseStatus, Result};
+use crate::test_clients::TestClient;
+use parsec_interface::requests::Result;
+use parsec_interface::requests::{ProviderID, ResponseStatus};
 
 const HASH: [u8; 32] = [
     0x69, 0x3E, 0xDB, 0x1B, 0x22, 0x79, 0x03, 0xF4, 0xC0, 0xBF, 0xD6, 0x91, 0x76, 0x37, 0x84, 0xA2,
@@ -39,7 +40,7 @@ fn reuse_to_sign() -> Result<()> {
 fn should_have_been_deleted() {
     let mut client = TestClient::new();
 
-    if client.get_cached_provider(Opcode::PsaDestroyKey) == ProviderID::Tpm {
+    if client.provider().unwrap() == ProviderID::Tpm {
         // This test does not make sense for the TPM Provider.
         return;
     }
--- a/tests/test_clients/mod.rs
+++ b/tests/test_clients/mod.rs
@@ -1,0 +1,318 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+pub mod raw_request;
+pub mod stress;
+
+pub use raw_request::RawRequestClient;
+
+pub use parsec_client::core::request_client::RequestClient;
+pub use parsec_client::error;
+
+use log::error;
+use parsec_client::auth::AuthenticationData;
+use parsec_client::core::basic_client::BasicClient;
+use parsec_client::error::Error;
+use parsec_interface::operations::list_providers::ProviderInfo;
+use parsec_interface::operations::psa_algorithm::{Algorithm, AsymmetricSignature, Hash};
+use parsec_interface::operations::psa_key_attributes::*;
+use parsec_interface::operations::psa_key_attributes::{KeyAttributes, KeyPolicy, UsageFlags};
+use parsec_interface::requests::{Opcode, ProviderID, ResponseStatus, Result};
+use std::collections::HashSet;
+use std::time::Duration;
+
+/// Client structure automatically choosing a provider and high-level operation functions.
+#[derive(Debug)]
+pub struct TestClient {
+    basic_client: BasicClient,
+    created_keys: Option<HashSet<(String, String, ProviderID)>>,
+}
+
+fn convert_error(err: Error) -> ResponseStatus {
+    if let Error::Service(resp_status) = err {
+        resp_status
+    } else {
+        panic!(
+            "Expected to obtain a service error, but got a client error instead: {:?}",
+            err
+        );
+    }
+}
+
+impl TestClient {
+    /// Creates a TestClient instance.
+    ///
+    /// The implicit provider chosen for servicing cryptographic operations is decided through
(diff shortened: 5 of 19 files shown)


Cargo.toml
```diff
@@ -40,12 +40,13 @@ derivative = "1.0.3"
 version = "3.0.0"
 
 [dev-dependencies]
-parsec-client-test = { git = "https://github.com/parallaxsecond/parsec-client-test", tag = "0.3.0" }
 num_cpus = "1.10.1"
 picky-asn1-der = "0.2.2"
 picky-asn1 = "0.2.1"
 serde = { version = "1.0", features = ["derive"] }
 sha2 = "0.8.1"
+parsec-client = { git = "https://github.com/parallaxsecond/parsec-client-rust" }
```
hug-dev, line 48: Let's publish it as a crate! I think it would make sense to do it now and pull it as a normal crates.io dependence! To get rid of the different versions in dependences issue.

tests/per_provider/normal_tests/ping.rs
```diff
@@ -12,17 +12,18 @@
...
 use parsec_interface::requests::Opcode;
 use parsec_interface::requests::ProviderID;
-use parsec_interface::requests::{ResponseStatus, Result};
+use parsec_interface::requests::ResponseStatus;
 
 #[test]
 fn test_ping() -> Result<()> {
     let mut client = TestClient::new();
-    let version = client.ping(ProviderID::Core)?;
+    let version = client.ping()?;
```
hug-dev, line 26: So much nicer to not have to put the redundant `ProviderID::Core`

tests/test_client/abstract_client.rs
```diff
@@ -0,0 +1,329 @@
+// Copyright 2019 Contributors to the Parsec project.
```
hug-dev, line 1: 2020 I think 😁

tests/test_client/abstract_client.rs
```diff
@@ -0,0 +1,329 @@
...
+use parsec_interface::operations::list_providers::ProviderInfo;
+use parsec_interface::operations::psa_algorithm::{Algorithm, AsymmetricSignature, Hash};
+use parsec_interface::operations::psa_key_attributes::*;
+use parsec_interface::operations::psa_key_attributes::{KeyAttributes, KeyPolicy, UsageFlags};
+use parsec_interface::requests::{Opcode, ProviderID};
+use std::collections::{HashMap, HashSet};
+
+/// Client structure automatically choosing a provider and high-level operation functions.
+#[derive(Debug)]
+pub struct TestClient {
```
hug-dev, line 16: About naming, I am thinking: this structure `TestClient` in `test_client/abstract_client.rs` does not feel totally right. Ideally we should keep the same name of the main structure of a file and its name. Any idea?

Suggestion:
* have the folder `tests/test_clients` (note the **ss**)
* inside have `Abstract`, `Stress` and `Raw` structures under `abstract.rs`, `stress.rs` and `raw.rs`. Dropping `test_client` as it is in the upper module's name 💇‍♂️

tests/test_client/abstract_client.rs
```diff
@@ -0,0 +1,329 @@
...
+    op_client: BasicClient,
+    cached_opcodes: Option<HashMap<ProviderID, HashSet<Opcode>>>,
```
hug-dev, line 18: Is caching opcodes still helpful for our tests? We might only need to know **one** crypto provider ID.

tests/test_client/abstract_client.rs
```diff
@@ -0,0 +1,329 @@
...
+    provider: Option<ProviderID>,
```
hug-dev, line 19: Could this be the same provider as the one in the basic client?

tests/test_client/abstract_client.rs
```diff
@@ -0,0 +1,329 @@
...
+    auth: AuthenticationData,
```
hug-dev, line 20: Same question here

Cargo.toml
ionut-arm: I'm sorta relying on the CI to tell me if the client actually works 😅  if it does, will replace!

tests/test_client/abstract_client.rs
ionut-arm: Ok, I wasn't sure because I thought I saw you changing some to `2019`

Cargo.toml
hug-dev: I see, makes sense!

tests/test_client/abstract_client.rs
hug-dev: I should have changed all ranges `2019 - 2020` to `2019` but kept `2020` to `2020`. If not I made a mistake!

tests/test_client/abstract_client.rs
ionut-arm: > We might only need to know one crypto provider ID.

And isn't that a very good example where provider discovery happens? 😸 You have only one provider in the system so the test client finds it for you.

tests/test_client/abstract_client.rs
ionut-arm: I think this one makes sense to remove

tests/test_client/abstract_client.rs
hug-dev: Yes sure, the provider discovery is an excellent idea! I meant since we are doing all requests to one single provider, we might not need to cache this `HashMap` but just be able to do the discovery in the builder and use that as the `provider` field of the BasicClient?

tests/test_client/abstract_client.rs
ionut-arm: I was thinking about that, but given how the whole provider discovery thing works, this is the way to force a test to use one specific provider, because the flow goes like this:
* you ask for an operation to be executed, say, generate key
* the client checks to see if you hardcoded a value (this `Option<ProviderID>`)
* if yes, it uses that
* if no, it checks the cache for a provider that it can use

Most of our tests are provider-independent so would have to include some sort of "active provider" discovery mechanism. But it's also useful to have the option to force the client to use a particular provider. This can't be done with the `implicit_provider` of the underlying `BasicClient` because that one isn't optional, so you wouldn't know whether it was hardcoded by the test or if it was there from the previous request.

ionut-arm: Also, I just realised - do we actually have any tests where we attempt to make a request against a provider that isn't active?

hug-dev: > this is the way to force a test to use one specific provider

I do not think we are actually using that feature! The only occurence of `set_provider` in our tests is in `basic::wrong_provider_code` to test that a crypto operation on core fails. That wouldn't be possible at the `BasicClient` level anyway because of the new restrictions.

That could be useful in the `all_providers` tests, if you want to execute requests on multiple providers.

> This can't be done with the implicit_provider of the underlying BasicClient because that one isn't optional, so you wouldn't know whether it was hardcoded by the test or if it was there from the previous request.

Ok I think I get you, you can change the provider ID just for one request and then set it to `None` again so that it uses the one originally selected by the discovery.

Can you do that with a method `replace_provider` that replace the current `implicit_provider` with the one given and returns the old one? That way this logic could be done in the test itself when and if it is needed in the future.

The reason I am asking is that if we remove this logic, we can actually use in all of our tests the `BasicClient` methods directly for most operations and not have this wrapper in the test client. That would make the code much simpler I think.

hug-dev: > Also, I just realised - do we actually have any tests where we attempt to make a request against a provider that isn't active?

I don't think so, and it would be good to add!

tests/test_client/abstract_client.rs
ionut-arm: `abstract` is an identifier.

Also, I'm not sure about having the clients named `Abstract`, `Stress` etc. - while it is true that they reside within a `test_clients` module, it's confusing to take `Client` part out of the structure name, as when you want to use it somewhere you shouldn't need to do `test_clients::Stress`; even this looks dubious to me - what's `Stress` exactly? I think the whole idea of namespacing works and is helpful in some circumstances, but is a bit over the top in others. 

I'll rename the modules within `test_clients`, but will keep the `Client` part of the struct name.

README.md
```diff
@@ -80,17 +80,47 @@ $ cd parsec
...
+        },
+        key_algorithm: Algorithm::AsymmetricSignature(sign_alg),
+    },
+};
 let key_name = String::from("🔑 What shall I sign? 🔑");
-client.generate_rsa_sign_key(key_name.clone()).unwrap();
-let signature = client.sign(key_name,
-                            String::from("Platform AbstRaction for SECurity").into_bytes())
-                      .unwrap();
+client.psa_generate_key(key_name.clone(), key_attrs).unwrap();
```
hug-dev, line 118: These key name cloning are so annoying :/ I am wondering if we should make the functions take a reference and clone when creating the requests. If we do I am not sure if that should be part of this layer or above.

tests/per_provider/provider_cfg/tpm/config.toml
```diff
@@ -16,4 +16,4 @@ manager_type = "OnDisk"
 provider_type = "Tpm"
 key_id_manager = "on-disk-manager"
 tcti = "mssim"
-owner_hierarchy_auth = "tpm_pass"
+owner_hierarchy_auth = ""
```
hug-dev, line 19: Is that not good anymore?

tests/test_clients/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+pub struct TestClient {
+    basic_client: BasicClient,
+    created_keys: Option<HashSet<(String, String, ProviderID)>>,
+}
+
+impl TestClient {
+    /// Creates a TestClient instance.
+    ///
+    /// The implicit provider chosen for servicing cryptographic operations is decided through
+    /// a call to `list_providers`, followed by choosing the first non-Core provider.
```
hug-dev, line 33: Nice!

tests/test_clients/mod.rs
```diff
@@ -0,0 +1,305 @@
...
+    pub fn new() -> TestClient {
+        let mut client = TestClient {
+            basic_client: BasicClient::new(
+                AuthenticationData::AppIdentity(String::from("root")),
+                ProviderID::Core,
+            ),
```
hug-dev, line 39: This constructor feels a bit like the chicken and the egg: you need an implicit provider for non-core operations but you can only do that with `list_providers` and you need the basic client instance for that.
That is a client change I did not notice before but what do you think if we either take the following decisions:
1. put the first-crypto provider discovery inside the `BasicClient` constructor. Not so basic now...
2. remove the provider argument from the constructor and always but `Core`, with doc comments saying that `list_providers` is the first thing expected.
3. remove the provider argument and put an `Option<ProviderID>` in the `BasicCLient` structure, initialised at `None` by the constructor. We also create a new client error which is throwm when no `implicit_provider` has been selected.

I personally like 3 best.

hug-dev: Maybe we should continue this in the appropriate GitHub issue

tests/test_clients/raw_request.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright 2019 Contributors to the Parsec project.
```
hug-dev, line 1: I think those copyrights as well should be 2020

tests/per_provider/persistent_after.rs
```diff
@@ -51,6 +52,6 @@ fn should_have_been_deleted() {
         client
             .destroy_key(key_name)
             .expect_err("This key should have been destroyed."),
-        ResponseStatus::PsaErrorDoesNotExist
+        Error::Service(ResponseStatus::PsaErrorDoesNotExist)
```
hug-dev, line 55: If you want, to avoid having to wrap all `ResponseStatus` in `Error::Service`, you could make the test client return `Result<_, ResponseStatus>` panicking if the error is not `Error::Service`.
I think all of the test client requests should go to the service and back, otherwise we are not really testing the service.

README.md
ionut-arm: I know... we can either do that or make a Rust type for key names that we implement `Copy` for

tests/per_provider/provider_cfg/tpm/config.toml
ionut-arm: 😅 my mistake, I changed that for local testing, but forgot to reset it

tests/test_clients/mod.rs
ionut-arm: We can try and settle it here, as it's not entirely clear if it's something lacking in the `BasicClient` or here.

3 makes sense, but I'm guessing you wouldn't want to change the `implicit_provider` back to `None`. I think that matching could be done efficiently, so I can do that as another PR in the client.

tests/per_provider/persistent_after.rs
ionut-arm: I didn't mind it that much, but I can change it to unwrap the value before returning

tests/test_clients/mod.rs
hug-dev: I guess if `implicit_provider` is private, you would only change it to `Some(ProviderID::...)` through methods. Also checking that it's not changing it to `Core`.
One more suggestion to make it even more clear in the client: renaming `implicit_provider` to `implicit_crypto_provider` which what it really is!

Those things give me the feeling that there would be even more reasons to separate our structures (opcodes, provide trait, provider ID, response status) between Crypto (PSA) and Core ones.

ionut-arm: > One more suggestion to make it even more clear in the client: renaming implicit_provider to implicit_crypto_provider which what it really is!

ionut-arm: I think I'll keep it as `implicit_provider` - there will be operations that are not explicitly crypto but which need a provider defined (the smart defaulting stuff) - and while they're not available now, it doesn't hurt to have them ready. As you were saying, I'd rather give the user an option to set it to whatever they wish, regardless of whether it is allowed for currently supported operations. It feels like more of a problem with our `ProviderID` definition than with this field

hug-dev: > there will be operations that are not explicitly crypto

Those operations will still be aimed at a a crypto provider though, right? They can still change it with the `set_crypto_provider_id` methods I think.
My suggestion was mainly about removing the provider ID argument for the `new` method.

hug-dev: I would be happy with no renaming at all but just the `implicit_provider` being an `Option<ProviderID>`, the `new` method setting it at `None` and crypto methods returning a new error if it is `None`.
No need to make more checks in the `set_provider` method for now.

README.md
```diff
@@ -80,17 +80,47 @@ $ cd parsec
 $ RUST_LOG=info cargo run
 ```
 
-Parsec Client Libraries can now communicate with the service. For example using the Rust Test client,
+Parsec Client Libraries can now communicate with the service. For example using the Rust client,
```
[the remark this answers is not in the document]
ionut-arm: Uhm... Now that we moved the provider out of the constructor for the client and we advise users to go through `list_providers`, `list_opcodes` first... I feel like this example is becoming quite heavy... it went from like 10 lines to probably 40-50 if I do the discovery part too... What do? Link to the client documentation once that's out?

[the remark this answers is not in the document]
hug-dev: That's probably the best thing to do! Even more so because the example we would put here would not be checked for compilation often and might not even work at times 😬

Cargo.toml
```diff
@@ -40,12 +40,13 @@ derivative = "1.0.3"
 version = "3.0.0"
 
 [dev-dependencies]
-parsec-client-test = { git = "https://github.com/parallaxsecond/parsec-client-test", tag = "0.3.0" }
 num_cpus = "1.10.1"
 picky-asn1-der = "0.2.2"
 picky-asn1 = "0.2.1"
 serde = { version = "1.0", features = ["derive"] }
 sha2 = "0.8.1"
+parsec-client = "0.1.0"
```
hug-dev, line 48: 👏

PR #159, opened 2020-04-24
Update partners file with web links and logos

paulhowardarm:
All logos used with permission.

Signed-off-by: Paul Howard <paul.howard@arm.com>


paulhowardarm:
Thanks @hug-dev for the reminder on the old logo. I've pushed a second commit to do that.

hug-dev:
Thank you!

PR #160, opened 2020-04-27
Move end to end tests to own crate

### LINKED ISSUE #155: Integration tests should be isolated in their crate
The tests we put in `tests` are executed against the `parsec` binary, through the Unix socket. Cargo however uses `tests` for integration tests of the library crate, ie the modules made public in `src/lib.rs`.

We should keep `tests/` for the integration tests we possibly want to do for the modules we export in the library crate and move our tests in another crate.

That crate could be at the root of this repository and contain:

* in `src`: the different test clients
* in `tests`: our integration tests

The `ci.sh` script will have to be modified as well.

ionut-arm:
This commit moves the PARSEC end-to-end tests into their own crate and
refactors the test script along with it.

Fixes #155


e2e_tests/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+# This crate should NOT be upstreamed to crates.io; it is only intended for 
+# PARSEC testing purposes (both locally and on CI)
+[package]
```
hug-dev, line 3: You can add [this](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field):
```toml
[package]
# ...
publish = false
```
to enforce it won't be done by mistake!

e2e_tests/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+license = "Apache-2.0"
+edition = "2018"
+
+[dependencies]
+num_cpus = "1.13.0"
+picky-asn1-der = "0.2.2"
+picky-asn1 = "0.2.1"
+serde = { version = "1.0", features = ["derive"] }
+sha2 = "0.8.1"
+parsec-client = { version = "0.2.0", features = ["testing"] }
```
hug-dev, line 20: 👏 Nice not to have to add the interface next to that!

tests/ci.sh
```diff
@@ -63,12 +63,11 @@ while [ "$#" -gt 0 ]; do
                 error_msg "Only one provider name must be given"
             fi
             PROVIDER_NAME=$1
+            CONFIG_PATH="e2e_tests/provider_cfg/$1/config.toml"
```
hug-dev, line 66: 🤯

e2e_tests/Cargo.toml
ionut-arm: Done!

PR #162, opened 2020-04-28
Add integration test

ionut-arm:
This commit adds an integration tests and updates the `ci.sh` script to
reflect the recent changes in testing framework.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -31,7 +31,7 @@ It is meant to be executed inside one of the container
 which Dockerfiles are in tests/per_provider/provider_cfg/*/
 or tests/all_providers/
 
-Usage: ./tests/ci.sh [--no-cargo-clean] [--no-stress-test] PROVIDER_NAME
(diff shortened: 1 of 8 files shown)


ci.sh
```diff
@@ -110,12 +110,10 @@ if rustup component list | grep -q clippy; then
     cargo clippy --all-targets $FEATURES -- -D clippy::all -D clippy::cargo
 fi
 
-echo "Unit tests"
-RUST_BACKTRACE=1 cargo test --lib $FEATURES
-echo "Doc tests"
-RUST_BACKTRACE=1 cargo test --doc $FEATURES
+echo "Unit, doc and integration tests"
+RUST_BACKTRACE=1 cargo test $FEATURES
```
hug-dev, line 114: I am wondering if there will be any problem with the `mappings` folders created in the integration tests, if it ever clashes with the ones in the end-to-end tests. Maybe it would be nice to rename them `PROVIDER_mappings` and/or delete them at the end?

ionut-arm: makes sense, will remove the dir after these tests!

PR #165, opened 2020-04-30
Split provider code into separate modules

### LINKED ISSUE #133: Separate provider code into modules
Currently all providers are implemented as "monolithic modules" - i.e. all functionality is implemented in one large file. 

For ease of use it would be much better to split them and put each type of operation/method into it own file, e.g. `sign.rs`, `key_management.rs` etc.

This should be done for all providers.

All utility methods (e.g. for managing interaction with the `KeyIDManager`) should reside in their own module(s) as well.

ionut-arm:
This commit splits the code for the 3 existing providers into separate
modules: one for key management (containing the code for key
destruction, creation, import and export) and one for asymetric signing
(currently containing code for signing and verifying hashes).

Fixes #133

### DIFF
--- a/src/providers/pkcs11_provider/asym_sign.rs
+++ b/src/providers/pkcs11_provider/asym_sign.rs
@@ -1,0 +1,192 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use super::Pkcs11Provider;
+use super::{key_management::get_key_info, utils, KeyPairType, ReadWriteSession, Session};
+use crate::authenticators::ApplicationName;
+use crate::key_info_managers::KeyTriple;
+use log::{error, info};
+use parsec_interface::operations::psa_algorithm::*;
+use parsec_interface::operations::{psa_sign_hash, psa_verify_hash};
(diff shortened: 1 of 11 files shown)


src/providers/pkcs11_provider/asym_sign.rs
```diff
@@ -0,0 +1,192 @@
...
+                // DigestAlgorithmIdentifier value representing the SHA256 OID with no parameters.
+                // The OID used is: "2.16.840.1.101.3.4.2.1".
+                // It would be better to use the DigestInfo structure defined in this file but the
+                // AlgorithmIdentifier structure does not currently support the simple SHA256 OID.
+                // See Devolutions/picky-rs#19
+                let mut digest_info = vec![
+                    0x30, 0x31, 0x30, 0x0D, 0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04,
+                    0x02, 0x01, 0x05, 0x00, 0x04, 0x20,
+                ];
+                digest_info.append(&mut hash);
```
hug-dev, line 84: Not related at all with your PR but we should change that one of these days, since it has been fixed... by you 😋

PR #167, opened 2020-05-04
Improve digest handling in PKCS11 provider

ionut-arm:
This commit aims to improve our handling of digest data within the
PKCS11 provider. Previously we had harcoded the prefix to the actual
digest, assuming the digest was SHA256 - we are now generating a proper
DigestInfo structure that can be serialized to a ASN.1 DER encoding.

The reason for handling this data ourselves in the PKCS11 provider is
because the PKCS11 interface expects users to provide the formatting.
All other providers only take the digest data and generate the metadata
themselves.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ structopt = "0.3.5"
 derivative = "2.1.1"
 version = "3.0.0"
 hex = "0.4.2"
+picky = "5.0.0"
 
 [dev-dependencies]
 ring = "0.16.12"
--- a/src/providers/pkcs11_provider/asym_sign.rs
+++ b/src/providers/pkcs11_provider/asym_sign.rs
@@ -8,7 +8,16 @@ use log::{error, info};
 use parsec_interface::operations::psa_algorithm::*;
 use parsec_interface::operations::{psa_sign_hash, psa_verify_hash};
 use parsec_interface::requests::{ProviderID, ResponseStatus, Result};
+use picky::{algorithm_identifier::SHAVariant, AlgorithmIdentifier};
+use picky_asn1::wrapper::OctetStringAsn1;
 use pkcs11::types::CK_MECHANISM;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize)]
+struct DigestInfo {
+    oid: AlgorithmIdentifier,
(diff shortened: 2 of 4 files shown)


ionut-arm:
This PR is bound to fail until https://github.com/Devolutions/picky-rs/pull/34 is merged. This is because the changes I submitted to their project were faulty initially, without me realising it.

Cargo.toml
```diff
@@ -38,6 +38,7 @@ bincode = "1.1.4"
 structopt = "0.3.5"
 derivative = "2.1.1"
 version = "3.0.0"
+picky = "4.7.0"
```
ionut-arm, line 41: This should be added to the README

src/providers/pkcs11_provider/asym_sign.rs
```diff
@@ -70,18 +79,13 @@ impl Pkcs11Provider {
...
-                    0x02, 0x01, 0x05, 0x00, 0x04, 0x20,
-                ];
-                digest_info.append(&mut hash);
+                let digest_info = DigestInfo {
+                    oid: AlgorithmIdentifier::new_sha(SHAVariant::SHA2_256),
+                    digest: hash.into(),
+                };
+                let digest_info = picky_asn1_der::to_vec(&digest_info)
+                    // should not fail - if it does, there's some error in our stack
+                    .or(Err(ResponseStatus::PsaErrorGenericError))?;
```
hug-dev, line 88: I am wondering about the return status here, is it right to return `GenericError`? I am thinking on how this function (`to_vec`) could fail. Could that be because `digest` is not a `SHA256`, in which case this could be a `PsaErrorInvalidArgument`?

ionut-arm: I'll look into the code for `to_vec`, but I doubt it - I think it simply serializes the contents of that reference, regardless of what's in there.

ionut-arm: It only takes the value and serializes it, no checks done:

https://github.com/Devolutions/picky-rs/blob/master/picky-asn1-der/src/ser/mod.rs#L17

hug-dev: Yes I think that is the most appropriate one then.

PR #168, opened 2020-05-04
Improve TPM provider

### LINKED ISSUE #120: Allow TPM owner hierarchy auth to be non-string
Currently we only allow the owner hierarchy to be protected with a password (i.e. a String auth value), however this can be set to a binary value. 

A suggestion would be to support in the `owner_hierarchy_auth` field values of the form `str:...` and `hex:...`, defaulting to considering the value a string if no prefix is specified. Presumably the work in the service should just be around how this value is passed to the TSS crate, which should do the parsing behind the scenes. The issue tracking the TSS crate changes is here: https://github.com/parallaxsecond/rust-tss-esapi/issues/37

### LINKED ISSUE #121: TPM provider should establish most-secure primitives for itself

ionut-arm:
This commit includes a couple of fixes in the TPM provider.
* the TSS crate used has been updated to the latest one, including a few
functionality updates; this now allows the session hash algorithm and
the cipher used for sessions and primary key
* more control over the format of the authentication value is now
offered to admins; more precisely, they can provide string versions of
hex values, prefixed by "hex:"

Fixes #120 
Fixes #121

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -74,5 +74,8 @@ key_info_manager = "on-disk-manager"
 # - "tabrmd": uses the TPM2 Access Broker & Resource Management Daemon
 #tcti = "mssim"
 # (Required) Authentication value for performing operations on the TPM Owner Hierarchy. The string can
-# be empty, however we strongly suggest that you use a secure password.
+# be empty, however we strongly suggest that you use a secure passcode.
+# To align with TPM tooling, PARSEC allows "owner_hierarchy_auth" to have a prefix indicating a string value,
+# e.g. "str:password", or to represent a string version of a hex value, e.g. "hex:1a2b3c". If no prefix is
+# provided, the value is considered to be a string.
 #owner_hierarchy_auth = "password"
--- a/e2e_tests/provider_cfg/tpm/config.toml
+++ b/e2e_tests/provider_cfg/tpm/config.toml
@@ -16,4 +16,4 @@ manager_type = "OnDisk"
 provider_type = "Tpm"
 key_info_manager = "on-disk-manager"
 tcti = "mssim"
-owner_hierarchy_auth = "tpm_pass"
+owner_hierarchy_auth = "hex:74706d5f70617373" # "tpm_pass" in hex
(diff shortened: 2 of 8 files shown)


config.toml
```diff
@@ -74,5 +74,8 @@ key_info_manager = "on-disk-manager"
 # - "tabrmd": uses the TPM2 Access Broker & Resource Management Daemon
 #tcti = "mssim"
 # (Required) Authentication value for performing operations on the TPM Owner Hierarchy. The string can
-# be empty, however we strongly suggest that you use a secure password.
+# be empty, however we strongly suggest that you use a secure passcode.
+# To align with TPM tooling, PARSEC allows "owner_hierarchy_auth" to have a prefix indicating a string value,
+# e.g. "str:password", or to represent a string version of a hex value, e.g. "hex:1a2b3c". If no prefix is
+# provided, the value is considered to be a string.
 #owner_hierarchy_auth = "password"
```
hug-dev, line 81: Should we set `str:password` as default?

e2e_tests/provider_cfg/tpm/config.toml
```diff
@@ -16,4 +16,4 @@ manager_type = "OnDisk"
 provider_type = "Tpm"
 key_info_manager = "on-disk-manager"
 tcti = "mssim"
-owner_hierarchy_auth = "tpm_pass"
+owner_hierarchy_auth = "hex:74706d5f70617373" # "tpm_pass" in hex
```
hug-dev, line 19: Is that different from `"str:tpm_pass"`?

ionut-arm: No, it isn't, just wanted to test it.

config.toml
ionut-arm: We could, but as stated above it's not really needed

PR #171, opened 2020-05-11
Add ECDSA support for TPM provider

ionut-arm:
This commit expands the supported algorithms for the TPM provider to
include ECDSA key creation, as well as signing, verification and public
key export for them.

### DIFF
--- a/src/providers/tpm_provider/key_management.rs
+++ b/src/providers/tpm_provider/key_management.rs
@@ -13,7 +13,6 @@ use parsec_interface::operations::{
     psa_destroy_key, psa_export_public_key, psa_generate_key, psa_import_key,
 };
 use parsec_interface::requests::{ProviderID, ResponseStatus, Result};
-use picky_asn1::wrapper::IntegerAsn1;
 
 // Public exponent value for all RSA keys.
 const PUBLIC_EXPONENT: [u8; 3] = [0x01, 0x00, 0x01];
@@ -69,17 +68,9 @@ impl TpmProvider {
         app_name: ApplicationName,
         op: psa_generate_key::Operation,
     ) -> Result<psa_generate_key::Result> {
-        if op.attributes.key_type != KeyType::RsaKeyPair {
-            error!("The TPM provider currently only supports creating RSA key pairs.");
-            return Err(ResponseStatus::PsaErrorNotSupported);
-        }
-
         let key_name = op.key_name;
         let attributes = op.attributes;
         let key_triple = KeyTriple::new(app_name, ProviderID::Tpm, key_name);
-        // This should never panic on 32 bits or more machines.
-        let key_size = std::convert::TryFrom::try_from(op.attributes.key_bits)
-            .expect("Conversion to usize failed.");
 
         let mut store_handle = self
             .key_info_store
@@ -91,7 +82,7 @@ impl TpmProvider {
             .expect("ESAPI Context lock poisoned");
 
         let (key_context, auth_value) = esapi_context
-            .create_rsa_signing_key(key_size, AUTH_VAL_LEN)
+            .create_signing_key(utils::parsec_to_tpm_params(attributes)?, AUTH_VAL_LEN)
             .or_else(|e| {
                 error!("Error creating a RSA signing key: {}.", e);
                 Err(utils::to_response_status(e))
@@ -199,7 +190,7 @@ impl TpmProvider {
             .lock()
             .expect("ESAPI Context lock poisoned");
 
-        let (password_context, _key_attributes) = get_password_context(&*store_handle, key_triple)?;
+        let (password_context, key_attributes) = get_password_context(&*store_handle, key_triple)?;
 
         let pub_key_data = esapi_context
             .read_public_key(password_context.context)
@@ -208,18 +199,9 @@ impl TpmProvider {
                 Err(utils::to_response_status(e))
             })?;
 
-        let key = RsaPublicKey {
-            // To produce a valid ASN.1 RSAPublicKey structure, 0x00 is put in front of the positive
-            // modulus if highest significant bit is one, to differentiate it from a negative number.
-            modulus: IntegerAsn1::from_unsigned_bytes_be(pub_key_data),
-            public_exponent: IntegerAsn1::from_signed_bytes_be(PUBLIC_EXPONENT.to_vec()),
-        };
-        let key_data = picky_asn1_der::to_vec(&key).or_else(|err| {
-            error!("Could not serialise key elements: {}.", err);
-            Err(ResponseStatus::PsaErrorCommunicationFailure)
-        })?;
-
-        Ok(psa_export_public_key::Result { data: key_data })
+        Ok(psa_export_public_key::Result {
+            data: utils::pub_key_to_bytes(pub_key_data, key_attributes)?,
+        })
     }
 
     pub(super) fn psa_destroy_key_internal(
@@ -234,10 +216,9 @@ impl TpmProvider {
             .write()
             .expect("Key store lock poisoned");
 
-        let error_closure = |e| Err(key_info_managers::to_response_status(e));
         if store_handle
             .remove(&key_triple)
-            .or_else(error_closure)?
+            .map_err(key_info_managers::to_response_status)?
             .is_none()
         {
             error!(
--- a/src/providers/tpm_provider/utils.rs
+++ b/src/providers/tpm_provider/utils.rs
@@ -2,11 +2,18 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use log::error;
-use parsec_interface::requests::ResponseStatus;
+use parsec_interface::operations::psa_algorithm::*;
+use parsec_interface::operations::psa_key_attributes::*;
+use parsec_interface::requests::{ResponseStatus, Result};
 use picky_asn1::wrapper::IntegerAsn1;
 use serde::{Deserialize, Serialize};
+use std::convert::TryFrom;
+use std::convert::TryInto;
+use tss_esapi::abstraction::transient::KeyParams;
 use tss_esapi::response_code::{Error, Tss2ResponseCodeKind};
-use tss_esapi::utils::TpmsContext;
+use tss_esapi::utils::algorithm_specifiers::{EllipticCurve, HashingAlgorithm};
+use tss_esapi::utils::{AsymSchemeUnion, PublicKey, Signature, SignatureData, TpmsContext};
+const PUBLIC_EXPONENT: [u8; 3] = [0x01, 0x00, 0x01];
 
 /// Convert the TSS library specific error values to ResponseStatus values that are returned on
 /// the wire protocol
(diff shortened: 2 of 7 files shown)


src/providers/tpm_provider/utils.rs
```diff
@@ -84,3 +91,134 @@ pub struct PasswordContext {
...
+        KeyType::EccKeyPair { .. } | KeyType::EccPublicKey { .. } => {
+            let p_size = usize::try_from(key_attributes.key_bits / 8).unwrap();
+            if data.len() != p_size * 2 {
+                return Err(ResponseStatus::PsaErrorInvalidArgument);
+            }
+
+            let mut r = data;
+            let s = r.split_off(p_size);
+            Ok(SignatureData::EcdsaSignature { r, s })
+        }
```
hug-dev, line 221: I assume that this extracts `r` and `s` values from the concatenation `data`? Basically what is described in [our book](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_algorithm.html#asymmetricsignature-algorithm):

> The representation of the signature as a byte string consists of the concatenation of the signature values r and s. Each of r and s is encoded as an N-octet string, where N is the length of the base point of the curve in octets. Each value is represented in big-endian order, with the most significant octet first.

Maybe it would be nice to put a comment to explain that and put some context 😃

src/providers/tpm_provider/utils.rs
```diff
@@ -84,3 +91,134 @@ pub struct PasswordContext {
...
+}
+
+fn elliptic_curve_point_to_octet_string(mut x: Vec<u8>, mut y: Vec<u8>) -> Vec<u8> {
+    let mut octet_string = vec![0x04];
+    octet_string.append(&mut x);
+    octet_string.append(&mut y);
+    octet_string
+}
+
+pub fn signature_data_to_bytes(data: SignatureData, p_byte_size: usize) -> Result<Vec<u8>> {
```
hug-dev, line 181: The `p_byte_size` parameter is only used for the ECDSA signature arm. Would it be better if a `KeyAttributes` parameter type is passed instead (to mirror the `bytes_to_signature_data`) function and `p_byte_size` computed from `key_bits`?
That function could also check if the key type is matching the signature type and return `InvalidArgument`, same way as the other function.

hug-dev: Also, this function is public while the other one is not, any reason for that?

src/providers/tpm_provider/utils.rs
```diff
@@ -84,3 +91,134 @@ pub struct PasswordContext {
...
+                );
+                return Err(ResponseStatus::PsaErrorGenericError);
+            }
+            Ok(elliptic_curve_point_to_octet_string(x, y))
+        }
+    }
+}
+
+fn elliptic_curve_point_to_octet_string(mut x: Vec<u8>, mut y: Vec<u8>) -> Vec<u8> {
+    let mut octet_string = vec![0x04];
```
hug-dev, line 175: I think some comments would be nice to explain where this `0x04` come from

src/providers/tpm_provider/utils.rs
```diff
@@ -84,3 +91,134 @@ pub struct PasswordContext {
...
+        (EccFamily::SecpR1, 192) => Ok(EllipticCurve::NistP192),
+        (EccFamily::SecpR1, 224) => Ok(EllipticCurve::NistP224),
+        (EccFamily::SecpR1, 256) => Ok(EllipticCurve::NistP256),
+        (EccFamily::SecpR1, 384) => Ok(EllipticCurve::NistP384),
+        (EccFamily::SecpR1, 512) => Ok(EllipticCurve::NistP521),
+        _ => Err(ResponseStatus::PsaErrorNotSupported),
+    }
+}
+
+pub fn pub_key_to_bytes(pub_key: PublicKey, p_byte_size: usize) -> Result<Vec<u8>> {
```
hug-dev, line 153: Same for the `p_byte_size` param here

src/providers/tpm_provider/utils.rs
```diff
@@ -84,3 +91,134 @@ pub struct PasswordContext {
...
+        })
+        .or(Err(ResponseStatus::PsaErrorGenericError)),
+        PublicKey::Ecc { x, y } => {
+            if x.len() != p_byte_size || y.len() != p_byte_size {
+                error!(
+                    "Received ECC public key with invalid size: x - {} bytes; y - {} bytes",
+                    x.len(),
+                    y.len()
+                );
+                return Err(ResponseStatus::PsaErrorGenericError);
```
hug-dev, line 167: Isn't it `InvalidArgument` for this one?

hug-dev: or `CommunicationFailure`, if the information comes from the TPM?

src/providers/tpm_provider/utils.rs
```diff
@@ -84,3 +91,134 @@ pub struct PasswordContext {
...
+    }
+}
+
+fn convert_curve_to_tpm(curve_family: EccFamily, key_size: u32) -> Result<EllipticCurve> {
+    match (curve_family, key_size) {
+        (EccFamily::SecpR1, 192) => Ok(EllipticCurve::NistP192),
+        (EccFamily::SecpR1, 224) => Ok(EllipticCurve::NistP224),
+        (EccFamily::SecpR1, 256) => Ok(EllipticCurve::NistP256),
+        (EccFamily::SecpR1, 384) => Ok(EllipticCurve::NistP384),
+        (EccFamily::SecpR1, 512) => Ok(EllipticCurve::NistP521),
```
hug-dev, line 148: Just checking the number, we have 512 on one side and 521 on the other, is it a typo?

src/providers/tpm_provider/utils.rs
```diff
@@ -84,3 +91,134 @@ pub struct PasswordContext {
...
+        }
+        KeyType::EccKeyPair { curve_family } => Ok(KeyParams::Ecc {
+            scheme: convert_asym_scheme_to_tpm(attributes.key_policy.key_algorithm)?,
+            curve: convert_curve_to_tpm(curve_family, attributes.key_bits)?,
+        }),
+        _ => Err(ResponseStatus::PsaErrorNotSupported),
+    }
+}
+
+fn convert_asym_scheme_to_tpm(algorithm: Algorithm) -> Result<AsymSchemeUnion> {
```
hug-dev, line 116: Would it be better to take `AsymmetricSignature` as parameter here? Or rename the function `convert_scheme_to_tpm` if that still makes sense for other algorithms.

src/providers/tpm_provider/utils.rs
```diff
@@ -84,3 +91,134 @@ pub struct PasswordContext {
...
+                _ => return Err(ResponseStatus::PsaErrorInvalidArgument),
+            };
+            Ok(KeyParams::Rsa {
+                size,
+                scheme: convert_asym_scheme_to_tpm(attributes.key_policy.key_algorithm)?,
+                pub_exponent: 0,
+            })
+        }
+        KeyType::EccKeyPair { curve_family } => Ok(KeyParams::Ecc {
+            scheme: convert_asym_scheme_to_tpm(attributes.key_policy.key_algorithm)?,
```
hug-dev, line 109: Does the `scheme` field of `KeyParams::Ecc` carry the same meaning as the algorithm in the PSA policy?
My understanding is that the `scheme` is the algorithm used with the key for sign/verify operation whereas the policy algorithm is the set of algorithms allowed to be used with taht key.
How would you handle a policy algorithm with `Hash::Any` for example? (I am not criticising, just raising general questions!)
Can the `scheme` of a key be changed on-the-fly on an operation per operation level? As in, changing it with the `alg` field of the Parsec `Operation` that have it.

src/providers/tpm_provider/key_management.rs
```diff
@@ -208,18 +200,12 @@ impl TpmProvider {
...
-        let key_data = picky_asn1_der::to_vec(&key).or_else(|err| {
-            error!("Could not serialise key elements: {}.", err);
-            Err(ResponseStatus::PsaErrorCommunicationFailure)
-        })?;
-
-        Ok(psa_export_public_key::Result { data: key_data })
+        Ok(psa_export_public_key::Result {
+            data: utils::pub_key_to_bytes(
+                pub_key_data,
+                usize::try_from(key_attributes.key_bits / 8).unwrap(), // should not fail on supported platforms
```
hug-dev, line 206: Not at all in scope with this PR, but maybe we should implement a `From` the error returned by `TryFrom` to `ResponseStatus` to be able to remove those `unwrap`, even though they should not happen on the correct platforms.

src/providers/tpm_provider/utils.rs
ionut-arm: > Also, this function is public while the other one is not, any reason for that?

Only made the ones that are needed public - `utils` is not public past the provider anyway, so it's all limited to within the provider module.

src/providers/tpm_provider/utils.rs
ionut-arm: It's a typo in the spec...

src/providers/tpm_provider/utils.rs
ionut-arm: > Does the scheme field of KeyParams::Ecc carry the same meaning as the algorithm in the PSA policy?

> How would you handle a policy algorithm with Hash::Any for example?

> Can the scheme of a key be changed on-the-fly on an operation per operation level?

Yes, although in TPM world there isn't any wildcard, so the scheme on the key and the one used when signing/verifying have to be the same. Also, like in the PSA spec, there isn't any way of changing the scheme associated with a key. Hence the method below, `convert_asym_scheme_to_tpm`, does the correct thing by returning `PsaErrorNotSupported` when anything like that is encountered.

src/providers/tpm_provider/utils.rs
ionut-arm: Maybe `CommunicationFailure` - it isn't invalid argument because the public key comes from us, not from the user.

src/providers/tpm_provider/utils.rs
ionut-arm: Yes and no - asymmetric encryption schemes exist, it's just that at the moment we don't deal with them. I could split this further into multiple methods, for each kind of asymmetric scheme, but I don't think it's wrong to take an algorithm here - especially since pattern matching in Rust allows things like below. Otherwise you'd have two distinct matches doing the same work.

src/providers/tpm_provider/utils.rs
ionut-arm: > That function could also check if the key type is matching the signature type and return InvalidArgument, same way as the other function.

Not sure if that's sensible, but could do. The `SignatureData` is coming from the TPM and there is already a conversion step there, from the raw FFI data to this.

src/providers/tpm_provider/utils.rs
hug-dev: Ok, that makes sense!
We will need to make sure to verify, test and document that wildcard algorithms are not supported in the policy for TPM keys so that the algorithm in the key policy **is** the algotithm that the key will use.

src/providers/tpm_provider/utils.rs
hug-dev: Got it, I forgot about asymmetric encryption

src/providers/tpm_provider/utils.rs
hug-dev: `CommunicationFailure` sounds right then

src/providers/tpm_provider/key_management.rs
ionut-arm: That would be from this type: https://doc.rust-lang.org/std/num/struct.TryFromIntError.html

src/providers/tpm_provider/utils.rs
hug-dev: Yes I don't know if it would be useful then, apart from making sure that `data` is of the form we expect. I am just thinking that the caller would wonder what put for the `p_byte_size` parameter if doing RSA

ionut-arm: > I am just thinking that the caller would wonder what put for the p_byte_size parameter if doing RSA

Yeah, that makes sense!!

src/providers/tpm_provider/key_management.rs
ionut-arm: I'm guessing that would be a `GenericError`? Or would it be some sort of hardware error?

hug-dev: `GenericError` with a `error!` log message saying why seems fine I think

src/providers/tpm_provider/utils.rs
ionut-arm: I've added a test to verify that creating a key with the wildcard algorithm returns `NotSupported`. I've also changed the `parsec_to_tpm_signature` method to take in an `AsymmetricSignature` value too, given from the operation object, to use for setting the asymmetric scheme for the signature, it seemed more correct to do it that way. But the method you highlighted here wasn't changed because this is literally about converting the policy used on the key rather than on a particular use of the key.

src/providers/tpm_provider/utils.rs
ionut-arm: Dun!

src/providers/tpm_provider/utils.rs
ionut-arm: Done!

src/providers/tpm_provider/utils.rs
```diff
@@ -84,3 +91,159 @@ pub struct PasswordContext {
...
+    })
+}
+
+fn bytes_to_signature_data(data: Vec<u8>, key_attributes: KeyAttributes) -> Result<SignatureData> {
+    match key_attributes.key_type {
+        KeyType::RsaKeyPair | KeyType::RsaPublicKey => Ok(SignatureData::RsaSignature(data)),
+        KeyType::EccKeyPair { .. } | KeyType::EccPublicKey { .. } => {
+            // ECDSA signature data is represented the concatenation of the two result values, r and s,
+            // in big endian format, as described here:
+            // https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_algorithm.html#asymmetricsignature-algorithm
```
hug-dev, line 237: Thanks!

tests/providers/tpm.rs
```diff
@@ -93,3 +125,82 @@ fn verify_with_ring() {
...
+    let mut op = gen_ecc_sign_key_op(key_name);
+    op.attributes.key_policy.key_algorithm =
+        Algorithm::AsymmetricSignature(AsymmetricSignature::Ecdsa {
+            hash_alg: Hash::Any,
+        });
+    assert_eq!(
+        TPM_PROVIDER.psa_generate_key(app_name, op).unwrap_err(),
+        ResponseStatus::PsaErrorNotSupported
+    );
+}
```
hug-dev, line 206: Cool!

PR #172, opened 2020-05-13
Add default context cipher selection for TPM provider

ionut-arm:
This commit allows the TPM provider to identify the most appropriate
cipher that is supported by the underlying TPM, using it for session
encryption and as the main algorithm for the primary storage key.


src/providers/tpm_provider/mod.rs
```diff
@@ -216,6 +217,33 @@ impl TpmProviderBuilder {
         }
     }
 
+    unsafe fn find_default_context_cipher(&self) -> std::io::Result<Cipher> {
```
hug-dev, line 220: Maybe a comment explaining why `unsafe` is needed and why it is safe under our conditions (I think it is covered by the `Safety` notice of the calling function) would be good.

PR #173, opened 2020-05-15
Improve handling of list_opcodes

ionut-arm:
This commit adjusts the handling of `list_opcodes` to match the
documentation. This implies allowing the Core provider to reply with a
set of opcodes on behalf of the other providers.

### DIFF
--- a/e2e_tests/tests/all_providers/mod.rs
+++ b/e2e_tests/tests/all_providers/mod.rs
@@ -22,9 +22,6 @@ fn list_providers() {
 }
 
 #[test]
-// ListOpcodes is not currently supported only by the Core Provider, see
-// parallaxsecond/parsec-operations#9
-#[ignore]
 fn list_opcodes() {
     let mut client = TestClient::new();
     let mut crypto_providers_opcodes = HashSet::new();
--- a/src/providers/core_provider/mod.rs
+++ b/src/providers/core_provider/mod.rs
@@ -10,6 +10,7 @@ use log::error;
 use parsec_interface::operations::list_providers::ProviderInfo;
 use parsec_interface::operations::{list_opcodes, list_providers, ping};
 use parsec_interface::requests::{Opcode, ProviderID, ResponseStatus, Result};
+use std::collections::{HashMap, HashSet};
 use std::io::{Error, ErrorKind};
 use std::str::FromStr;
 use uuid::Uuid;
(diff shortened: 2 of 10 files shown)


src/providers/core_provider/mod.rs
```diff
@@ -26,37 +27,27 @@ const SUPPORTED_OPCODES: [Opcode; 3] = [Opcode::ListProviders, Opcode::ListOpcod
 pub struct CoreProvider {
     wire_protocol_version_min: u8,
     wire_protocol_version_maj: u8,
-    providers: Vec<ProviderInfo>,
+    providers: HashMap<ProviderInfo, HashSet<Opcode>>,
```
hug-dev, line 30: It might be a bit more efficient/simpler to add a `providers_opcodes: HashMap<ProviderID, HashSet<Opcode>>` and rename the current `providers` field `providers_info`. That way you do not have to iterate through all the providers and the logic is a bit more simple in `list_opcodes`.

src/providers/core_provider/mod.rs
```diff
@@ -75,16 +66,42 @@ impl Provide for CoreProvider {
...
+                )))?,
+                description: String::from("Software provider that implements only administrative (i.e. no cryptographic) operations"),
+                vendor: String::new(),
+                version_maj: crate_version.major,
+                version_min: crate_version.minor,
+                version_rev: crate_version.patch,
+                id: ProviderID::Core,
+            },
+            SUPPORTED_OPCODES.iter().copied().collect(),
+        );
```
hug-dev, line 98: I think having this in an independant method (`describe`) is quite nice to identify this information more easily and mirror the other providers.

ionut-arm: Yeah, I ended up shuffling things around a few times, in the current format it makes sense to re-implement describe on the core provider

src/providers/core_provider/mod.rs
ionut-arm: Indeed, but you'd have duplicate data (the provider ID). I'll try and figure out why we don't include `opcodes` in the `ProviderInfo` in the first place.

src/providers/core_provider/mod.rs
ionut-arm: Oh, yeah! `describe` requires that an instance of the object already exists, so then you can't add it before you build it. You could build the provider, describe it and then add that information in itself, but that feels counter-intuitive as well. I think I'll leave it like this.

src/providers/core_provider/mod.rs
ionut-arm: Done

e2e_tests/tests/all_providers/mod.rs
```diff
@@ -22,9 +22,6 @@ fn list_providers() {
 }
 
 #[test]
-// ListOpcodes is not currently supported only by the Core Provider, see
-// parallaxsecond/parsec-operations#9
-#[ignore]
```
hug-dev, line 27: 🤶 Yes!

PR #182, opened 2020-06-07
Add GlobalConfig

ionut-arm:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,6 +37,17 @@
 // This one is hard to avoid.
 #![allow(clippy::multiple_crate_versions)]
 
+#[allow(unused)]
+macro_rules! format_error {
+    ($message:expr, $error:expr) => {
+        if crate::utils::GlobalConfig::log_error_details() {
+            log::error!("{}; Error: {}", $message, $error)
+        } else {
+            log::error!("{};", $message)
+        }
+    };
+}
+
 pub mod authenticators;
 pub mod back;
 pub mod front;
--- a/src/utils/global_config.rs
+++ b/src/utils/global_config.rs
@@ -1,0 +1,51 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering;
+
+/// Configuration values that affect most or all the
+/// components of the service.
+#[derive(Default, Debug)]
+pub struct GlobalConfig {
+    log_error_details: AtomicBool,
+}
+
+impl GlobalConfig {
+    const fn new() -> Self {
+        GlobalConfig {
+            log_error_details: AtomicBool::new(false),
+        }
+    }
+
+    /// Determine whether error logs should include detailed
+    /// information about the error
+    pub fn log_error_details() -> bool {
+        GLOBAL_CONFIG.log_error_details.load(Ordering::Relaxed)
+    }
+}
+
+static GLOBAL_CONFIG: GlobalConfig = GlobalConfig::new();
+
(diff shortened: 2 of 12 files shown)


src/utils/global_config.rs
```diff
@@ -0,0 +1,63 @@
...
+
+impl GlobalConfig {
+    /// Determine whether error logs should include detailed
+    /// information about the error
+    pub fn verbose_log() -> bool {
+        GLOBAL_CONFIG.0.borrow().verbose_log
+    }
+}
+
+struct GlobalConfigWrapper(RefCell<GlobalConfig>);
```
hug-dev, line 20: We could also go full-safe and use a `RwLock`? Maybe the performence penalty of calling `read` is not too big?

src/lib.rs
```diff
@@ -37,6 +38,17 @@
 // This one is hard to avoid.
 #![allow(clippy::multiple_crate_versions)]
 
+#[allow(unused)]
```
hug-dev, line 41: I am wondering if `format_error` and `verbose_logs` are the best names. The purpose of this flag is to hide the error messages coming from the crypto APIs that we use, unless you want to use it for other things?
If it is a security option, maybe it could be a bit more obvious in its name:
* `verbose_log` -> `print_crypto_lib_errors`
* `format_error` -> `crypto_lib_error`

ionut-arm:
> items affecting all the service, but should be limited to that.

What about configuration values that could apply to all providers at the same time (e.g. a limit on the max number of characters a key name can have)?

hug-dev:
> 
> 
> > items affecting all the service, but should be limited to that.
> 
> What about configuration values that could apply to all providers at the same time (e.g. a limit on the max number of characters a key name can have)?

Yes, too!

src/utils/global_config.rs
ionut-arm: I'll drop the `GlobalConfigWrapper` and go with `AtomicBool` in the `GlobalConfig`, in the hope that we won't need non-integer-or-bool configs in the future.

src/lib.rs
ionut-arm: I think a change in `verbose_log` makes sense, but `format_error` seems reasonable, since it handles the whole formatting decision for the caller.
Also, it's not just about crypto library errors, it's also about errors coming from `bincode` when (de)serializing `KeyInfo` structures, for example.

`verbose_log` -> `log_error_objects`? `log_error_details`?

hug-dev: `log_error_details` seems good!

ionut-arm:
I'll actually update the log messages in a different PR

src/lib.rs
```diff
@@ -6,6 +6,7 @@
 //! SECurity) service. For a more in-depth guide of the system architecture,
 //! supported operations and other Parsec-related topics, see our
 //! [Parsec Book](https://parallaxsecond.github.io/parsec-book/index.html).
+#![macro_use]
```
hug-dev, line 9: Is this needed to use the macro? I thought that with recent versions of `rustc` you could just import the macro name as a function.

src/utils/global_config.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    pub fn with_log_error_details(mut self, log_error_details: bool) -> Self {
+        self.log_error_details = log_error_details;
+
+        self
+    }
+
+    /// This must only be called in a single-threaded environment where
+    /// `GLOBAL_CONFIG` is guaranteed to not be in use - i.e. during
+    /// service (re-)configuration.
```
hug-dev, line 48: I think now this can be called anywhere safely.

src/lib.rs
ionut-arm: Indeed! I've removed it, thought it was deprecated only for importing

PR #183, opened 2020-06-08
Improve log security

### LINKED ISSUE #175: Check logs for sensitive data
As part of the threat model it was suggested that logs might be a source of sensitive data being logged. We should review all logging sources and determine if any such data is being stored. We should also keep this in mind for future work.

ionut-arm:
This commit refactors most of the log commands and conditionally hides
any potentially sensitive information.

Fixes #175

### DIFF
--- a/src/key_info_managers/on_disk_manager/mod.rs
+++ b/src/key_info_managers/on_disk_manager/mod.rs
@@ -187,12 +187,14 @@ impl OnDiskKeyInfoManager {
         for app_name_dir_path in list_dirs(&mappings_dir_path)?.iter() {
             for provider_dir_path in list_dirs(&app_name_dir_path)?.iter() {
                 for key_name_file_path in list_files(&provider_dir_path)?.iter() {
-                    info!("Found mapping file: {:?}.", key_name_file_path);
+                    if crate::utils::GlobalConfig::log_error_details() {
+                        info!("Found mapping file: {:?}.", key_name_file_path);
+                    }
                     let mut key_info = Vec::new();
                     let mut key_info_file = File::open(&key_name_file_path)?;
                     let _ = key_info_file.read_to_end(&mut key_info)?;
                     let key_info = bincode::deserialize(&key_info[..]).or_else(|e| {
-                        error!("Error deserializing key info ({}).", e);
+                        format_error!("Error deserializing key info", e);
                         Err(Error::new(ErrorKind::Other, "error deserializing key info"))
                     })?;
                     match base64_data_triple_to_key_triple(
@@ -210,11 +212,18 @@ impl OnDiskKeyInfoManager {
                             let _ = key_store.insert(key_triple, key_info);
                         }
                         Err(string) => {
-                            error!("Failed to convert the mapping path found to an UTF-8 string (error: {}).", string);
+                            format_error!(
+                                "Failed to convert the mapping path found to an UTF-8 string",
+                                string
+                            );
                         }
                     }
                 }
             }
+        }
+
+        if !crate::utils::GlobalConfig::log_error_details() {
+            info!("Found {} mapping files", key_store.len());
         }
 
         Ok(OnDiskKeyInfoManager {
@@ -240,7 +249,7 @@ impl OnDiskKeyInfoManager {
 
         let mut mapping_file = fs::File::create(&key_name_file_path)?;
         mapping_file.write_all(&bincode::serialize(key_info).or_else(|e| {
-            error!("Error serializing key info ({}).", e);
+            format_error!("Error serializing key info", e);
             Err(Error::new(ErrorKind::Other, "error serializing key info"))
         })?)
     }
--- a/src/providers/pkcs11_provider/key_management.rs
+++ b/src/providers/pkcs11_provider/key_management.rs
@@ -58,7 +58,11 @@ pub fn create_key_id(
     match store_handle.insert(key_triple.clone(), key_info) {
         Ok(insert_option) => {
             if insert_option.is_some() {
-                warn!("Overwriting Key triple mapping ({})", key_triple);
+                if crate::utils::GlobalConfig::log_error_details() {
+                    warn!("Overwriting Key triple mapping ({})", key_triple);
+                } else {
+                    warn!("Overwriting Key triple mapping");
+                }
             }
             let _ = local_ids_handle.insert(key_id);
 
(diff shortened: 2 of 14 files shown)


src/key_info_managers/on_disk_manager/mod.rs
```diff
@@ -187,12 +187,14 @@ impl OnDiskKeyInfoManager {
         for app_name_dir_path in list_dirs(&mappings_dir_path)?.iter() {
             for provider_dir_path in list_dirs(&app_name_dir_path)?.iter() {
                 for key_name_file_path in list_files(&provider_dir_path)?.iter() {
-                    info!("Found mapping file: {:?}.", key_name_file_path);
+                    if crate::utils::GlobalConfig::log_error_details() {
+                        info!("Found mapping file: {:?}.", key_name_file_path);
+                    }
```
hug-dev, line 192: Even if `log_error_details` is not set, it would be nice to know that *a* mapping file was found. Maybe this is the case for a `format_info` macro?

And `format_warning` for completeness 🤡

src/providers/pkcs11_provider/key_management.rs
```diff
@@ -280,7 +286,15 @@ impl Pkcs11Provider {
...
-            error!("If the bits field is non-zero (value is {}) it must be equal to the size of the key in data.", key_attributes.bits);
+            if crate::utils::GlobalConfig::log_error_details() {
+                error!(
+                    "`bits` field of key attributes (value: {}) was not equal to the size of the key in `data` (value: {}).",
+                    key_attributes.bits,
+                    modulus_object.len() * 8
+                );
+            } else {
+                error!("`bits` field of key attributes was not equal to the size of the key in `data`.");
+            }
```
hug-dev, line 297: I think we lose the information of it being possibly zero

src/providers/pkcs11_provider/mod.rs
```diff
@@ -113,7 +113,11 @@ impl Pkcs11Provider {
                         ) {
                             Ok(key_id) => key_id,
                             Err(response_status) => {
-                                error!("Error getting the KeyInfo for triple:\n{}\n(error: {}), continuing...", key_triple, response_status);
+                                if crate::utils::GlobalConfig::log_error_details() {
+                                    error!("Error getting the KeyInfo for triple:\n{}\n(error: {}), continuing...", key_triple, response_status);
+                                } else {
+                                    error!("Error getting the KeyInfo, continuing...");
+                                }
```
hug-dev, line 120: I am thinking, to cater for this formatting, could we modify the `format_error` macro to be the following:
```rust
format_error!("Error finding key objects ({})", e);
```
and depending if permitted, the `()` will be expended with the error or left empty? That way we can place the error wherever we want in the format string?

src/providers/pkcs11_provider/utils.rs
```diff
@@ -48,11 +48,13 @@ pub fn rv_to_response_status(rv: CK_RV) -> ResponseStatus {
         s @ CKR_CURVE_NOT_SUPPORTED
         | s @ CKR_DOMAIN_PARAMS_INVALID
         | s @ CKR_FUNCTION_NOT_SUPPORTED => {
-            error!("Not supported value ({:?})", s);
+            if crate::utils::GlobalConfig::log_error_details() {
+                error!("Not supported value ({:?})", s);
```
hug-dev, line 52: Is it not a `format_error` here?

src/providers/pkcs11_provider/mod.rs
ionut-arm: Unfortunately the first argument into `error!` (and all other macros) has to be a string literal. And also, it feels a bit weird to do string processing just to be able to use that macro everywhere

src/providers/pkcs11_provider/utils.rs
ionut-arm: Notice the missing `else` branch - I thought that printing `Not supported value` while returning `PsaErrorNotSupported` is kinda redundant... In the other ones at least we're giving some more info

src/providers/pkcs11_provider/key_management.rs
ionut-arm: ? if `bits` is 0 then it doesn't go into this branch of the if on line 282/288

ionut-arm: Or do you mean if `modulus_object.len()` is 0?

src/key_info_managers/on_disk_manager/mod.rs
ionut-arm: The thing is, for errors `format_error` makes sense because we want to potentially display the error, but for warning and info it tends to be various bits of data that we might or might not want to log

ionut-arm: Having 100 `Found mapping file` seems weird, but I can add a `Found {} mapping files` and the count just before returning from this method?

ionut-arm: The reason why in some cases I decided to not print anything is because it seems like it's not adding much value, but I'm open to reconsidering them!

src/providers/pkcs11_provider/utils.rs
ionut-arm: (oh, also, it's not a `format_error` because `s` doesn't implement `Display` 😬 which is required for `format_error` (because it uses `{}`, not `{:?}`)

src/key_info_managers/on_disk_manager/mod.rs
hug-dev: When testing persistence, it is nice to see that keys were recognised and added, a count might help then!

src/providers/pkcs11_provider/key_management.rs
hug-dev: I just meant that before this message might help the developper giving him the information that he can either leave this field to 0 or he has to put the correct size. Currently it only seems like he *has* to put the exact size.

src/key_info_managers/on_disk_manager/mod.rs
ionut-arm: Added here: https://github.com/parallaxsecond/parsec/pull/183/files#diff-1b1e7f5d68119cb3267c219b6be4ce9eR225

hug-dev: Nice, thanks!

PR #185, opened 2020-06-11
Add trace logging on Provide method calls

hug-dev:
That would allow to make sure that Parsec is indeed useful :)


hug-dev:
> 
> 
> Could you please also add two trace logs for every method along the way to/from the providers?
> 
> For example for `BackEndHandler`, when entering the `execute_request` method you log `execute_request ingress` and after you get the result from provider you log `execute_request egress`

✨ It's all done ✨

Issue #177, opened 2020-06-02
Use psa-crypto crate in the Mbed Crypto Provider


hug-dev:
The Mbed Crypto provider is still using the legacy code to perform the conversion between Rust native types from the Parsec interface and FFI types to call the PSA Crypto API.
These conversion can now directly be used from the crate as well as the build scripts to find and link with Mbed Crypto.

sbailey-arm:
I'm intending to work on this.

PR #186, opened 2020-06-17
Move Parsec over to psa-crypto 

sbailey-arm:
#177 

**Note:** It currently won't build due to local dependencies. Once `psa-crypto` and `parsec-interface` are updated on crates.io, we can update the dependencies in this PR to make it build. Currently this PR is for everyone to see and discuss the changes.


src/providers/mbed_provider/key_management.rs
```diff
@@ -13,14 +11,16 @@ use parsec_interface::operations::{
     psa_destroy_key, psa_export_public_key, psa_generate_key, psa_import_key,
 };
 use parsec_interface::requests::{ProviderID, ResponseStatus, Result};
+use psa_crypto::operations::key_management as new_key_management;
```
sbailey-arm, line 14: I wasn't sure what to do about this rename. I thought leaving it as `use psa_crypto::operations::key_management;` might get confusing as it is already in a file called `key_management.rs`. Thoughts? Maybe `psa_crypto_key_management`?

src/providers/mbed_provider/mod.rs
```diff
@@ -107,6 +101,11 @@ impl MbedProvider {
             // Delete those who are not present and add to the local_store the ones present.
             match store_handle.get_all(ProviderID::MbedCrypto) {
                 Ok(key_triples) => {
+                    if let Err(error) = psa_crypto::init() {
+                        error!("Error {} when initialising Mbed Crypto library.", error);
+                        return None;
+                    }
```
ionut-arm, line 107: What's the purpose of this init if the one on line 77 happens?

src/providers/mbed_provider/key_management.rs
```diff
@@ -43,13 +43,13 @@ fn create_key_id(
...
+) -> Result<key::psa_key_id_t> {
+    let mut key_id = rand::random::<key::psa_key_id_t>();
     while local_ids_handle.contains(&key_id)
-        || key_id == 0
-        || key_id > PSA_MAX_PERSISTENT_KEY_IDENTIFIER
+        || key_id < constants::PSA_KEY_ID_USER_MIN
+        || key_id > constants::PSA_KEY_ID_USER_MAX
     {
-        key_id = rand::random::<psa_key_id_t>();
+        key_id = rand::random::<key::psa_key_id_t>();
```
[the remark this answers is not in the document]
ionut-arm: I'm not sure if it matters, the handle value on its own doesn't really represent a crypto liability, we could even use a counter

build.rs
```diff
@@ -1,268 +1,3 @@
...
-                .mbed_path
-                .unwrap_or_else(|| env::var("OUT_DIR").unwrap()),
-            mbed_version,
-        );
-        println!("cargo:rustc-link-lib=static=mbedcrypto");
-    }
-
-    Ok(())
-}
+fn main() {}
```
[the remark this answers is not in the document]
sbailey-arm: Yeah, I did remove it at first but ran into compilation errors. @hug-dev found [this](https://github.com/rust-lang/cargo/issues/879#issuecomment-215491052).

src/providers/mbed_provider/key_management.rs
```diff
@@ -135,23 +124,20 @@ impl MbedProvider {
...
+            Ok(_) => Ok(psa_generate_key::Result {}),
+            Err(error) => {
+                remove_key_id(
+                    &key_triple,
+                    key_id,
+                    &mut *store_handle,
+                    &mut local_ids_handle,
+                )?;
+                let error = ResponseStatus::from(error);
+                error!("Generate key status: {}", error);
```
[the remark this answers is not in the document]
sbailey-arm: Ok, thanks for the heads up. I stuck with `error!` because that's what was there before but I'll switch it to `format_error!`.

src/providers/mbed_provider/mod.rs
sbailey-arm: I'm not sure why or how that got there! It definitely shouldn't be.

src/providers/mbed_provider/key_management.rs
```diff
@@ -43,13 +48,12 @@ fn create_key_id(
...
-) -> Result<psa_key_id_t> {
-    let mut key_id = rand::random::<psa_key_id_t>();
-    while local_ids_handle.contains(&key_id)
-        || key_id == 0
-        || key_id > PSA_MAX_PERSISTENT_KEY_IDENTIFIER
-    {
-        key_id = rand::random::<psa_key_id_t>();
+) -> Result<key::psa_key_id_t> {
+    let mut rng = SmallRng::from_entropy();
+    let mut key_id = rng.gen_range(key::PSA_KEY_ID_USER_MIN, key::PSA_KEY_ID_USER_MAX);
```
hug-dev, line 53: `gen_range` might be exclusive of its `high` [parameter](https://docs.rs/rand/0.7.3/rand/trait.Rng.html#method.gen_range) whereas (to be confirmed) `PSA_KEY_ID_USER_MAX` is inclusive. It might then need to be `key::PSA_KEY_ID_USER_MAX + 1` for the second parameter.

src/providers/mbed_provider/mod.rs
```diff
@@ -28,17 +26,13 @@ use uuid::Uuid;
     trivial_casts
 )]
 #[allow(clippy::all)]
-mod psa_crypto_binding {
```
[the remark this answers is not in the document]
hug-dev: And even the one above, at line 21 😋 If we keep it, I am afraid it will apply on to the next `mod`, which might hides somes lints for that module!

sbailey-arm:
> (maybe a few `Safety` as well) but it is all ok.
I did find one more Safety comment that shouldn't be there but I've left the `destroy` one in on purpose as it still uses an `unsafe` block. Should that comment also be removed?

src/providers/mbed_provider/key_management.rs
sbailey-arm: ah yes, my bad. I'd expect `PSA_KEY_ID_USER_MAX` to be inclusive.

hug-dev:
> 
> 
> > (maybe a few `Safety` as well) but it is all ok.
> > I did find one more Safety comment that shouldn't be there but I've left the `destroy` one in on purpose as it still uses an `unsafe` block. Should that comment also be removed?

Oh, my bad then, if it is only on `destroy`, that is fine!

hug-dev:
Versions `0.2.0` of `psa-crypto` and `0.16.0` of `parsec-interface` have just been published on crates.io so you can now use it in this PR!

sbailey-arm:
> I am thinking of a few things as next steps to simplify the provider further:
> 
>     * I think it would be possible to remove the `local_ids` field and replace it with a `AtomicU32` counter?
We could do. How would we deal with holes when keys are destroyed? E.g. Allocate keys 1 to 100, then destroy 34, 67, 78 and 91.
>     * The `key_handle_mutex` might not be needed anymore if updating Mbed Crypto solved the thread-safety issues. If it is still needed (Parsec stress tests fail without it), it might be better suited in the `psa-crypto` crate.
[This](https://github.com/ARMmbed/mbedtls/pull/3313) is the issue for that and its still open so I think we'll hit issues. I can try testing it without though, and if it causes a problem, look into moving it to `psa-crypto`.
>     * I am not sure if the `key_slot_semaphore` is needed at all. It limits the number of open key slots at a time but if that is a constraint of Mbed Crypto we might just return the error
`stress_test` and `normal_tests` pass after removing the semaphore guard request in each operation method so it looks like we can remove it.

hug-dev:
I would prefer to keep the changes of this PR as they are and tackle those different simplifications in future issues/PR.s, if you do not mind. I can open a new issue for those points and we can move the dicsussion there?

sbailey-arm:
> I would prefer to keep the changes of this PR as they are and tackle those different simplifications in future issues/PR.s, if you do not mind. I can open a new issue for those points and we can move the dicsussion there?

Sound good!

PR #188, opened 2020-06-19
Fixed PKCS#11 provieder failing failed_created_key_should_be_removed test

sbailey-arm:
PKCS#11 provider wasn't removing the key ID from the key ID store on two possible failure paths.


ionut-arm:
> Nice find 👌 I don't know how if, in the future, it is possible to use Drop to implement those cleanup automatically, there are so many parameters to that call.

The parameters are one problem, another is the fact that in `Drop` you don't know whether the function succeeded or not.

PR #189, opened 2020-06-22
Implement mitigation 4 of TM

hug-dev:
Adds trace logging whenever requests, responses, mappings or commands
are sent or received.

### DIFF
--- a/src/front/front_end.rs
+++ b/src/front/front_end.rs
@@ -56,35 +56,65 @@ impl FrontEndHandler {
                 return;
             }
         };
+
         // Check if the request was sent without authentication
-        let response = if AuthType::NoAuth == request.header.auth_type {
-            let response = self.dispatcher.dispatch_request(request, None);
-            trace!("dispatch_request egress");
-            response
+        let (app_name, err_response) = if AuthType::NoAuth == request.header.auth_type {
+            (None, None)
         // Otherwise find an authenticator that is capable to authenticate the request
         } else if let Some(authenticator) = self.authenticators.get(&request.header.auth_type) {
             // Authenticate the request
             match authenticator.authenticate(&request.auth) {
                 // Send the request to the dispatcher
                 // Get a response back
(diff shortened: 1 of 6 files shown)


src/front/front_end.rs
```diff
@@ -56,35 +56,65 @@ impl FrontEndHandler {
...
+                }
+            };
+            let response = self.dispatcher.dispatch_request(request, app_name.clone());
+            trace!("dispatch_request egress");
+            response
+        };
+
+        // Serialise the response into bytes
         // Write bytes to stream
         match response.write_to_stream(&mut stream) {
```
ionut-arm, line 105: Apologies for not asking this earlier, but why not just put a `trace!("dispatch_request egress");` just before this match? Or did you not want to write it if the `dispatcher` wasn't called?

src/front/front_end.rs
```diff
@@ -56,35 +56,65 @@ impl FrontEndHandler {
...
+            if crate::utils::GlobalConfig::log_error_details() {
+                if let Some(app_name_string) = app_name.clone() {
+                    info!(
+                        "New request received from application name \"{}\"",
+                        app_name_string
+                    )
+                } else {
+                    info!("New request received without authentication")
+                }
+            };
```
ionut-arm, line 97: Nice 👌

src/key_info_managers/on_disk_manager/mod.rs
```diff
@@ -209,6 +206,12 @@ impl OnDiskKeyInfoManager {
                         ))?,
                     ) {
                         Ok(key_triple) => {
+                            if crate::utils::GlobalConfig::log_error_details() {
```
ionut-arm, line 209: Maybe we should rename the flag to `log_details`

src/front/front_end.rs
hug-dev: Yes so that it is only written when the `dispatcher` has been called.

PR #190, opened 2020-06-23
Removed duplicate macros for sign output size and export pub key size.

sbailey-arm:
Uses newly available functions from `psa-crypto`.


src/providers/mbed_provider/mod.rs
```diff
@@ -21,7 +21,6 @@ use uuid::Uuid;
 mod asym_sign;
 #[allow(dead_code)]
 mod key_management;
-mod utils;
```
ionut-arm, line 24: 🎉

Issue #187, opened 2020-06-18
Further simplification of the Mbed Crypto provider


hug-dev:
In the conversation of #186 :

I am thinking of a few things as next steps to simplify the provider further, I said:

1. I think it would be possible to remove the local_ids field and replace it with a AtomicU32 counter?
2. The key_handle_mutex might not be needed anymore if updating Mbed Crypto solved the thread-safety issues. If it is still needed (Parsec stress tests fail without it), it might be better suited in the psa-crypto crate.
3. I am not sure if the key_slot_semaphore is needed at all. It limits the number of open key slots at a time but if that is a constraint of Mbed Crypto we might just return the error

to which @sbailey-arm replied:

1. We could do. How would we deal with holes when keys are destroyed? E.g. Allocate keys 1 to 100, then destroy 34, 67, 78 and 91.

2. This is the issue for that and its still open so I think we'll hit issues. I can try testing it without though, and if it causes a problem, look into moving it to psa-crypto.

3. stress_test and normal_tests pass after removing the semaphore guard request in each operation method so it looks like we can remove it.

Let's continue the discussion here and see how to address those points.

cc @ionut-arm

hug-dev:
> We could do. How would we deal with holes when keys are destroyed? E.g. Allocate keys 1 to 100, then destroy 34, 67, 78 and 91.

You are right that there will be holes but I think the window of values is so big that it should not be a problem when Parsec is used. Maybe only in stress test?

> This is the issue for that and its still open so I think we'll hit issues. I can try testing it without though, and if it causes a problem, look into moving it to psa-crypto.

I think it is better placed in `psa-crypto`. The goal of PSA Crypto in the end is to be totally agnostic of the different implementations, but it looks like now we are tying more and more things to Mbed Crypto. I think that is fine for now.
One thing to note as well is that `Mutex` won't be in the `no-std` feature.

> stress_test and normal_tests pass after removing the semaphore guard request in each operation method so it looks like we can remove it.

Nice! I think that running out of slots is a valid error to return anyway and not something we should handle in Parsec.

ionut-arm:
> You are right that there will be holes but I think the window of values is so big that it should not be a problem when Parsec is used. Maybe only in stress test?

If we run a stress test that generates 100 keys per second, we'd have to run it for almost 1 year and a half to run out of handles (if we assume that the whole 32 bit space is available).

hug-dev:
[...]

PR #191, opened 2020-06-25
Changed local_ids for Atomic counter and removed key_slot_semaphore.

sbailey-arm:
`key_slot_semaphore` removed - #187 1.

`local_ids` changed to an `AtomicU32` that store the current highest key ID including destroyed keys). On load, loops through all stored keys and sets the counter to the max ID. On creating a key ID, adds 1 to the counter. If key store fails and no thread has created another key in the mean time, decrements the counter to the value before the attempt to create a new key ID failed. On key destroy, counter not altered as extra complication deemed unnecessary. #187 2.

### DIFF
--- a/src/providers/mbed_provider/key_management.rs
+++ b/src/providers/mbed_provider/key_management.rs
@@ -1,9 +1,10 @@
 // Copyright 2020 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
-use super::{LocalIdStore, MbedProvider};
+use super::MbedProvider;
 use crate::authenticators::ApplicationName;
 use crate::key_info_managers;
 use crate::key_info_managers::{KeyInfo, KeyTriple, ManageKeyInfo};
+use log::error;
 use log::{info, warn};
 use parsec_interface::operations::psa_key_attributes::Attributes;
 use parsec_interface::operations::{
@@ -12,8 +13,7 @@ use parsec_interface::operations::{
 use parsec_interface::requests::{ProviderID, ResponseStatus, Result};
 use psa_crypto::operations::key_management as psa_crypto_key_management;
 use psa_crypto::types::key;
-use rand::rngs::SmallRng;
-use rand::{Rng, SeedableRng};
+use std::sync::atomic::{AtomicU32, Ordering::Relaxed};
 
 /// Gets a PSA Key ID from the Key Info Manager.
 /// Wrapper around the get method of the Key Info Manager to convert the key ID to the psa_key_id_t
@@ -46,16 +46,23 @@ fn create_key_id(
     key_triple: KeyTriple,
     key_attributes: Attributes,
     store_handle: &mut dyn ManageKeyInfo,
-    local_ids_handle: &mut LocalIdStore,
+    max_current_id: &AtomicU32,
 ) -> Result<key::psa_key_id_t> {
-    let mut rng = SmallRng::from_entropy();
-    let mut key_id = rng.gen_range(key::PSA_KEY_ID_USER_MIN, key::PSA_KEY_ID_USER_MAX + 1);
-
-    while local_ids_handle.contains(&key_id) {
-        key_id = rng.gen_range(key::PSA_KEY_ID_USER_MIN, key::PSA_KEY_ID_USER_MAX + 1);
+    // fetch_add adds 1 to the old value and returns the old value, so add 1 to local value for new ID
+    let new_key_id = max_current_id.fetch_add(1, Relaxed) + 1;
+    if new_key_id > key::PSA_KEY_ID_USER_MAX {
+        // If storing key failed and no other keys were created in the mean time, it is safe to
+        // decrement the key counter.
+        let _ = max_current_id.store(key::PSA_KEY_ID_USER_MAX, Relaxed);
+        error!(
+            "PSA max key ID limit of {} reached",
+            key::PSA_KEY_ID_USER_MAX
+        );
+        return Err(ResponseStatus::PsaErrorInsufficientMemory);
     }
+
     let key_info = KeyInfo {
-        id: key_id.to_ne_bytes().to_vec(),
+        id: new_key_id.to_ne_bytes().to_vec(),
         attributes: key_attributes,
     };
     match store_handle.insert(key_triple.clone(), key_info) {
@@ -63,25 +70,16 @@ fn create_key_id(
             if insert_option.is_some() {
                 warn!("Overwriting Key triple mapping ({})", key_triple);
             }
-            let _ = local_ids_handle.insert(key_id);
-
-            Ok(key_id)
+            Ok(new_key_id)
         }
         Err(string) => Err(key_info_managers::to_response_status(string)),
     }
 }
 
-fn remove_key_id(
-    key_triple: &KeyTriple,
-    key_id: key::psa_key_id_t,
-    store_handle: &mut dyn ManageKeyInfo,
-    local_ids_handle: &mut LocalIdStore,
-) -> Result<()> {
+fn remove_key_id(key_triple: &KeyTriple, store_handle: &mut dyn ManageKeyInfo) -> Result<()> {
+    // ID Counter not affected as overhead and extra complication deemed unnecessary
     match store_handle.remove(key_triple) {
-        Ok(_) => {
-            let _ = local_ids_handle.remove(&key_id);
-            Ok(())
-        }
+        Ok(_) => Ok(()),
         Err(string) => Err(key_info_managers::to_response_status(string)),
     }
 }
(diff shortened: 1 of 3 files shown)


src/providers/mbed_provider/key_management.rs
```diff
@@ -46,42 +45,41 @@ fn create_key_id(
...
-
-    while local_ids_handle.contains(&key_id) {
-        key_id = rng.gen_range(key::PSA_KEY_ID_USER_MIN, key::PSA_KEY_ID_USER_MAX + 1);
+    // fetch_add adds 1 to the old value and returns the old value, so add 1 for new ID
+    let new_key_id = max_current_id.fetch_add(1, Relaxed) + 1;
+    if new_key_id > key::PSA_KEY_ID_USER_MAX {
+        // If storing key failed and no other keys were created in the mean time, it is safe to
+        // decrement the key counter.
+        let _ = max_current_id.compare_and_swap(new_key_id, new_key_id - 1, Relaxed);
+        return Err(ResponseStatus::PsaErrorBadState);
```
sbailey-arm, line 56: Wasn't sure on the error to return if there are no more key IDs available.

hug-dev: I think `BadState` is an error referring to multi-parts operations. It might be better to use `InsufficientMemory`. Also, could you please add an `error!` log to note that we've reached the limit?

src/providers/mbed_provider/key_management.rs
```diff
@@ -46,42 +45,41 @@ fn create_key_id(
...
-    let mut key_id = rng.gen_range(key::PSA_KEY_ID_USER_MIN, key::PSA_KEY_ID_USER_MAX + 1);
-
-    while local_ids_handle.contains(&key_id) {
-        key_id = rng.gen_range(key::PSA_KEY_ID_USER_MIN, key::PSA_KEY_ID_USER_MAX + 1);
+    // fetch_add adds 1 to the old value and returns the old value, so add 1 for new ID
+    let new_key_id = max_current_id.fetch_add(1, Relaxed) + 1;
+    if new_key_id > key::PSA_KEY_ID_USER_MAX {
+        // If storing key failed and no other keys were created in the mean time, it is safe to
+        // decrement the key counter.
+        let _ = max_current_id.compare_and_swap(new_key_id, new_key_id - 1, Relaxed);
```
hug-dev, line 55: As you said in your issue comment, do you think this is always safe?

src/providers/mbed_provider/key_management.rs
```diff
@@ -46,42 +45,41 @@ fn create_key_id(
...
-    local_ids_handle: &mut LocalIdStore,
+    max_current_id: &AtomicU32,
 ) -> Result<key::psa_key_id_t> {
-    let mut rng = SmallRng::from_entropy();
-    let mut key_id = rng.gen_range(key::PSA_KEY_ID_USER_MIN, key::PSA_KEY_ID_USER_MAX + 1);
-
-    while local_ids_handle.contains(&key_id) {
-        key_id = rng.gen_range(key::PSA_KEY_ID_USER_MIN, key::PSA_KEY_ID_USER_MAX + 1);
+    // fetch_add adds 1 to the old value and returns the old value, so add 1 for new ID
+    let new_key_id = max_current_id.fetch_add(1, Relaxed) + 1;
```
hug-dev, line 51: I guess it is fine to not care about overflows as `PSA_KEY_ID_USER_MAX` is equal to the defined value of `0x3FFFFFFF`.

src/providers/mbed_provider/mod.rs
```diff
@@ -104,7 +97,9 @@ impl MbedProvider {
                         let pc_key_id = key::Id::from_persistent_key_id(key_id);
                         match key::Attributes::from_key_id(pc_key_id) {
                             Ok(_) => {
-                                let _ = local_ids_handle.insert(key_id);
+                                if key_id > mbed_provider.id_counter.load(Relaxed) {
+                                    mbed_provider.id_counter.store(key_id, Relaxed);
+                                }
```
hug-dev, line 102: It might be more efficient to find the maximum locally, in a `psa_key_id_t` variable and then store it in the `id_counter` at the end. I believe the atomic operations impose restrictions on compiler optimizations.

src/providers/mbed_provider/key_management.rs
sbailey-arm: There's one problem I've thought of (see second paragraph) but in general, yes. In that particular case, the new ID would be created with the value of PSA_KEY_ID_USER_MAX + 1, and max_current_id would be set to that. If no other thread tries to create a new ID, max_current_id will still be PSA_KEY_ID_USER_MAX + 1, which will then be decremented back down to PSA_KEY_ID_USER_MAX. If a new thread does try to create a new ID, it will increment it to PSA_KEY_ID_USER_MAX + 2. The original thread will then fail the `compare` and not touch PSA_KEY_ID_USER_MAX.

It could overflow though, theoretically there's theoretically a chance that mac_current_id could keep going up if new threads stop any older threads from decrementing it. Maybe it'd be better to just hard reset it to PSA_KEY_ID_USER_MAX if it fails the check? No key will ever be created if max_current_id is at the ID limit so we should be able to set it back.
Thread #1 comes in, increments it to MAX + 1 and realises its too high. Then thread #2 comes in, increments it to MAX + 2 and also realises its too high. Thread #1 runs again and resets to MAX. Then thread #2 does the same (though it's already at MAX).

Decrementing in the other case (line 74) should also be safe. If no threads have created key IDs in the mean time, the compare will match and the counter will be decremented correctly. If a thread has come in in the mean time, the compare will fail and the counter will not be decremented. 

Am I missing something?

src/providers/mbed_provider/mod.rs
sbailey-arm: Ah, didn't realise there would be much overhead in just using the atomic variable. I will change it.

src/providers/mbed_provider/mod.rs
```diff
@@ -70,9 +68,8 @@ impl MbedProvider {
         }
         let mbed_provider = MbedProvider {
             key_info_store,
-            local_ids: RwLock::new(HashSet::new()),
             key_handle_mutex: Mutex::new(()),
-            key_slot_semaphore: Semaphore::new(PSA_KEY_SLOT_COUNT),
+            id_counter: AtomicU32::new(key::PSA_KEY_ID_USER_MIN),
```
ionut-arm, line 72: This is wrong - the value of the counter has to be persisted. Which makes me wonder how we'll manage that........

src/providers/mbed_provider/key_management.rs
ionut-arm: > Maybe it'd be better to just hard reset it to PSA_KEY_ID_USER_MAX if it fails the check? No key will ever be created if max_current_id is at the ID limit so we should be able to set it back.

Yeah, just commented the same thing in the issue, that sounds sensible!

As for the second decrement - it works, but is it needed? I'd say that having an operation that isn't deterministic isn't a good idea (regardless of correctness) and if you end up crashing through the handle limit then you have bigger issues on your system.

src/providers/mbed_provider/mod.rs
ionut-arm: Oh, we can implement `Drop` on the `MbedProvider` and write it somewhere - sounds like we need some new "service" for persisting arbitrary data. Then we can read it back when the service is started

sbailey-arm: Finding the highest currently used ID is taken care of (lines 100 to 102). While looping through all keys in the store to check they are still valid, it keeps track of the highest ID in use.

src/providers/mbed_provider/key_management.rs
ionut-arm: Is it a good idea to return `InsufficientMemory`? I guess in conjuction with a log message it would be easy to understand what's going on, but based on the error code alone it might be confusing.

src/providers/mbed_provider/mod.rs
ionut-arm: Ah! makes sense, didn't make the correlation between the two!

src/providers/mbed_provider/key_management.rs
sbailey-arm: Yeah, I saw 😋 

Ok. Yeah if you're going that high, saving a key ID here and there isn't going to make a difference, I'll remove it.

src/providers/mbed_provider/key_management.rs
sbailey-arm: I was thinking about adding a new error but didn't want to touch `parsec-interface` without discussing first.

ionut-arm: I think the problem with a new response code like that is we shouldn't be exposing internal issues (or, not ones that mean nothing to a client). `InsufficientMemory` is probably ok if we document what it means in our context - we just add an entry [here](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_generate_key.html#specific-response-status-codes)

src/providers/mbed_provider/key_management.rs
sbailey-arm: Better safe than sorry or overkill? 😉

PR #192, opened 2020-06-29
Modify socket path

hug-dev:
Following the secure deployment guide, modify the socket path to
/tmp/parsec/parsec.sock
Removes socket activation from the systemd unit files. Modifies the
permission of the created socket so that all users who can access the
/tmp/parsec folder can use it.
Adds back the `Cargo.lock` file which was removed.


systemd-daemon/parsec.service
```diff
@@ -1,9 +1,10 @@
...
-Description=PARSEC Service
-Documentation=https://github.com/parallaxsecond/parsec
+Description=Parsec Service
+Documentation=https://parallaxsecond.github.io/parsec-book/parsec_service/install_parsec_linux.html
 
 [Service]
-Type=notify
-NonBlocking=true
-Environment=RUST_LOG=info
+WorkingDirectory=/home/parsec/
```
anta5010, line 6: Should we depend on existence of "parsec" user account and its home? What's the reason?

In the Yocto layer, for example, I install parsec binary into ${bindir}/parsec (although we might want to install it into ${sbindir}) and parsec config into /etc/parsec/config.toml

hug-dev: > Should we depend on existence of "parsec" user account and its home? What's the reason?

This PR is a draft depending on if we accept the secure deployment guidelines described in #41
The existence of the `parsec` user and its home is an easy way to ensure that the above guidelines are respected while installing Parsec as an user systemd service. The systemd files can still be customised by the knowledgable admin during deployment to make it suit better to the target environment.

PR #193, opened 2020-06-30
Updated Parsec to use latest parsec-interface (0.17.0)

sbailey-arm:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ name = "parsec"
 path = "src/bin/main.rs"
 
 [dependencies]
-parsec-interface = "0.16.0"
+parsec-interface = "0.17.0"
 rand = { version = "0.7.2", features = ["small_rng"] }
 base64 = "0.10.1"
 uuid = "0.7.4"
@@ -41,6 +41,7 @@ version = "3.0.0"
 hex = "0.4.2"
 picky = "5.0.0"
 psa-crypto = { version = "0.2.1" , default-features = false, features = ["with-mbed-crypto"], optional = true }
+zeroize = { version = "1.1.0", features = ["zeroize_derive"] }
 
 [dev-dependencies]
 ring = "0.16.12"
--- a/src/authenticators/direct_authenticator/mod.rs
+++ b/src/authenticators/direct_authenticator/mod.rs
@@ -13,6 +13,7 @@ use super::Authenticate;
 use log::error;
 use parsec_interface::requests::request::RequestAuth;
 use parsec_interface::requests::{ResponseStatus, Result};
+use parsec_interface::secrecy::ExposeSecret;
 use std::str;
 
 #[derive(Copy, Clone, Debug)]
@@ -20,11 +21,11 @@ pub struct DirectAuthenticator;
 
 impl Authenticate for DirectAuthenticator {
     fn authenticate(&self, auth: &RequestAuth) -> Result<ApplicationName> {
-        if auth.is_empty() {
+        if auth.buffer.expose_secret().is_empty() {
             error!("The direct authenticator does not expect empty authentication values.");
             Err(ResponseStatus::AuthenticationError)
         } else {
-            match str::from_utf8(auth.bytes()) {
+            match str::from_utf8(auth.buffer.expose_secret()) {
                 Ok(str) => Ok(ApplicationName(String::from(str))),
                 Err(_) => {
                     error!("Error parsing the authentication value as a UTF-8 string.");
@@ -47,7 +48,7 @@ mod test {
         let authenticator = DirectAuthenticator {};
 
         let app_name = "app_name".to_string();
-        let req_auth = RequestAuth::from_bytes(app_name.clone().into_bytes());
+        let req_auth = RequestAuth::new(app_name.clone().into_bytes());
 
         let auth_name = authenticator
             .authenticate(&req_auth)
@@ -60,7 +61,7 @@ mod test {
     fn failed_authentication() {
         let authenticator = DirectAuthenticator {};
         let status = authenticator
-            .authenticate(&RequestAuth::from_bytes(vec![0xff; 5]))
+            .authenticate(&RequestAuth::new(vec![0xff; 5]))
             .expect_err("Authentication should have failed");
 
         assert_eq!(status, ResponseStatus::AuthenticationError);
@@ -70,7 +71,7 @@ mod test {
     fn empty_auth() {
         let authenticator = DirectAuthenticator {};
         let status = authenticator
-            .authenticate(&RequestAuth::from_bytes(Vec::new()))
+            .authenticate(&RequestAuth::new(Vec::new()))
             .expect_err("Empty auth should have failed");
 
         assert_eq!(status, ResponseStatus::AuthenticationError);
--- a/src/providers/mbed_provider/key_management.rs
+++ b/src/providers/mbed_provider/key_management.rs
@@ -11,6 +11,7 @@ use parsec_interface::operations::{
     psa_destroy_key, psa_export_public_key, psa_generate_key, psa_import_key,
 };
 use parsec_interface::requests::{ProviderID, ResponseStatus, Result};
+use parsec_interface::secrecy::ExposeSecret;
 use psa_crypto::operations::key_management as psa_crypto_key_management;
 use psa_crypto::types::key;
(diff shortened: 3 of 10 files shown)


Cargo.toml
```diff
@@ -41,6 +41,7 @@ version = "3.0.0"
 hex = "0.4.2"
 picky = "5.0.0"
 psa-crypto = { version = "0.2.0" , default-features = false, features = ["with-mbed-crypto"], optional = true }
+secrecy = { version = "0.6.0", features = ["serde"] }
```
ionut-arm, line 44: The interface already exports a version of this crate (`parsec_interface::secrecy`) - it's better to use that one to avoid potential version mismatches in the future.

src/authenticators/direct_authenticator/mod.rs
```diff
@@ -13,18 +13,19 @@ use super::Authenticate;
...
 impl Authenticate for DirectAuthenticator {
     fn authenticate(&self, auth: &RequestAuth) -> Result<ApplicationName> {
-        if auth.is_empty() {
+        if auth.buffer.expose_secret().is_empty() {
             error!("The direct authenticator does not expect empty authentication values.");
             Err(ResponseStatus::AuthenticationError)
         } else {
-            match str::from_utf8(auth.bytes()) {
+            match str::from_utf8(auth.buffer.expose_secret()) {
                 Ok(str) => Ok(ApplicationName(String::from(str))),
```
ionut-arm, line 29: Since you're doing this change anyway, I'm wondering if it would make sense to put the `String` inside `ApplicationName` into a `Secret` too. The app name sort of is a client-level secret...

I guess the alternative would be to derive the `zeroize(drop)` when the change for that is made at some point.

Thoughts?

src/providers/mbed_provider/key_management.rs
```diff
@@ -212,7 +217,9 @@ impl MbedProvider {
         let export_length = psa_crypto_key_management::export_public(id, &mut buffer)?;
 
         buffer.resize(export_length, 0);
-        Ok(psa_export_public_key::Result { data: buffer })
+        Ok(psa_export_public_key::Result {
+            data: buffer.into(),
```
ionut-arm, line 221: Would these changes need to be made in the other providers too?

sbailey-arm: Probably. I can't test all providers, resolving merge conflict and making tweaks for your first comment so Github can run CI.

sbailey-arm: Setup my VM for TPM now so it'll build.

You're right, TPM and PKCS11 providers need updating. Hit a problem with TPM - it uses `SignatureData::EcdsaSignature` from `tss-esapi` which uses `Vec<u8>` to store the two parts of the signature. Do we have to accept that those can't be changed to `Zeroize`, or do we have other options?

sbailey-arm: Ah, you and @hug-dev are the authors of that, so I guess we don't have to just accept it 🎉

ionut-arm: > Do we have to accept that those can't be changed to Zeroize, or do we have other options?

That can be done in a best-effort way for now, zeroizing when/where we can, and we can go and update the `tss-esapi` crate to help with that later. Ideally we should be treating this data consistently across the stack (well, the parts that we control anyway). After I'm done with the client I'll look at zeroizing in the TPM crate.

src/authenticators/direct_authenticator/mod.rs
hug-dev: I am wondering if the application name, similarly to the key name, is a secret. In our TM, it's only the authentication token which is one. I don't know if an attacker could do anything with this information.

ionut-arm: Which made me realise something - we need to somehow determine which applications use which authenticator. Imagine the following scenario: a Parsec service that has a JWT token authenticator and a direct authenticator (for some reason) - an attacker finds the name of an application and issues requests using direct authentication. Essentially, if there are two or more authenticators, an attacker will always target the easier one to break, if possible.

I think it's right not to consider the app name a secret (even though for direct authentication that *is* the auth token).

src/providers/pkcs11_provider/key_management.rs
```diff
@@ -276,16 +277,17 @@ impl Pkcs11Provider {
...
-            format_error!("Failed to parse RsaPublicKey data", e);
-            remove_key_id(
-                &key_triple,
-                key_id,
-                &mut *store_handle,
-                &mut local_ids_handle,
-            )?;
-            Err(ResponseStatus::PsaErrorInvalidArgument)
-        })?;
+        let public_key: RsaPublicKey = picky_asn1_der::from_bytes(op.data.expose_secret())
```
hug-dev, line 280: That is very interesting because `Secret` now allows use to see more precisely where we share our assets with dependencies. Currently, we clearly state in our TM that "depencencies are not checked for vulnerabilities". If we want to check that at some point, we can start by grepping whenever we use `expose_secret`. For example here, what do we know what `picky_asn1_der::from_bytes` is doing with our asset?

src/providers/tpm_provider/utils.rs
```diff
@@ -233,13 +234,13 @@ pub fn signature_data_to_bytes(data: SignatureData, key_attributes: Attributes)
 }
 
 pub fn parsec_to_tpm_signature(
-    data: Vec<u8>,
+    data: &Zeroizing<Vec<u8>>,
```
hug-dev, line 237: Does it work if you remove the reference? Since `Zeroizing` implements `Deref`, I think that `to_vec` takes the reference automatically.

src/providers/tpm_provider/utils.rs
```diff
@@ -233,13 +234,13 @@ pub fn signature_data_to_bytes(data: SignatureData, key_attributes: Attributes)
...
     key_attributes: Attributes,
     signature_alg: AsymmetricSignature,
 ) -> Result<Signature> {
     Ok(Signature {
         scheme: convert_asym_scheme_to_tpm(Algorithm::AsymmetricSignature(signature_alg))?,
-        signature: bytes_to_signature_data(data, key_attributes)?,
+        signature: bytes_to_signature_data(data.to_vec(), key_attributes)?,
```
hug-dev, line 243: That is one of those place in the TPM crate where we need to check zeroing, or only take a reference if possible.

src/providers/tpm_provider/utils.rs
ionut-arm: I think it's actually ok to pass the `Zeroizing` ownership in here, we don't need the signature data in the buffer after this conversion. Ideally the whole `Zeroizing` will be passed into the TSS signature crate.

src/providers/tpm_provider/utils.rs
ionut-arm: We can actually send the `Zeroizing` down in `bytes_to_signature_data` where we convert to the TSS type, no need to clone the buffers.

ionut-arm: Alternatively we can use `AsMut<[u8]>`

sbailey-arm: Yeah, this one I'll change `bytes_to_signature_data` to take a `Zeroizing`

Cargo.toml
```diff
@@ -41,6 +41,7 @@ version = "3.0.0"
 hex = "0.4.2"
 picky = "5.0.0"
 psa-crypto = { version = "0.2.1" , default-features = false, features = ["with-mbed-crypto"], optional = true }
+zeroize = { version = "1.1.0", features = ["zeroize_derive"] }
```
sbailey-arm, line 44: Should this be changed to a public export in `parsec-interface`, to match `secrecy`?

ionut-arm: No, not really - the reasoning for only exporting `secrecy` despite using types from both in the operations is that for `Secret` you *need* to have access to `secrecy` in order to create or use an object, you can't just do `.into()` from the right type (like for `Zeroizing`). Also, if you want to access the data in `Secret` you need the `ExposeSecret` trait, whereas `Zeroizing` has `AsRef` and `AsMut` implemented so it can be used easily.

There is a point to be made that someone seeing `Zeroizing` might try and do `Zeroizing::new(...)` - don't know how we could discourage that. Maybe in the interface documentation.

PR #196, opened 2020-07-06
Added asymmetric encrypt and decrypt to Mbed Crypto provider

sbailey-arm:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ name = "parsec"
 path = "src/bin/main.rs"
 
 [dependencies]
-parsec-interface = "0.17.0"
+parsec-interface = "0.18.0"
 rand = { version = "0.7.2", features = ["small_rng"] }
 base64 = "0.10.1"
 uuid = "0.7.4"
@@ -40,7 +40,7 @@ derivative = "2.1.1"
 version = "3.0.0"
 hex = "0.4.2"
 picky = "5.0.0"
-psa-crypto = { version = "0.2.1" , default-features = false, features = ["with-mbed-crypto"], optional = true }
+psa-crypto = { version = "0.2.2" , default-features = false, features = ["with-mbed-crypto"], optional = true }
 zeroize = { version = "1.1.0", features = ["zeroize_derive"] }
 picky-asn1-x509 = { version = "0.1.0", optional = true }
 
--- a/e2e_tests/tests/per_provider/normal_tests/asym_encryption.rs
+++ b/e2e_tests/tests/per_provider/normal_tests/asym_encryption.rs
@@ -1,0 +1,200 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use e2e_tests::TestClient;
+use parsec_client::core::interface::requests::{Opcode, ResponseStatus};
+use rand::rngs::OsRng;
+use rsa::{PaddingScheme, PublicKey, RSAPublicKey};
+
+const PLAINTEXT_MESSAGE: [u8; 32] = [
+    0x69, 0x3E, 0xDB, 0x1B, 0x22, 0x79, 0x03, 0xF4, 0xC0, 0xBF, 0xD6, 0x91, 0x76, 0x37, 0x84, 0xA2,
+    0x94, 0x8E, 0x92, 0x50, 0x35, 0xC2, 0x8C, 0x5C, 0x3C, 0xCA, 0xFE, 0x18, 0xE8, 0x81, 0x37, 0x78,
+];
+
+const PRIVATE_KEY: &str = "MIICWwIBAAKBgQCd+EKeRmZCKLmg7LasWqpKA9/01linY75ujilf6v/Kb8UP9r/E\
+cO75Pvi2YPnYhBadmVOVxMOqS2zmKm1a9VTegT8dN9Unf2s2KbKrKXupaQTXcrGG\
+SB/BmHeWeiqidEMw7i9ysjHK4KEuacmYmZpvKAnNWMyvQgjGgGNpsNzqawIDAQAB\
+AoGAcHlAxXyOdnCUqpWgAtuS/5v+q06qVJRaFFE3+ElT0oj+ID2pkG5wWBqT7xbh\
+DV4O1CtFLg+o2OlXIhH3RpoC0D0x3qfvDpY5nJUUhP/w7mtGOwvB08xhXBN2M9fk\
+PNqGdrzisvxTry3rp9qDduZlv1rTCsx8+ww3iI4Q0coD4fECQQD4KAMgIS7Vu+Vm\
+zQmJfVfzYCVdr4X3Z/JOEexb3eu9p1Qj904sLu9Ds5NO7atT+qtDYVxgH5kQIrKk\
+mFNAx3NdAkEAovZ+DaorhkDiL/gFVzwoShyc1A6AWkH791sDlns2ETZ1WwE/ccYu\
+uJill/5XA9RKw6whUDzzNTsv7bFkCruAZwJARP5y6ALxz5DfFfbZuPU1d7/6g5Ki\
+b4fh8VzAV0ZbHa6hESLYBCbEdRE/WolvwfiGl0RBd6QxXTAYdPS46ODLLQJARrz4\
+urXDbuN7S5c9ukBCvOjuqp4g2Q0LcrPvOsMBFTeueXJxN9HvNfIM741X+DGOwqFV\
+VJ8gc1rd0y/NXVtGwQJAc2w23nTmZ/olcMVRia1+AFsELcCnD+JqaJ2AEF1Ng6Ix\
+V/X2l32v6t3B57sw/8ce3LCheEdqLHlSOpQiaD7Qfw==";
+
+#[allow(dead_code)]
+const PUBLIC_KEY: &str = "MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQCd+EKeRmZCKLmg7LasWqpKA9/0\
+1linY75ujilf6v/Kb8UP9r/EcO75Pvi2YPnYhBadmVOVxMOqS2zmKm1a9VTegT8d\
+N9Unf2s2KbKrKXupaQTXcrGGSB/BmHeWeiqidEMw7i9ysjHK4KEuacmYmZpvKAnN\
+WMyvQgjGgGNpsNzqawIDAQAB";
+
+const ENCRYPTED_MESSAGE: &str =
+"ebr0Q/lPf+905a66RjABlZJ8Xl9ZpTHrwVAHd1+sKOT0G4uCUd+q2mpKGljODiMn5gvMj8aMjTOZUROBmrZQpCnB8GCqpGtEOjJtpJy5AdfMTK+QZVvTnvEia1NTjYIoRNCSfFXTQP/ZsAfq2ViiymqwYXM270pHxS3TvBdQH9A=";
+
+const ORIGINAL_MESSAGE: &str = "This is a test!";
+
+#[test]
+fn simple_asym_encrypt_rsa_pkcs() {
+    let key_name = String::from("asym_encrypt_and_decrypt_rsa_pkcs");
+    let mut client = TestClient::new();
+
+    if !client.is_operation_supported(Opcode::PsaAsymmetricEncrypt) {
+        return;
+    }
+
+    client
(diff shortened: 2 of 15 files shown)


sbailey-arm:
This PR won't pass CI until [`parsec-interface-rs`](https://github.com/parallaxsecond/parsec-interface-rs/pull/63) and [`parsec-client-rust`](https://github.com/parallaxsecond/parsec-client-rust/pull/36) have been updated on creates.io. Will mark as ready for review once they have been updated 😃.

src/providers/mbed_provider/asym_encryption.rs
```diff
@@ -0,0 +1,101 @@
...
+            salt_buff,
+            &mut ciphertext,
+        ) {
+            Ok(output_size) => {
+                ciphertext.resize(output_size, 0);
+                Ok(psa_asymmetric_encrypt::Result { ciphertext })
+            }
+            Err(error) => {
+                let error = ResponseStatus::from(error);
+                format_error!("Encrypt status: {}", error);
```
ionut-arm, line 54: The `format_error` macro doesn't need the `{}` part of the "format string", that gets added inside the macro!

src/providers/mbed_provider/asym_encryption.rs
```diff
@@ -0,0 +1,101 @@
...
+        let mut plaintext = vec![0u8; buffer_size];
+
+        match asym_encryption::decrypt(id, op.alg, &op.ciphertext, salt_buff, &mut plaintext) {
+            Ok(output_size) => {
+                plaintext.resize(output_size, 0);
+                Ok(psa_asymmetric_decrypt::Result { plaintext })
+            }
+            Err(error) => {
+                // When dropped, will zero buffer incase anything was written before error
+                let _ = Secret::new(plaintext);
```
ionut-arm, line 94: Easier/quicker to just `plaintext.zeroize()` here (though you need to import the `Zeroize` trait).

e2e_tests/tests/per_provider/normal_tests/asym_encryption.rs
```diff
@@ -0,0 +1,115 @@
...
+    let ciphertext = rsa_pub_key.encrypt(&mut OsRng, PaddingScheme::new_pkcs1v15_encrypt(), &PLAINTEXT_MESSAGE).unwrap();
+
+    let plaintext = client.asymmetric_decrypt_message_with_rsapkcs1v15(
+        key_name.clone(),
+        ciphertext,
+    ).unwrap();
+
+    assert_eq!(&PLAINTEXT_MESSAGE[..], &plaintext[..]);
+
+}
```
ionut-arm, line 114: 🙆‍♂️ niiiiiice!

src/providers/mbed_provider/asym_encryption.rs
sbailey-arm: huh, I didn't know you could do that 🤯! How does it work? `plaintext` is a regular `Vec<u8>' and I can't see anything obvious in the documentation that states `Zeroizable` is implemented for `Vec<u8>`. Is it `u8` implements `DefaultIsZeros`, so implements `Zeroizable` in the blanket implementation and `Vec<u8>` 'inherits' (derefs?) that?

ionut-arm: [`Zeroize` is implemented on `Vec`s](https://docs.rs/zeroize/1.1.0/zeroize/trait.Zeroize.html#impl-Zeroize-for-Vec%3CZ%3E) - it's actually what you were using in the other PR when doing `plaintext.zeroize()` for the protobuf types - those types don't use `Zeroizing` or anything from the crate

sbailey-arm: Oh yeah... 🤦‍♂️

Cargo.toml
```diff
@@ -40,8 +40,9 @@ derivative = "2.1.1"
 version = "3.0.0"
 hex = "0.4.2"
 picky = "5.0.0"
-psa-crypto = { version = "0.2.1" , default-features = false, features = ["with-mbed-crypto"], optional = true }
+psa-crypto = { version = "0.2.2" , default-features = false, features = ["with-mbed-crypto"], optional = true }
 zeroize = { version = "1.1.0", features = ["zeroize_derive"] }
+secrecy = { version = "0.6.0", features = ["serde"] }
```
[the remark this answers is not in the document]
sbailey-arm: Nope, I have removed it. Wonder if there's a `clippy` setting to catch that...

src/providers/mbed_provider/asym_encryption.rs
```diff
@@ -0,0 +1,101 @@
...
+            .lock()
+            .expect("Grabbing key handle mutex failed");
+        let id = key::Id::from_persistent_key_id(key_id);
+        let key_attributes = key::Attributes::from_key_id(id)?;
+
+        op.validate(key_attributes)?;
+        let salt_buff = match &op.salt {
+            Some(salt) => Some(salt.as_slice()),
+            None => None,
+        };
```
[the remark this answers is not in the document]
sbailey-arm: I don't think that's quite it. `op.salt` is an `Option<Zeroizing<Vec<u8>>>` and I need `salt_buff` to be `Option<&[u8]>`. There may still be some more concise way of writing it that I'm not aware of though.

src/providers/mbed_provider/asym_encryption.rs
sbailey-arm: Ah, that's why the log outputs have been a bit odd...!

sbailey-arm: What about in this case?
```Rust
format_error!(
       "Error {} when opening a persistent Mbed Crypto key.",
       e
);
```

ionut-arm: Ooops! I didn't notice that one when it went in... I wrote `format_error` to always log in the same format "Brief description of problem; error: {}" - the last bit is only written when the config flag for verbose logging is on. You can probably change that to `("Failed to open persistent Mbed Crypto key", e)`

src/providers/mbed_provider/asym_encryption.rs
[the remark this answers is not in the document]
ionut-arm: Huh! I thought that since `Zeroizing` implements `Deref`, it would work - actually, it seems `as_deref` is what I should've recommended, though it doesn't solve the whole problem either - `op.salt.as_deref()` yields `Option<&Vec<u8>>`

Cargo.toml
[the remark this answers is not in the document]
ionut-arm: [WIP](https://github.com/rust-lang/rust/pull/73945) it seems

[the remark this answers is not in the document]
est31: That PR only provides the functionality of cargo-udeps in upstream tools. You can install it today!

e2e_tests/tests/per_provider/normal_tests/asym_encryption.rs
```diff
@@ -0,0 +1,157 @@
...
+
+#[test]
+fn simple_asym_encrypt_rsa_pkcs() {
+    let key_name = String::from("asym_encrypt_and_decrypt_rsa_pkcs");
+    let mut client = TestClient::new();
+
+    if client.provider() != Some(ProviderID::MbedCrypto) {
+        // Not supported by current provider
+        return;
+    }
```
sbailey-arm, line 22: Is there a neater way of doing this? 
Would be nice if the tests could check which operations the current provider supports and only test those.

ionut-arm: You can always put an `#[cfg(feature = "mbed-crypto-provider")]` on the test

ionut-arm: Ah, this is in the e2e crate... Hmmm

ionut-arm: I don't think the `if` is 100% correct anyway, because you could have multiple providers set up and the client chooses a different one, doesn't mean the Mbed Crypto one isn't available. Maybe we should change the `set_provider` method to check if the provider actually exists, and returns an error if not. Then we can check on that.

sbailey-arm: > client chooses a different one

Are the tests not done in isolation? As in, there will only ever be one provider at once during testing?

> Maybe we should change the set_provider method (on the test client) to check if the provider actually exists, and returns an error if not. Then we can check on that.

So have `set_provider` check `basic_client.list_providers` to see if that provider exists (in this case, `MbedCrypto`) and if it does, it returns `Ok` and we can run the tests?

ionut-arm: > Are the tests not done in isolation? As in, there will only ever be one provider at once during testing?

Yeah, that's true, this is in the per_provider test thing - forget that.

I think we can do it in another way, actually, makes it more scalable - call `list_opcodes` and check if the operation you want is supported. If it isn't, `return`. It's not ideal because it will still appear as a test even when it's not supported, but there's no need to do anything to enable the test when you implement it in some new provider.

sbailey-arm: > I think we can do it in another way, actually, makes it more scalable - call list_opcodes and check if the operation you want is supported. If it isn't, return. It's not ideal because it will still appear as a test even when it's not supported, but there's no need to do anything to enable the test when you implement it in some new provider.

That sounds more like what I was thinking. The other two solutions don't deal with the fact that the tests are still run either. Could do with a way of forcing text output even if `--show-output` isn't specified, so we can at least say 'Hey, this provider doesn't actually support this operation'. I can have a poke tomorrow.

Cargo.toml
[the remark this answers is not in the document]
hug-dev: Thanks @est31! We should definitely add that on our CIs!

src/providers/mbed_provider/asym_encryption.rs
```diff
@@ -0,0 +1,96 @@
...
+use psa_crypto::types::key;
+use zeroize::Zeroize;
+
+impl MbedProvider {
+    pub(super) fn psa_asymmetric_encrypt_internal(
+        &self,
+        app_name: ApplicationName,
+        op: psa_asymmetric_encrypt::Operation,
+    ) -> Result<psa_asymmetric_encrypt::Result> {
+        info!("Mbed Provider - Asym Encrypt");
```
hug-dev, line 19: I think those `info!` are not needed now that we have the `ingress`/`egress` trace everywhere

src/providers/mbed_provider/asym_encryption.rs
```diff
@@ -0,0 +1,94 @@
...
+
+        match asym_encryption::decrypt(id, op.alg, &op.ciphertext, salt_buff, &mut plaintext) {
+            Ok(output_size) => {
+                plaintext.resize(output_size, 0);
+                Ok(psa_asymmetric_decrypt::Result {
+                    plaintext: plaintext.into(),
+                })
+            }
+            Err(error) => {
+                plaintext.zeroize();
```
hug-dev, line 87: Ah good point! Do you think, even in case of error, it might contain some decrypted data? Does the spec say anything about it?

src/providers/mbed_provider/asym_encryption.rs
sbailey-arm: Should they be removed from the other operations too?

hug-dev: If you don't mind 👀

src/providers/mbed_provider/asym_encryption.rs
sbailey-arm: hmm it doesn't mention it in the spec that I can see. A quick look at the `mbedtls` code shows that it [only copies into plaintext buffer if successful](https://github.com/ARMmbed/mbedtls/blob/527b87890db72da39f6185f977b8230655a42ddd/library/rsa.c#L1712).

hug-dev: Yes I reckon that is the safe, default, behaviour that you would expect. We can remove the `zeroize` here I think.

src/providers/pkcs11_provider/asym_sign.rs
```diff
@@ -25,8 +25,6 @@ impl Pkcs11Provider {
         app_name: ApplicationName,
         op: psa_sign_hash::Operation,
     ) -> Result<psa_sign_hash::Result> {
-        info!("Pkcs11 Provider - Asym Sign");
```
hug-dev, line 28: Nice! Thanks for removing those as well.

PR #201, opened 2020-07-08
Migrated uses of a locally declared RsaPublic key to new create picky-asn1-x509

sbailey-arm:
The folks over at Devolutions have [put together a create with some useful types in it](https://github.com/Devolutions/picky-rs/issues/12#issuecomment-654877827). We use `RsaPublicKey` at various places in Parsec so I've moved them over from local declarations of an `RsaPublicKey` to `picky-asn1-x509::RSAPublicKey`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ hex = "0.4.2"
 picky = "5.0.0"
 psa-crypto = { version = "0.2.1" , default-features = false, features = ["with-mbed-crypto"], optional = true }
 zeroize = { version = "1.1.0", features = ["zeroize_derive"] }
+picky-asn1-x509 = { version = "0.1.0", optional = true }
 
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -42,6 +42,7 @@ hex = "0.4.2"
 picky = "5.0.0"
 psa-crypto = { version = "0.2.1" , default-features = false, features = ["with-mbed-crypto"], optional = true }
 zeroize = { version = "1.1.0", features = ["zeroize_derive"] }
+picky-asn1-x509 = { version = "0.1.0", optional = true }
```
hug-dev, line 45: Could you also add it in the related section in the `README.md` file? we try to do this for all dependencies that we use. Tbf I don't know if that still makes sense, but let's keep the habit for now 😄

PR #204, opened 2020-07-13
Added PsaExportKey

sbailey-arm:

### DIFF
--- a/e2e_tests/src/lib.rs
+++ b/e2e_tests/src/lib.rs
@@ -17,7 +17,7 @@ use parsec_client::core::interface::operations::psa_algorithm::{
     Algorithm, AsymmetricEncryption, AsymmetricSignature, Hash,
 };
 use parsec_client::core::interface::operations::psa_key_attributes::{
-    Attributes, Lifetime, Policy, Type, UsageFlags,
+    Attributes, Lifetime, Policy, Type, UsageFlags, EccFamily,
 };
 use parsec_client::core::interface::requests::{Opcode, ProviderID, ResponseStatus, Result};
 use parsec_client::core::secrecy::{ExposeSecret, Secret};
@@ -222,6 +222,34 @@ impl TestClient {
         )
     }
 
+    pub fn generate_ecc_key_pair_secpk1_deterministic_ecdsa_sha256(&mut self, key_name: String) -> Result<()> {
+        self.generate_key(
+            key_name,
+            Attributes {
+                lifetime: Lifetime::Persistent,
+                key_type: Type::EccKeyPair {curve_family: EccFamily::SecpK1},
+                bits: 256,
+                policy: Policy {
+                    usage_flags: UsageFlags {
+                        sign_hash: true,
+                        verify_hash: false,
+                        sign_message: true,
+                        verify_message: false,
+                        export: true,
+                        encrypt: false,
+                        decrypt: false,
+                        cache: false,
+                        copy: false,
+                        derive: false,
+                    },
+                    permitted_algorithms: AsymmetricSignature::DeterministicEcdsa {
+                        hash_alg: Hash::Sha256.into(),
+                    }.into(),
+                },
+            },
+        )
+    }
+
     /// Imports and creates a key with specific attributes.
     pub fn import_key(
         &mut self,
@@ -245,7 +273,7 @@ impl TestClient {
 
     /// Import a 1024 bit RSA key pair
     /// The key pair can only be used for encryption and decryption with RSA PKCS 1v15
-    pub fn import_rsa_key_pair(&mut self, key_name: String, data: Vec<u8>) -> Result<()> {
+    pub fn import_rsa_key_pair_for_encryption(&mut self, key_name: String, data: Vec<u8>) -> Result<()> {
         self.import_key(
             key_name,
             Attributes {
@@ -258,7 +286,7 @@ impl TestClient {
                         verify_hash: false,
                         sign_message: false,
                         verify_message: true,
-                        export: false,
+                        export: true,
                         encrypt: true,
                         decrypt: true,
                         cache: false,
@@ -303,6 +331,13 @@ impl TestClient {
             },
             data,
         )
+    }
+
+    /// Exports a key
+    pub fn export_key(&mut self, key_name: String) -> Result<Vec<u8>> {
+        self.basic_client
+            .psa_export_key(key_name)
+            .map_err(convert_error)
     }
 
     /// Exports a public key.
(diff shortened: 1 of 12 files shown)


sbailey-arm:
Won't pass CI until releases are made of the dependency crates.

src/providers/mbed_provider/key_management.rs
```diff
@@ -190,7 +189,7 @@ impl MbedProvider {
 
         let id = key::Id::from_persistent_key_id(key_id);
         let key_attributes = key::Attributes::from_key_id(id)?;
-        let buffer_size = key_attributes.export_key_output_size()?;
+        let buffer_size = key_attributes.export_public_key_output_size()?;
```
sbailey-arm, line 192: Thought I'd point out, this change doesn't actually affect anything. `export_key_output_size` and `export_publid_key_output_size` both call the same function in `psa-crypto`. But since we have a `public` version, I thought we should use it 😉.

e2e_tests/tests/per_provider/normal_tests/export_key.rs
```diff
@@ -0,0 +1,211 @@
...
+fn check_public_rsa_export_format() -> Result<()> {
+    let mut client = TestClient::new();
+
+    if !client.is_operation_supported(Opcode::PsaExportKey) {
+        return Ok(());
+    }
+
+    let key_name = String::from("check_public_rsa_export_format");
+    client.generate_rsa_sign_key(key_name.clone())?;
+    let public_key = client.export_key(key_name)?;
```
ionut-arm, line 92: I'm confused - why does this return just the public key?

e2e_tests/src/lib.rs
```diff
@@ -222,6 +222,32 @@ impl TestClient {
         )
     }
 
+    pub fn generate_ecc_key_pair(&mut self, key_name: String) -> Result<()> {
```
ionut-arm, line 225: I'm guessing this is meant strictly for exporting the key? You can probably rewrite it to do the whole process - generate, export, destroy

e2e_tests/tests/per_provider/normal_tests/export_key.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    if !client.is_operation_supported(Opcode::PsaExportKey) {
+        return;
+    }
+
+    let key_name = String::from("export_key_matches_import");
+
+    let decoded_key = base64::decode(PRIVATE_KEY).unwrap();
+    client
+        .import_rsa_key_pair(key_name.clone(), decoded_key.to_vec())
```
ionut-arm, line 207: I don't remember seeing when this was added, so apologies for the late comment: I think it's worth including in the method name what the key will be usable for - I thought it would be for signing, initially

e2e_tests/tests/per_provider/normal_tests/export_key.rs
sbailey-arm: Because the key is just a public key (this test is just a straight copy pasta from `export_public_key` 😉), so that's all it will export.

From the spec - `For public keys, with key types for which PSA_KEY_TYPE_IS_PUBLIC_KEY() is true, the format is the same as for psa_export_public_key().`

e2e_tests/src/lib.rs
sbailey-arm: Yeah, I'm not sure about all of these separate `generate`, `export`, etc. methods. I was copying the was its already done but I've ended up with a load of different methods. I guess the aim is to make them generic so they can easily be used in different tests.

e2e_tests/tests/per_provider/normal_tests/export_key.rs
ionut-arm: > Because the key is just a public key 

That's the part that I don't get - `generate_rsa_sign_key` creates an RSA key pair with `export` set to true. In `export_public_key` you actually call `export_public_key` so it makes sense that the result is just the public part, would've thought that `export` should give you the full key material.

e2e_tests/tests/per_provider/normal_tests/export_key.rs
sbailey-arm: Ah, it says in the method comment what it is for (same as `import_rsa_public_key`) but I agree it should be added to the method name 🙂.

e2e_tests/tests/per_provider/normal_tests/export_key.rs
sbailey-arm: Ah, I see what you mean now. I've updated the test to include doing the same thing with an RSAPrivateKey and that passes... So maybe the deserialisation method can cope with it? Or the test isn't really valid...?

ionut-arm: I think the deserialization will (almost) always work, but won't necessarily give you a good result. You'll probably need to compare it with something known or re-import it and make sure it works? Open to suggestions...

e2e_tests/tests/per_provider/normal_tests/export_key.rs
```diff
@@ -0,0 +1,212 @@
...
+        return Ok(());
+    }
+
+    let key_name = String::from("check_public_rsa_export_format");
+    client.generate_rsa_sign_key(key_name.clone())?;
+    let key = client.export_key(key_name)?;
+
+    // That should not fail if the bytes are in the expected format.
+    let _public_key: RSAPublicKey = picky_asn1_der::from_bytes(&key).unwrap();
+    let _private_key: RSAPrivateKey = picky_asn1_der::from_bytes(&key).unwrap();
```
ionut-arm, line 96: I don't think the implementation of [`RSAPrivateKey`](https://docs.rs/picky-asn1-x509/0.1.0/src/picky_asn1_x509/private_key_info.rs.html#141) is particularly safe - it doesn't fail on `from_bytes`, it will probably bomb out if you try to call `primes()` if the data you fed into it isn't valid. I've just tried it - if you feed in a public key and attempt to get the primes you get a panic.

We should probs raise a PR with them to change that to a proper implementation.

hug-dev:
New versions of `psa-crypto` and `parsec-interface` have been published for you to use!

hug-dev:
And `parsec-client` as well, for the tests

hug-dev:
I was going to pick your brain but it seems like our `picky` friends published a breaking change as a [minor release](https://github.com/Devolutions/picky-rs/commit/f2c78419841bb385c87361dbcd614e3a41b099ea#diff-dfe6d0c88c21bc288dbbfd21a0d6665c) which we are automatically picking, hence the missing `algorithm_identifier` module in the PKCS11 provider. No need to pick on them though!! It's not a pickle 🥒

PR #205, opened 2020-07-15
Added user and group checks. Auto create socket dir.

sbailey-arm:
Parsec must be run as user `parsec` and `parsec` must be member of group `parsec-client`.
This is disabled if feature `testing` is specified.

Also auto created `/tmp/parsec` if it does not already exist with ownership set to user `parsec` and group `parsec_clients`. User `parsec` must be a member of `parsec_clients` in order for it to be able to change group ownership to it

### DIFF
--- a/src/front/domain_socket.rs
+++ b/src/front/domain_socket.rs
@@ -8,6 +8,8 @@ use super::listener;
 use listener::Connection;
 use listener::Listen;
 use log::error;
+#[cfg(not(feature = "no-parsec-user-and-clients-group"))]
+use std::ffi::CString;
 use std::fs;
 use std::fs::Permissions;
 use std::io::{Error, ErrorKind, Result};
@@ -18,6 +20,10 @@ use std::path::Path;
 use std::time::Duration;
 
 static SOCKET_PATH: &str = "/tmp/parsec/parsec.sock";
+#[cfg(not(feature = "no-parsec-user-and-clients-group"))]
+const PARSEC_USERNAME: &str = "parsec";
+#[cfg(not(feature = "no-parsec-user-and-clients-group"))]
+const PARSEC_GROUPNAME: &str = "parsec-clients";
 
 /// Unix Domain Socket IPC manager
 ///
@@ -32,13 +38,11 @@ pub struct DomainSocketListener {
 
 impl DomainSocketListener {
     /// Initialise the connection to the Unix socket.
-    ///
-    /// # Panics
-    /// - if a file/socket exists at the path specified for the socket and `remove_file`
-    /// fails
-    /// - if binding to the socket path fails
     pub fn new(timeout: Duration) -> Result<Self> {
-        // If this Parsec instance was socket activated (see the `parsec.socket`
+        #[cfg(not(feature = "no-parsec-user-and-clients-group"))]
+        DomainSocketListener::check_user_details()?;
+
+        // is Parsec instance was socket activated (see the `parsec.socket`
         // file), the listener will be opened by systemd and passed to the
         // process.
         // If Parsec was service activated or not started under systemd, this
@@ -46,10 +50,14 @@ impl DomainSocketListener {
         let listener = match sd_notify::listen_fds()? {
             0 => {
                 let socket = Path::new(SOCKET_PATH);
-
-                if socket.exists() {
+                let parent_dir = socket.parent().unwrap();
+                if !parent_dir.exists() {
+                    fs::create_dir_all(parent_dir)?;
+                } else if socket.exists() {
                     fs::remove_file(&socket)?;
                 }
+                #[cfg(not(feature = "no-parsec-user-and-clients-group"))]
+                DomainSocketListener::set_socket_dir_permissions(parent_dir)?;
 
                 let listener = UnixListener::bind(SOCKET_PATH)?;
                 listener.set_nonblocking(true)?;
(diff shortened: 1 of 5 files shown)


sbailey-arm:
I've tested this on my machine with a `parsec` user. If run as a user other than `parsec`, it complains about running as incorrect user. If run as `parsec` but `parsec_clients` group does not exist, throws error about not finding `parsec_clients`. If run as `parsec` and `parsec_clients` exists but `parsec` is not a member, throws error about `parsec` not being a member of `parsec_clients`. If run as `parsec` and `parsec` is a member of `parsec_clients`, it will create `/tmp/parsec` if required.

If I add my local user to `parsec_clients` and run Parsec as `parsec`, my local user can successfully run `cargo test normal_tests`. If my local user is not a member of `parsec_clients`, it fails.

sbailey-arm:
(This isn't going to pass our stock CI as it won't be using the `testing` feature or the `parsec` user with `parsec_clients` group)

ionut-arm:
> 
> 
> I've tested this on my machine with a `parsec` user. If run as a user other than `parsec`, it complains about running as incorrect user. If run as `parsec` but `parsec_clients` group does not exist, throws error about not finding `parsec_clients`. If run as `parsec` and `parsec_clients` exists but `parsec` is not a member, throws error about `parsec` not being a member of `parsec_clients`. If run as `parsec` and `parsec` is a member of `parsec_clients`, it will create `/tmp/parsec` if required.
> 
> If I add my local user to `parsec_clients` and run Parsec as `parsec`, my local user can successfully run `cargo test normal_tests`. If my local user is not a member of `parsec_clients`, it fails.

I really like having the ability to do this! It'll make life much easier for admins... but, at the same time, I'd also like to give them an option - at the moment the only way to not have this run is by using the `testing` feature and that's clearly undesirable.

I was thinking you could add a new section in the config file, something like 
```toml
[fs-permissions]
service-user = "parsec"
clients-group = "parsec-clients"
```
If the section exists, the feature is enabled, and it will be enabled in our example config file. We could add these fields (and a boolean flag tracking if it's enabled) in the `GlobalConfig`, so that other components can use it - `Listener`, maybe even `OnDiskManager`.

Thoughts?

sbailey-arm:
> If the section exists, the feature is enabled, and it will be enabled in our example config file. We could add these fields (and a boolean flag tracking if it's enabled) in the `GlobalConfig`, so that other components can use it - `Listener`, maybe even `OnDiskManager`.

That sounds like a good idea. It makes it easier for admins to enable or disable this feature, and they can also specify their own Parsec user and group.

hug-dev:
I agree with having a separate `fs-permissions` feature for this but not sure with letting the possibility of changing the name of owner and group. These should ideally be hardcoded so that clients can check it as well.

ionut-arm:
> 
> 
> I agree with having a separate `fs-permissions` feature for this but not sure with letting the possibility of changing the name of owner and group. These should ideally be hardcoded so that clients can check it as well.

Makes sense. I really wish we could make this synchronisation problem disappear somehow - both around the paths and the usernames/groups

sbailey-arm:
> 
> 
> I agree with having a separate `fs-permissions` feature for this but not sure with letting the possibility of changing the name of owner and group. These should ideally be hardcoded so that clients can check it as well.

So would this mean we have a `[fs-permissions]` section with a single bool config like `strict-user-and-groups` without the `service-user` or `clients-group`, allowing the admin to turn on or off the security feature without changing names?

hug-dev:
@ionut-arm If this feature is meant to be activated by default, do you think we could have it at build time? Like the one in the Rust client?

ionut-arm:
> 
> 
> @ionut-arm If this feature is meant to be activated by default, do you think we could have it at build time? Like the one in the Rust client?

Yeah, I think if we don't want to make any of the values configurable, just compiling it in (or not) makes sense

sbailey-arm:
Isn't this what's already there? The compile time option `testing` will disable the check, which is enabled by default. (perhaps the name could be changed to `parsec-user-and-clients group`)

ionut-arm:
> 
> 
> Isn't this what's already there? The compile time option `testing` will disable the check, which is enabled by default. (perhaps the name could be changed to `parsec-user-and-clients group`)

Yep, that makes sense, just renaming it should be fine

sbailey-arm:
I've changed the feature to `parsec-user-and-clients-group` which is enabled as a default feature, instead of using the `testing` feature which would disable the extra security checks.

hug-dev:
I guess on the CI we do not have to active the feature. But we should at some point add tests for a secure deployment where all checks are in place.

src/front/domain_socket.rs
```diff
@@ -32,25 +38,35 @@ pub struct DomainSocketListener {
...
+                    }
+                } else {
+                    // Invalid path - socket should not be created in root
+                    // `socket.path` returns None if there is no parent directory (i.e. is in root)
+                    error!("Socket path {} is invalid.", SOCKET_PATH);
+                    return Err(Error::new(
+                        ErrorKind::InvalidInput,
+                        "Socket path is invalid",
+                    ));
                 }
```
ionut-arm, line 69: Is this part actually needed, since the path is declared statically at the top?

sbailey-arm: I don't think so. It was added because `socket.parent()` _can_ fail, but should never in our case. Does it make sense to remove the `else`? It should never be hit, but then if you look at the code without the else, it looks like it would continue as normal, even if it failed. Maybe converting the `Option` that `parent()` returns to a `Result` and returning the error (if there is one - which there never should be!) instead, to make the code more compact?

ionut-arm: You can simply `unwrap` the result of `socket.parent()` - even if we ever change it to something where that fails, the service won't start so the CI will let us know.

sbailey-arm:
> 
> 
> I guess on the CI we do not have to active the feature. But we should at some point add tests for a secure deployment where all checks are in place.

I've modified the CI script to add `--no-default-features` which will remove `parsec-user-and-clients-group`. The ability to disable a single feature would be nice, but [I'm not sure it'll happen any time soon](https://github.com/rust-lang/cargo/issues/3126)...!

hug-dev:
For the problem you described, it might be better to do something similar as in the client: have a `no-fs-permission-check` feature, disabled by default. You can check in the code if the feature is not there. That way it would be easier to selectively remove the checks 👍

hug-dev:
There seems to have been two Travis Builds triggered by this PR. Weird.

src/front/domain_socket.rs
```diff
@@ -84,6 +92,95 @@ impl DomainSocketListener {
...
+            error!(
+                "Incorrect user. Parsec should be run as user {}.",
+                PARSEC_USERNAME
+            );
+            return Err(Error::new(
+                ErrorKind::PermissionDenied,
+                "Parsec run as incorrect user",
+            ));
+        }
+        // Check Parsec client group exists and parsec user is a member of it
```
hug-dev, line 148: Does `parsec` need to be part of `parsec-clients`? Since `parsec` owns the socket dir, can't they use it how they want?

hug-dev:
With this PR, the `parsec` and `parsec-clients` user/group will be mandatory to even run Parsec. This will have to be documented clearly in our book.
Also I am thinking when we will have our first authenticators. The `750` permissions will then have to be `755`, no longer needing the `parsec-clients` group. We would probably also prevent direct/none authentication.
By then this code will have to be modified to comply with that. Should we then stop at checking that the current user is `parsec` and the the permissions are `750`?

src/front/domain_socket.rs
sbailey-arm: A user has to be a member of a group in order to change permissions of the group to that group. So if `parsec` is not a member of `parsec-clients`, it cannot change the group permission of the directory to `parsec-clients`.

sbailey-arm:
> 
> 
> With this PR, the `parsec` and `parsec-clients` user/group will be mandatory to even run Parsec. This will have to be documented clearly in our book.
> Also I am thinking when we will have our first authenticators. The `750` permissions will then have to be `755`, no longer needing the `parsec-clients` group. We would probably also prevent direct/none authentication.
> By then this code will have to be modified to comply with that. Should we then stop at checking that the current user is `parsec` and the the permissions are `750`?

Should we just get it into that state now instead of coming up with a stopgap that will definitely be changed at somepoint?

src/front/domain_socket.rs
hug-dev: I see, that makes sense, thanks.

ionut-arm:
> 
> 
> With this PR, the `parsec` and `parsec-clients` user/group will be mandatory to even run Parsec. This will have to be documented clearly in our book.
> Also I am thinking when we will have our first authenticators. The `750` permissions will then have to be `755`, no longer needing the `parsec-clients` group. We would probably also prevent direct/none authentication.
> By then this code will have to be modified to comply with that. Should we then stop at checking that the current user is `parsec` and the the permissions are `750`?

I think some deployments might want to use this setup, even if we don't really recommend it. But maybe we should force them not to anyway - to at least use the peer credential authentication. That might be the "least secure" option we offer, converting the `direct` authentication.

`none` authentication is still needed for service discovery operations.

> Should we just get it into that state now instead of coming up with a stopgap that will definitely be changed at somepoint?

Not really, because right now we _need_ the `parsec-clients` group to do the "authentication" for us.

hug-dev:
> Should we just get it into that state now instead of coming up with a stopgap that will definitely be changed at somepoint?

My view is that yes we should. Only check that the current user is `parsec` and set the socket dir to be owned by `parsec` and `0o755`. Deployments without authentication would still have to do the whole socket directory creation themselves: I don't think they can do it once after Parsec is running.

hug-dev:
Oops sorry for closing 😅 

I agree with your points and I think this is fine to go with this now. My main concern is that there will be some changes needed in this code if we remove the `parsec-clients` requirement when the peer credential authentication is used.

Issue #199, opened 2020-07-08
Create a mechanism for the listener to pass system-level data to the authenticator


paulhowardarm:
### Summary
Create a mechanism where the front-end listener can pass implementation-specific metadata alongside the request so that the authenticator can use it as an additional source of validation.

### Rationale
At the moment, the authenticator only has one source of input: namely the authentication header bytes that are read from the request. But there are situations where other pieces of data can be obtained from the environment to help with the authentication and validation process. One example of this is peer metadata when unix domain socket is used as the transport. Unix domain sockets allow any incoming connection to be queried for peer metadata, which is a triplet value combining the user ID (UID), group ID (GID) and process ID (PID) of the peer process that is connecting. This information is sourced from the OS, and hence is trusted insofar as the OS is trusted. If this kind of information was made available to the authenticator, then it would allow for the caller's identity to be verified purely by the OS, without the need to resort to an external identity provider service. This mechanism would only be valuable in certain use cases. For example, if the transport is not based on domain sockets, then it clearly doesn't work. It clearly also doesn't work if the UID/GID/PID triplet does not adequately differentiate between client workloads. However, there could be an important class of deployment where these constraints and conventions are followed. In these deployments, an ability to use the peer metadata would represent a significant simplification relative to the complexity of an identity provider.

### Details
TBD - requires some design. There will have to be an abstraction in place for the metadata, and a suitable means of passing it from the **Listener** to the **FrontEndHandler** and subsequently the **Authenticator**.  This could be done as either a "push" (where the listener populates the data in advance), or a "pull" (where the listener gets the data on request). One thing to make sure of is that we need to treat the metadata itself as opaque. It happens to be a triplet of UID/GID/PID values in the case of domain sockets, but we can't rule out other packets of metadata that might be quite different. Several different designs are possible, but the important thing is that a specific listener such as the **DomainSocketListener** would be able to populate the metadata by making the appropriate method to get the peer credentials. (One wrinkle in the story is that the standard Rust UnixListener doesn't seem to have a method to get the peer cred, so we may also have to investigate other UDS support crates).

It will be important to maintain the design principles of Parsec, one of which is that it does not prescribe any specific transport for the wire protocol. Unix domain socket is just one possible transport. So it will be important to design this with suitable abstractions to maintain this loose coupling. Some types of transport might have no concept of metadata whatsoever, or might support types of metadata wholly different from the UID/GID/PID triple.

### Definition of Done
[...]

PR #207, opened 2020-07-23
Create `Connection` abstraction for client communication

joechrisellis:
This commit is in response to issue #199. Here, we introduce a
`Connection` struct which currently contains two things:

* `stream` -- an object representing the communication stream between
  client and service.
* `metadata` -- an enum instance that captures metadata about the
  connection.

This abstraction allows us to carry more information forwards toward the
frontend/authenticator/... .

Specifically, this abstraction was created with UNIX domain sockets in
mind (but the usefulness is not limited here). UNIX domain sockets allow
incoming connections to be queried for peer metadata, which is a triple
(uid, gid, pid) of the peer process that is connecting. Under certain
configurations, this can be used for authentication.

This commit places us in a position of being able to use said metadata
for authentication if needed.

### DIFF
--- a/src/authenticators/direct_authenticator/mod.rs
+++ b/src/authenticators/direct_authenticator/mod.rs
@@ -10,6 +10,7 @@
 
 use super::ApplicationName;
 use super::Authenticate;
+use crate::front::listener::ConnectionMetadata;
 use log::error;
 use parsec_interface::requests::request::RequestAuth;
 use parsec_interface::requests::{ResponseStatus, Result};
@@ -20,7 +21,11 @@ use std::str;
 pub struct DirectAuthenticator;
 
 impl Authenticate for DirectAuthenticator {
-    fn authenticate(&self, auth: &RequestAuth) -> Result<ApplicationName> {
+    fn authenticate(
+        &self,
+        auth: &RequestAuth,
+        _: Option<ConnectionMetadata>,
+    ) -> Result<ApplicationName> {
         if auth.buffer.expose_secret().is_empty() {
             error!("The direct authenticator does not expect empty authentication values.");
             Err(ResponseStatus::AuthenticationError)
@@ -49,9 +54,10 @@ mod test {
 
         let app_name = "app_name".to_string();
         let req_auth = RequestAuth::new(app_name.clone().into_bytes());
+        let conn_metadata = None;
 
         let auth_name = authenticator
-            .authenticate(&req_auth)
+            .authenticate(&req_auth, conn_metadata)
             .expect("Failed to authenticate");
 
         assert_eq!(auth_name.get_name(), app_name);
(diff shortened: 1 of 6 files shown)


src/front/listener.rs
```diff
@@ -25,6 +26,29 @@ pub struct ListenerConfig {
...
+    NoMetadata,
+}
+
+/// Represents a connection to a single client. Contains a stream, used for communication with the
+/// client, and some metadata associated with the connection that might be useful elsewhere (i.e.
+/// authentication, etc).
+pub struct Connection {
+    pub stream: Box<dyn ReadWrite + Send>,
+    pub metadata: ConnectionMetadata,
+}
```
ionut-arm, line 41: I think it's more idiomatic instead of having an enum with a variant `EnumType::NoSomething`, to actually use `Option<EnumType>`. You can probably just create an "empty" `UnixSocket` variant that we flesh out when we actually need that metadata.

src/front/listener.rs
```diff
@@ -25,6 +26,29 @@ pub struct ListenerConfig {
...
+
+impl fmt::Debug for Connection {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // NOTE: we omit the field `stream` because it doesn't implement `Debug`.
+        f.debug_struct("Connection")
+            .field("metadata", &self.metadata)
+            .finish()
+    }
+}
```
ionut-arm, line 50: This is perfectly correct, but you can do it more easily with `#[derivative(...)]` which allows you to ignore some fields from `Debug` - you can look at `BackEndHandler` as an example.

src/front/domain_socket.rs
```diff
@@ -105,7 +106,12 @@ impl Listen for DomainSocketListener {
                     format_error!("Failed to set stream as blocking", err);
                     None
                 } else {
-                    Some(Box::from(stream))
+                    Some(Connection {
+                        stream: Box::new(stream),
+                        // TODO: when possible, we want to replace this with the (uid, gid, pid)
+                        // triple for peer credentials.
+                        metadata: ConnectionMetadata::NoMetadata,
```
ionut-arm, line 113: It's not a conversation for here, just a random note, but I'd say that any IPC can/should have associated metadata - e.g. if we use TCP sockets, we'll have the peer IP address and port which could be used in the authentication.

src/front/listener.rs
joechrisellis: I completely agree. Will change this. :)

src/front/listener.rs
joechrisellis: TIL! Will change.

src/front/domain_socket.rs
joechrisellis: ACK 🙂

src/authenticators/direct_authenticator/mod.rs
```diff
@@ -49,9 +55,10 @@ mod test {
 
         let app_name = "app_name".to_string();
         let req_auth = RequestAuth::new(app_name.clone().into_bytes());
+        let conn_metadata = ConnectionMetadata::NoMetadata;
```
ionut-arm, line 58: You'll need to remove these and just pass `None` 😄

joechrisellis: Doh -- fixed. :)

PR #208, opened 2020-07-27
Add support for `psa_generate_random` operation for MbedCrypto provider

joechrisellis:
This patch adds support for the `psa_generate_random` operation for the MbedCrypto provider. Once again, I am expecting this to fail CI because it's dependent on the other patches. 😄


hug-dev:
Thanks for adding this! It will need new versions on `psa-crypto` and `parsec-interface`. Since more operations are going to be implemented there in the next few days, let's wait to version those repos to not have too many versioning to do :)

joechrisellis:
> Thanks for adding this! It will need new versions on `psa-crypto` and `parsec-interface`. Since more operations are going to be implemented there in the next few days, let's wait to version those repos to not have too many versioning to do :)

Sounds good to me! 💯

src/authenticators/peer_credentials_authenticator/mod.rs
```diff
@@ -0,0 +1,45 @@
+// Copyright 2019 Contributors to the Parsec project.
```
hug-dev, line 1: Should we commit that file in this PR? Are you trying to sneakily make it in 👀 ?

hug-dev (commented):
All look good to me except for the peer credential authenticator 😄

src/authenticators/peer_credentials_authenticator/mod.rs
joechrisellis: Haha, oops. Removed. 😄

hug-dev (approved):
🥂

src/providers/mod.rs
```diff
@@ -265,4 +265,14 @@ pub trait Provide {
         trace!("psa_raw_key_agreement ingress");
         Err(ResponseStatus::PsaErrorNotSupported)
     }
+
+    /// Execute a GenerateRandom operation.
+    fn psa_generate_random(
+        &self,
+        _app_name: ApplicationName,
```
ionut-arm, line 272: I'd argue against requesting an `app_name`. The operation can then be performed even for non-authenticated users, I think that should be alright!

src/providers/mbed_provider/generate_random.rs
```diff
@@ -0,0 +1,27 @@
...
+        let buffer_size = op.size;
+        let mut buffer = vec![0u8; buffer_size];
+
+        match generate_random(&mut buffer) {
+            Ok(_) => Ok(psa_generate_random::Result {
+                random_bytes: buffer.into(),
+            }),
+            Err(error) => {
+                let error = ResponseStatus::from(error);
+                format_error!("Generate random status: ", error);
```
ionut-arm, line 22:
```suggestion
                format_error!("Generate random status", error);
```

The `format_error!` macro adds the `: ` for you, but maybe I should change that since everyone seems tempted to use it as the normal `error!`

src/providers/mbed_provider/generate_random.rs
```diff
@@ -0,0 +1,27 @@
...
+use parsec_interface::operations::psa_generate_random;
+use parsec_interface::requests::{ResponseStatus, Result};
+use psa_crypto::operations::other::generate_random;
+
+impl MbedProvider {
+    pub(super) fn psa_generate_random_internal(
+        &self,
+        _app_name: ApplicationName,
+        op: psa_generate_random::Operation,
+    ) -> Result<psa_generate_random::Result> {
```
ionut-arm, line 12: Now that I look at it, you could just put this one straight in the `psa_generate_random` method, there isn't that much code to warrant an `_internal` one, but feel free to keep it here

ionut-arm (commented):
Thanks! 💯

src/providers/mbed_provider/generate_random.rs
joechrisellis: I think I'll leave this -- it looks like most of the operations have an `*_internal` method. If you have strong opinions here feel free to reply and I'll change. 😄

ionut-arm (approved):
Looks good!

PR #213, opened 2020-08-04
Rebase on new tss_esapi

puiterwijk:

### DIFF
--- a/src/providers/tpm_provider/utils.rs
+++ b/src/providers/tpm_provider/utils.rs
@@ -10,9 +10,10 @@ use picky_asn1_x509::RSAPublicKey;
 use serde::{Deserialize, Serialize};
 use std::convert::TryInto;
 use tss_esapi::abstraction::transient::KeyParams;
-use tss_esapi::response_code::{Error, Tss2ResponseCodeKind};
-use tss_esapi::utils::algorithm_specifiers::{EllipticCurve, HashingAlgorithm};
+use tss_esapi::constants::algorithm::{EllipticCurve, HashingAlgorithm};
+use tss_esapi::constants::response_code::Tss2ResponseCodeKind;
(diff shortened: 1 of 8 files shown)


src/providers/tpm_provider/utils.rs
```diff
@@ -99,11 +100,18 @@ pub fn parsec_to_tpm_params(attributes: Attributes) -> Result<KeyParams> {
...
+                    size,
+                    pub_exponent: 0,
+                })
+            } else {
+                Ok(KeyParams::RsaSign {
+                    size,
+                    scheme: convert_asym_scheme_to_tpm(attributes.policy.permitted_algorithms)?,
+                    pub_exponent: 0,
+                })
+            }
```
ionut-arm, line 114: I'm wondering if this should be an `else if attributes.is_hash_signable() || attributes.is_hash_verifiable()` with the fallback case something like `Err(ResponseStatus::PsaErrorNotSupported)`.

src/providers/tpm_provider/utils.rs
```diff
@@ -99,11 +100,18 @@ pub fn parsec_to_tpm_params(attributes: Attributes) -> Result<KeyParams> {
                 x @ 1024 | x @ 2048 | x @ 3072 | x @ 4096 => x.try_into().unwrap(), // will not fail on the matched values
                 _ => return Err(ResponseStatus::PsaErrorInvalidArgument),
             };
-            Ok(KeyParams::Rsa {
-                size,
-                scheme: convert_asym_scheme_to_tpm(attributes.policy.permitted_algorithms)?,
-                pub_exponent: 0,
-            })
+            if attributes.is_encrypt_permitted() {
```
ionut-arm, line 103: You should probably include `|| attributes.is_decrypt_permitted()`

PR #214, opened 2020-08-06
Add Unix peer credentials authenticator

joechrisellis:
This authenticator uses Unix peer credentials for authentication. Unix
peer credentials provide direct access to the (effective) uid/gid on the
other end of a domain socket connect, without cooperation between the
endpoints. This means that we can determine the uid/gid of the
connecting process, and therefore infer the username of the user that
owns said process.

This authenticator:

- grabs the (uid, gid) pair of the connecting process.
- determine the username of the owner of the connecting process based on
  the uid.
- creates an `ApplicationName` based on the username.

Note that this patch depends on the following PR being merged:

    https://github.com/rust-lang/rust/pull/75148

At the time of writing, this PR is currently under review and is not
merged into the Rust stdlib. This patch therefore will not build with
the current a stable/nightly compiler.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,9 @@ picky-asn1-x509 = { version = "0.3.2", optional = true }
 users = "0.10.0"
 libc = "0.2.77"
 
+[dev-dependencies]
+rand = { version = "0.7.3", features = ["small_rng"] }
+
 [package.metadata.docs.rs]
 features = ["pkcs11-provider", "tpm-provider", "tss-esapi/docs", "mbed-crypto-provider"]
 
--- a/src/authenticators/mod.rs
+++ b/src/authenticators/mod.rs
@@ -8,10 +8,10 @@
 //! used throughout the service for identifying the request initiator. The input to an authentication
 //! is the `RequestAuth` field of a request, which is parsed by the authenticator specified in the header.
 //! The authentication functionality is abstracted through an `Authenticate` trait.
-//!
-//! Currently only a simple Direct Authenticator component is implemented.
 
 pub mod direct_authenticator;
+
+pub mod unix_peer_credentials_authenticator;
 
 use crate::front::listener::ConnectionMetadata;
 use parsec_interface::operations::list_authenticators;
@@ -35,7 +35,7 @@ pub trait Authenticate {
     /// Authenticates a `RequestAuth` payload and returns the `ApplicationName` if successful. A
     /// optional `ConnectionMetadata` object is passed in too, since it is sometimes possible to
     /// perform authentication based on the connection's metadata (i.e. as is the case for UNIX
-    /// domain sockets with peer credentials).
+    /// domain sockets with Unix peer credentials).
     ///
     /// # Errors
     ///
--- a/src/authenticators/unix_peer_credentials_authenticator/mod.rs
+++ b/src/authenticators/unix_peer_credentials_authenticator/mod.rs
@@ -1,0 +1,186 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+//! Unix peer credentials authenticator
+//!
+//! The `UnixPeerCredentialsAuthenticator` uses Unix peer credentials to perform authentication. As
+//! such, it uses the effective Unix user ID (UID) to authenticate the connecting process. Unix
+//! peer credentials also allow us to access the effective Unix group ID (GID) of the connecting
+//! process, although this information is currently unused.
+//!
+//! Currently, the stringified UID is used as the application name.
+
+use super::ApplicationName;
+use super::Authenticate;
+use crate::front::listener::ConnectionMetadata;
+use log::error;
+use parsec_interface::operations::list_authenticators;
+use parsec_interface::requests::request::RequestAuth;
+use parsec_interface::requests::AuthType;
+use parsec_interface::requests::{ResponseStatus, Result};
+use parsec_interface::secrecy::ExposeSecret;
+use std::convert::TryInto;
+
+/// Unix peer credentials authenticator.
+#[derive(Copy, Clone, Debug)]
+pub struct UnixPeerCredentialsAuthenticator;
+
+impl Authenticate for UnixPeerCredentialsAuthenticator {
+    fn describe(&self) -> Result<list_authenticators::AuthenticatorInfo> {
+        Ok(list_authenticators::AuthenticatorInfo {
+            description: String::from(
+                "Uses Unix peer credentials to authenticate the client. Verifies that the self-declared \
+                Unix user identifier (UID) in the request's authentication header matches that which is \
+                found from the peer credentials."
+            ),
+            version_maj: 0,
+            version_min: 1,
+            version_rev: 0,
+            id: AuthType::PeerCredentials,
+        })
+    }
+
+    fn authenticate(
+        &self,
+        auth: &RequestAuth,
+        meta: Option<ConnectionMetadata>,
+    ) -> Result<ApplicationName> {
+        // Parse authentication request.
+        let expected_uid_bytes = auth.buffer.expose_secret();
+
+        const EXPECTED_UID_SIZE_BYTES: usize = 4;
+        let expected_uid: [u8; EXPECTED_UID_SIZE_BYTES] =
+            expected_uid_bytes.as_slice().try_into().map_err(|_| {
+                error!(
+                    "UID in authentication request is not the right size (expected: {}, got: {}).",
+                    EXPECTED_UID_SIZE_BYTES,
+                    expected_uid_bytes.len()
+                );
+                ResponseStatus::AuthenticationError
+            })?;
+        let expected_uid = u32::from_le_bytes(expected_uid);
+
+        let meta = meta.ok_or_else(|| {
+            error!("Authenticator did not receive any metadata; cannot perform authentication.");
+            ResponseStatus::AuthenticationError
+        })?;
+
+        #[allow(unreachable_patterns)]
+        let (uid, _gid, _pid) = match meta {
+            ConnectionMetadata::UnixPeerCredentials { uid, gid, pid } => (uid, gid, pid),
+            _ => {
+                error!("Wrong metadata type given to Unix peer credentials authenticator.");
+                return Err(ResponseStatus::AuthenticationError);
+            }
+        };
+
+        // Authentication is successful if the _actual_ UID from the Unix peer credentials equals
+        // the self-declared UID in the authentication request.
+        if uid == expected_uid {
+            Ok(ApplicationName(uid.to_string()))
+        } else {
+            error!("Declared UID in authentication request does not match the process's UID.");
+            Err(ResponseStatus::AuthenticationError)
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::super::Authenticate;
+    use super::UnixPeerCredentialsAuthenticator;
+    use crate::front::listener::ConnectionMetadata;
+    use parsec_interface::requests::request::RequestAuth;
+    use parsec_interface::requests::ResponseStatus;
+    use rand::Rng;
+    use std::os::unix::net::UnixStream;
+    use users::get_current_uid;
+
+    #[test]
+    fn successful_authentication() {
+        // This test should PASS; we are verifying that our username gets set as the application
+        // secret when using Unix peer credentials authentication with Unix domain sockets.
+
+        // Create two connected sockets.
+        let (sock_a, _sock_b) = UnixStream::pair().unwrap();
+        let (cred_a, _cred_b) = (sock_a.peer_cred().unwrap(), _sock_b.peer_cred().unwrap());
+
+        let authenticator = UnixPeerCredentialsAuthenticator {};
+
+        let req_auth_data = cred_a.uid.to_le_bytes().to_vec();
+        let req_auth = RequestAuth::new(req_auth_data);
+        let conn_metadata = Some(ConnectionMetadata::UnixPeerCredentials {
+            uid: cred_a.uid,
+            gid: cred_a.gid,
+            pid: None,
+        });
+
+        let auth_name = authenticator
+            .authenticate(&req_auth, conn_metadata)
+            .expect("Failed to authenticate");
+
+        assert_eq!(auth_name.get_name(), get_current_uid().to_string());
+    }
+
+    #[test]
+    fn unsuccessful_authentication_wrong_declared_uid() {
+        // This test should FAIL; we are trying to authenticate, but we are declaring the wrong
+        // UID.
+
+        // Create two connected sockets.
+        let (sock_a, _sock_b) = UnixStream::pair().unwrap();
+        let (cred_a, _cred_b) = (sock_a.peer_cred().unwrap(), _sock_b.peer_cred().unwrap());
+
+        let authenticator = UnixPeerCredentialsAuthenticator {};
+
+        let wrong_uid = cred_a.uid + 1;
+        let wrong_req_auth_data = wrong_uid.to_le_bytes().to_vec();
+        let req_auth = RequestAuth::new(wrong_req_auth_data);
+        let conn_metadata = Some(ConnectionMetadata::UnixPeerCredentials {
+            uid: cred_a.uid,
+            gid: cred_a.gid,
+            pid: cred_a.pid,
+        });
+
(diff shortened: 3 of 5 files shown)


joechrisellis:
Note: CI failure is expected since this patch depends on a feature that is not yet in the stdlib.

src/authenticators/uds_authenticator/mod.rs
```diff
@@ -0,0 +1,129 @@
+// Copyright 2019 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+//! Unix domain socket (UDS) peer credentials authenticator
+//!
+//! The `UDSAuthenticator` uses peer credentials from Unix domain sockets to perform
+//! authentication. As such, it uses the UID/GID to grab a username for the owner of the connecting
+//! process. This is used as the application name.
```
paulhowardarm, line 7: Do we gain anything by converting to a user name as the application name, rather than just using the stringified UID itself as the application ID string? This seems to have additional dependencies and operations that could conceivably fail. I'm also slightly fearful about converting to usernames given that the environment may be containerised, which could have implications in terms of how the UIDs convert to names in the transition between containers.

src/authenticators/mod.rs
```diff
@@ -12,6 +12,7 @@
 //! Currently only a simple Direct Authenticator component is implemented.
 
 pub mod direct_authenticator;
+pub mod uds_authenticator;
```
paulhowardarm, line 15: Is it useful to perhaps wrap this authenticator in a Rust/Cargo feature so that it can be selected for at build time? (Actually, looking at the rest of the change, perhaps you've already done that? But there's an indication that it's only temporary. I actually think there is value in having this be a selectable thing even longer term, because it's useful for people to be able to compile down a Parsec executable with only the precise pieces that their deployment needs. Providers and authenticators are the kinds of things that we might want to snip out, because they can be heavyweight, especially in terms of the dependencies that get dragged in with them).

src/authenticators/mod.rs
```diff
@@ -12,6 +12,7 @@
 //! Currently only a simple Direct Authenticator component is implemented.
```
paulhowardarm, line 12: This comment is outdated by the change.

src/authenticators/uds_authenticator/mod.rs
```diff
@@ -0,0 +1,129 @@
+// Copyright 2019 Contributors to the Parsec project.
```
paulhowardarm, line 1: I guess this should be 2020 now.

src/authenticators/uds_authenticator/mod.rs
hug-dev: > Do we gain anything by converting to a user name as the application name, rather than just using the stringified UID itself as the application ID string?

I think the only thing we gain is for it to look nicer in the `mappings` folder and in logs like `"New request received from application name \"{}\""`. So mostly for esthetics/debugging. Also, if we add any operation which takes an application name as parameter, it might be easier to specify the user name rather than the UID.

I guess I supposed that every UID must have an username mapping to it but if that is not true in some environments, I agree using the plain UID would be better.

joechrisellis: That's a good point.

I can't think of any reason why the stringified UID wouldn't work just fine. The only comment that I have is that the username is perhaps more 'human-friendly', but I don't think these strings will ever necessarily be exposed anyways.

Happy to change this. 🙂

src/authenticators/mod.rs
joechrisellis: Yes, good idea. Will do this. 👍

src/authenticators/mod.rs
joechrisellis: Ah, well spotted, thanks!

src/authenticators/uds_authenticator/mod.rs
joechrisellis: Fixed. :)

src/authenticators/uds_authenticator/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+            }
+        };
+
+        let (uid, _gid) = match meta {
+            ConnectionMetadata::PeerCredentials{uid, gid} => (uid, gid),
+            // TODO: add wildcard pattern when `ConnectionMetadata` has more possibilities.
+        };
+
+        Ok(ApplicationName(uid.to_string()))
+    }
```
ionut-arm, line 39: I thought the authentication was meant to verify that the `RequestAuth` contains a string with that UID, just as a "sanity" check

joechrisellis: Ah, I wasn't aware. Will fix that. 👍

paulhowardarm: Agreed, I think it should. I didn't follow through every line here and I thought there was some Rusty magic going on. My opinion is that this system should be fully explicit with a well-defined authentication type and expected header contents. The calling process should "declare" its UID in the auth header, and the service should use the peer credential to check that it matches.

joechrisellis: Yeah, it's definitely a fair point. This is now implemented. 🙂

src/authenticators/mod.rs
```diff
@@ -8,11 +8,12 @@
 //! used throughout the service for identifying the request initiator. The input to an authentication
 //! is the `RequestAuth` field of a request, which is parsed by the authenticator specified in the header.
 //! The authentication functionality is abstracted through an `Authenticate` trait.
-//!
-//! Currently only a simple Direct Authenticator component is implemented.
 
 pub mod direct_authenticator;
 
+#[cfg(feature = "uds-authenticator")]
+pub mod uds_authenticator;
```
paulhowardarm, line 15: I should have noticed this before, but I'm not sure why this is called UDS authenticator? Assuming UDS stands for Unix Domain Socket, that's the name of the transport rather than the method of authentication (which is UID). You could have all sorts of different authenticators on a UDS transport, so calling this a UDS authenticator doesn't quite characterise it accurately.

ionut-arm: Does `PeerCredentialAuthenticator` make more sense? I can imagine we'd be able to get peer credentials for other IPC mechanisms as well and we could apply the same logic for authentication there

joechrisellis: My bad; I made the wrong assumption that using a Unix domain socket as a transport meant that we would always want to use peer credentials for authentication. I agree with @ionut-arm, `PeerCredentialsAuthenticator` seems like a good name here. 🙂

Cargo.toml
```diff
@@ -70,3 +70,4 @@ pkcs11-provider = ["pkcs11", "picky-asn1-der", "picky-asn1", "picky-asn1-x509",
 tpm-provider = ["tss-esapi", "picky-asn1-der", "picky-asn1", "picky-asn1-x509"]
 all-providers = ["tpm-provider", "pkcs11-provider", "mbed-crypto-provider"]
 docs = ["pkcs11-provider", "tpm-provider", "tss-esapi/docs", "mbed-crypto-provider"]
+peer-credentials-authenticator = []
```
hug-dev, line 73: See [my related comment](https://github.com/parallaxsecond/parsec-book/pull/57#discussion_r485089884), if it is agreed, I would personally prefer all occurences of "peer credentials" to be renamed "Unix peer credentials".

src/front/domain_socket.rs
```diff
@@ -202,11 +203,10 @@ impl Listen for DomainSocketListener {
                     format_error!("Failed to set stream as blocking", err);
                     None
                 } else {
+                    let metadata = stream.metadata();
```
hug-dev, line 206: Is the `GetMetadata` trait needed here if we are already in Unix specific code? As in, could you just replace the implementation of that trait on `UnixStream` here?

src/front/domain_socket.rs
```diff
@@ -248,3 +248,22 @@ impl DomainSocketListenerBuilder {
         })?)
     }
 }
+
+impl GetMetadata for UnixStream {
+    #[cfg(feature = "peer-credentials-authenticator")]
+    fn metadata(&self) -> Option<ConnectionMetadata> {
+        let ucred = self
+            .peer_cred()
+            .expect("Failed to get peer credentials for Unix domain socket connection.");
```
hug-dev, line 257: I would prefer returning an error instead of panicking. In our case, you can log something and return `None`!

hug-dev (commented):
Thanks for adding it! Probably better to wait for the specs on the book and the PR in Rust to be merged before merging this one but thought of some early comments!

src/authenticators/unix_peer_credentials_authenticator/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+                "UID in authentication request is not the right size (expected: {}, got: {}).",
+                EXPECTED_UID_SIZE_BYTES,
+                expected_uid_bytes.len()
+            );
+            return Err(ResponseStatus::AuthenticationError);
+        }
+
+        let boxed_slice = expected_uid_bytes.into_boxed_slice();
+        let boxed_array: Box<[u8; 4]> = boxed_slice.try_into().unwrap();
+        let expected_uid = u32::from_le_bytes(*boxed_array);
```
hug-dev, line 65: I think you can directly use the `TryInto` from the slice to the owned type and use the failed case as a way to check that the auth buffer was not 4 bytes (so even checks if not empty 😃 ):
```rust
let expected_uid = auth.buffer.expose_secret();
let expected_uid: [u8; 4] = expected_uid_bytes.try_into().ok_or_else(|| {
    error!(
        "UID in authentication request is not the right size (expected: {}, got: {}).",
        EXPECTED_UID_SIZE_BYTES,
        expected_uid_bytes.len()
    );
    return Err(ResponseStatus::AuthenticationError);
})?;
let expected_uid = u32::from_le_bytes(expected_uid);
```

joechrisellis: Ooh, that's much nicer. 😄

src/authenticators/unix_peer_credentials_authenticator/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+    use std::os::unix::net::UnixStream;
+    use users::get_current_uid;
+
+    #[test]
+    fn successful_authentication() {
+        // This test should PASS; we are verifying that our username gets set as the application
+        // secret when using Unix peer credentials authentication with Unix domain sockets.
+
+        // Create two connected sockets.
+        let (sock_a, _sock_b) = UnixStream::pair().unwrap();
```
hug-dev, line 104: Pretty good idea to use that for unit tests!

src/front/domain_socket.rs
hug-dev: Just pinging this question as I think it is still valid

src/front/domain_socket.rs
```diff
@@ -248,3 +248,144 @@ impl DomainSocketListenerBuilder {
...
+            gid: ucred.gid,
+        })
+    }
+}
+
+// == IMPORTANT NOTE ==
+//
+// The code below has been cherry-picked from the following PR:
+//
+//     https://github.com/rust-lang/rust/pull/75148
```
hug-dev, line 274: Could you also please add the stabilizing issue https://github.com/rust-lang/rust/issues/42839?

hug-dev (commented):
Good idea of having that code directly here 👌

src/authenticators/unix_peer_credentials_authenticator/mod.rs
```diff
@@ -0,0 +1,191 @@
...
+        &self,
+        auth: &RequestAuth,
+        meta: Option<ConnectionMetadata>,
+    ) -> Result<ApplicationName> {
+        // Parse authentication request.
+        let expected_uid_bytes = auth.buffer.expose_secret();
+        if expected_uid_bytes.is_empty() {
+            error!("Expected UID in authentication request, but it is empty.");
+            return Err(ResponseStatus::AuthenticationError);
+        }
```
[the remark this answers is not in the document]
joechrisellis: > Small nit but I think that the empty case will also be checked in the following try_into (as the buffer will not be 4 bytes).

Nice catch, done.

> I would also say that you can use the hardcoded 4 in the following code instead of EXPECTED_UID_SIZE_BYTES as this size os probably not something meant to change often.

I don't know -- I kind of like this. It gives some obvious context to the magic number; when the user reads it, it is clear that the UID is supposed to be 4 bytes long. If that variable didn't exist it would be as clear, I don't think. 🙂

src/front/domain_socket.rs
joechrisellis: We could -- this was more just thinking for the future. Will revert this, apologies for the lag on responding to this comment. 😄

src/front/domain_socket.rs
joechrisellis: Done.

hug-dev (approved):
Thanks for applying the changes, it looks pretty good to me 👌

src/authenticators/unix_peer_credentials_authenticator/mod.rs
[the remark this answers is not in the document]
hug-dev: > I don't know -- I kind of like this. It gives some obvious context to the magic number; when the user reads it, it is clear that the UID is supposed to be 4 bytes long. If that variable didn't exist it would be as clear, I don't think. 🙂

Yes, it's true that this is more readable!

src/front/listener.rs
```diff
@@ -34,7 +34,13 @@ pub struct ListenerConfig {
...
 #[derive(Copy, Clone, Debug)]
 pub enum ConnectionMetadata {
-    // TODO: nothing here right now. Metadata types will be added as needed.
+    /// Unix peer credentials metadata for Unix domain sockets.
+    UnixPeerCredentials {
+        /// The effective UID of the connecting process.
+        uid: u32,
+        /// The effective GID of the connecting process.
+        gid: u32,
+    },
```
ionut-arm, line 43: Would it make sense to add the `pid` here as well? Don't know what it would be used for (if at all), but then again, we don't use the `gid` either...

src/authenticators/unix_peer_credentials_authenticator/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+        let conn_metadata = None;
+        let auth_result = authenticator.authenticate(&req_auth, conn_metadata);
+        assert_eq!(auth_result, Err(ResponseStatus::AuthenticationError));
+    }
+
+    #[test]
+    fn unsuccessful_authentication_wrong_metadata() {
+        // TODO: this test needs implementing when we have more than one metadata type. At the
+        // moment, the compiler just complains with an 'unreachable branch' message.
+    }
```
ionut-arm, line 187: I'd say an issue in github or even better a non-documentation comment on `ConnectionMetadata` are more likely to catch a developer's attention if some other variant is added to the `enum`, otherwise this will probs just be forgotten

src/authenticators/unix_peer_credentials_authenticator/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+        let expected_uid = u32::from_le_bytes(expected_uid);
+
+        let meta = meta.ok_or_else(|| {
+            error!("Authenticator did not receive any metadata; cannot perform authentication.");
+            ResponseStatus::AuthenticationError
+        })?;
+
+        let (uid, _gid) = match meta {
+            ConnectionMetadata::UnixPeerCredentials { uid, gid } => (uid, gid),
+            // TODO: add wildcard pattern when `ConnectionMetadata` has more possibilities.
```
ionut-arm, line 69: Can you add the wildcard pattern and "disable" the clippy warning? Or is it something the compiler doesn't like?

src/authenticators/unix_peer_credentials_authenticator/mod.rs
joechrisellis: Added a comment in the `ConnectionMetadata` struct that asks the user to grep for `TODO(new_metadata_variant)` in the tests and update accordingly when a new variant is added.

src/front/listener.rs
joechrisellis: True, good for completeness -- added.

src/authenticators/unix_peer_credentials_authenticator/mod.rs
joechrisellis: We can -- done.

ionut-arm (approved):
Thank you! :100:

PR #215, opened 2020-08-06
CoreProvider can query the other providers

ionut-arm:
This commit adds functionality to the CoreProvider that allows it to
call operations on the other providers in the system.
The change is required to support future functionality around
"composite-operations" - that is, operations which produce changes or
require information from all the providers.

### DIFF
--- a/src/back/backend_handler.rs
+++ b/src/back/backend_handler.rs
@@ -16,6 +16,7 @@ use parsec_interface::requests::{
 };
 use parsec_interface::requests::{BodyType, ProviderID};
 use std::io::{Error, ErrorKind};
+use std::sync::Arc;
 
 /// Back end handler component
 ///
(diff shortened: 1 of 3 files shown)


hug-dev:
> Would the DeleteClient and ListClients operation also be on the Provide trait and the CoreProvider would call them on the different providers?

With that there could be the nice semantic of:
* you call those operations on one crypto provider, you get the result only for this provider
* you call them on the `CoreProvider`, you get the results for all crypto providers

ionut-arm:
> 
> 
> > Would the DeleteClient and ListClients operation also be on the Provide trait and the CoreProvider would call them on the different providers?
> 
> With that there could be the nice semantic of:
> 
>     * you call those operations on one crypto provider, you get the result only for this provider
> 
>     * you call them on the `CoreProvider`, you get the results for all crypto providers

Yeah, I'm not sure, tbh... That's an option and I think for these operations it makes sense, maybe for others it won't. Every operation like this would need a method on `Provide`, but maybe for some we'll want to check somewhere along the line if it's allowed to go to non-core providers or not.

PR #218, opened 2020-08-13
Add Docker image for testing Parsec as a systemd service

joechrisellis:


hug-dev:
I think it's worth a try to test the installation using directly the virtual environment given by GitHub Actions (and maybe Travis CI for Arm?) instead of using Docker there.

hug-dev:
Even if `systemd` testing seems to be non-optimal on CI, I think the principle of the installation tests should remain and will be very useful to check that we are doing authentication and multi-tenancy right.

I am thinking of multiple tests, not sure if they would be best on their own workflow or put in common:
1. the service should fail to run (under any authentication) if the socket folder has wrong permissions
2. Under direct authentication using the `parsec-clients` group, a client not in that group should not be able to connect
3. Under peer credential authentication (when ready):
    1. even a client not in `parsec-clients` should be able to connect
    2. one client can not access another client's keys

hug-dev:
Closing for now, as the requirements changed for testing this. See the related issue.

PR #219, opened 2020-08-13
Add check to prevent the Parsec service from running as root

joechrisellis:
This commit introduces a guard at the top of `main` to ensure that
Parsec is not being run as root. Disallowing the Parsec service from
running as root reduces attack surface in line with the principle of
least privilege.

This behaviour can be overridden by setting the `PARSEC_ALLOW_ROOT`
environment variable.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -2,6 +2,11 @@
 
 # (Required) Core settings apply to the service as a whole rather than to individual components within it.
 [core_settings]
+# Whether or not to allow the service to run as the root user. If this is false, the service will refuse to
+# start if it is run as root. If this is true, the safety check is disabled and the service will be allowed to
+# start even if it is being run as root. The recommended (and default) setting is FALSE; allowing Parsec to
+# run as root violates the principle of least privilege.
+#allow_root = false
 # Size of the thread pool used for processing requests. Defaults to the number of processors on
 # the machine.
 #thread_pool_size = 8
--- a/src/bin/main.rs
+++ b/src/bin/main.rs
@@ -41,6 +41,7 @@ use std::sync::{
 };
 use std::time::Duration;
 use structopt::StructOpt;
+use users::get_current_uid;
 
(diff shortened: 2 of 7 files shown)


src/bin/main.rs
```diff
@@ -61,6 +63,16 @@ struct Opts {
 const MAIN_LOOP_DEFAULT_SLEEP: u64 = 10;
 
 fn main() -> Result<()> {
+    // Guard against running as root. This check can be overridden by setting the PARSEC_ALLOW_ROOT
+    // environment variable.
+    if get_current_uid() == 0 && env::var("PARSEC_ALLOW_ROOT").is_err() {
```
ionut-arm, line 68: This can probably be added to the config file instead of an env var, and do the check after the file is parsed

config.toml
```diff
@@ -2,6 +2,11 @@
 
 # (Required) Core settings apply to the service as a whole rather than to individual components within it.
 [core_settings]
+# Whether or not to allow the service to run as the root user. If this is false, the service will refuse to
+# start if it is run as root. If this is true, the safety check is disabled and the service will be allowed to
+# start even if it is being run as root. The recommended setting is FALSE; allowing Parsec to run as root
```
ionut-arm, line 7: Probably add a "default" in there - `the recommended and default setting is FALSE`

src/bin/main.rs
```diff
@@ -84,6 +85,17 @@ fn main() -> Result<()> {
         )
     })?;
 
+    // Guard against running as root. This check can be overridden by changing `allow_root` inside
+    // the config file.
+    let allow_root = config.core_settings.allow_root.unwrap_or(false);
+    if get_current_uid() == 0 && !allow_root {
```
ionut-arm, line 91: You can swap these two so the `get_current_uid` isn't triggered if `allow_root` is true

PR #220, opened 2020-08-14
Add implementation for ListAuthenticators operation

joechrisellis:


src/providers/core_provider/mod.rs
```diff
@@ -32,8 +40,11 @@ pub struct CoreProvider {
     wire_protocol_version_maj: u8,
     provider_info: Vec<ProviderInfo>,
     provider_opcodes: HashMap<ProviderID, HashSet<Opcode>>,
+    authenticator_info: Vec<AuthenticatorInfo>,
     #[derivative(Debug = "ignore")]
     prov_list: Vec<Arc<dyn Provide + Send + Sync>>,
+    #[derivative(Debug = "ignore")]
+    auth_list: Vec<Arc<dyn Authenticate + Send + Sync>>,
```
ionut-arm, line 47: I don't think you need this here, you only need the `authenticator_info` above - that makes the process simpler, as you don't have to make ownership of authenticators shared between the `CoreProvider` and the `FrontEndHandler`.

The `provider_info` example is a bit skewed now because we actually need an interface to the other providers for the upcoming operations. Previously it only took in a vector of `ProviderInfo`

joechrisellis: Do you think there might be a case for having this in the future? I agree that it would be much simpler without it, but at the same time, I was thinking that there might be a situation in the future where the core provider might need access to the authenticators.

If you can't see a future where this is needed then I'll gladly simplify! 👍

ionut-arm: It probably will at some point (e.g. for dynamically setting the public key of an Identity Provider or stuff like that), but I don't know if it's good practice to keep it in here for now. I'll defer this to @hug-dev to express his thoughts on this too

hug-dev: I would say it is better to remove it for now if it is not needed. I agree that it might be used in the future but I think adding it in the future might not be that hard.

Looking at the different fields of the `CoreProvider` structures, could we simplify though? For example can the `provider_info` and `provider_opcodes` data be directly computed from `prov_list`? Maybe it is similar for `authenticator_info` and `auth_list`. This is not to change in this PR anyway but something to remember.

src/utils/service_builder.rs
```diff
@@ -108,24 +110,31 @@ impl ServiceBuilder {
             return Err(Error::new(ErrorKind::InvalidData, "need one provider"));
         }
 
-        let backend_handlers = build_backend_handlers(providers)?;
+        // The authenticators supported by the Parsec service.
+        let mut authenticators: Vec<(AuthType, Authenticator)> = Vec::new();
+        authenticators.push((AuthType::Direct, Arc::from(DirectAuthenticator {})));
```
hug-dev, line 115: I guess it does not matter now because we currently only have one authenticator, but is everything in place here to keep the list ordered if another authenticator is added?

hug-dev (commented):
Agree with the code change, but just need clarification over the order of the list returned.

src/providers/core_provider/mod.rs
joechrisellis: Okay -- I've changed this now.

Note that this will still fail CI because of versioning woes. I will wait until @sbailey-arm's operations are merged (which includes a version bump) then I will rebase on top of that. 🙂

src/utils/service_builder.rs
joechrisellis: Yes -- the order here is important. The order in which the elements are added here is the order in which they will be returned to any client requesting them. Do you want me to add a comment here to make this clearer? Or perhaps we could extract this out into its own function?

src/providers/core_provider/mod.rs
hug-dev: What new versioning do you need? I believe that the client was published with new versions of everything including ListAuthenticators

hug-dev: Oh I see what you mean, ignore my comment 😅

src/utils/service_builder.rs
hug-dev: I think a comment would be good, to note for the future that the order here is important!

joechrisellis: Done!

PR #227, opened 2020-08-18
Update attribute handling in PKCS11 provider

ionut-arm:
This commit updates the way we handle parameters for the PKCS11
provider, making the process of adding support for new key types more
straight forward.

### DIFF
--- a/e2e_tests/Cargo.toml
+++ b/e2e_tests/Cargo.toml
@@ -28,3 +28,9 @@ uuid = "0.7.4"
 rsa = "0.3.0"
 picky-asn1-x509 = "0.1.0"
 base64 = "0.12.3"
+
+[features]
+mbed-crypto-provider = []
+tpm-provider = []
+pkcs11-provider = []
+all-providers = ["pkcs11-provider","tpm-provider","mbed-crypto-provider"]
--- a/src/providers/pkcs11_provider/asym_encryption.rs
+++ b/src/providers/pkcs11_provider/asym_encryption.rs
@@ -6,17 +6,17 @@ use crate::authenticators::ApplicationName;
 use crate::key_info_managers::KeyTriple;
 use log::{info, trace};
 use parsec_interface::operations::psa_algorithm::Algorithm;
-use parsec_interface::operations::{psa_sign_hash, psa_verify_hash};
+use parsec_interface::operations::{psa_asymmetric_decrypt, psa_asymmetric_encrypt};
 use parsec_interface::requests::{ProviderID, Result};
 use std::convert::TryFrom;
 use utils::CkMechanism;
 
 impl Pkcs11Provider {
-    pub(super) fn psa_sign_hash_internal(
+    pub(super) fn psa_asymmetric_encrypt_internal(
         &self,
         app_name: ApplicationName,
-        op: psa_sign_hash::Operation,
-    ) -> Result<psa_sign_hash::Result> {
+        op: psa_asymmetric_encrypt::Operation,
+    ) -> Result<psa_asymmetric_encrypt::Result> {
         let key_triple = KeyTriple::new(app_name, ProviderID::Pkcs11, op.key_name.clone());
         let store_handle = self.key_info_store.read().expect("Key store lock poisoned");
         let (key_id, key_attributes) = get_key_info(&key_triple, &*store_handle)?;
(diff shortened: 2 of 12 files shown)


ionut-arm:
This PR is only intended to make the job of expanding our support easier for when we add asymmetric encryption (for example). I'm sure there are opportunities for improving the structure of what I wrote, this is just a draft.

src/providers/pkcs11_provider/utils.rs
```diff
@@ -1,14 +1,173 @@
...
+                let params: CK_RSA_PKCS_OAEP_PARAMS_PTR = &mut params;
+                CK_MECHANISM {
+                    mechanism: mech.mech_type().into(),
+                    pParameter: params as CK_VOID_PTR,
+                    ulParameterLen: len,
+                }
+            }
+        }
+    }
+}
```
ionut-arm, line 170: This works? More precisely, keeping a reference to those `params` structures, but dropping the object afterwards

src/providers/pkcs11_provider/utils.rs
```diff
@@ -258,3 +417,134 @@ impl Drop for Session<'_> {
...
+    Vec<CK_ATTRIBUTE>,
+    CK_MECHANISM_TYPE,
+)> {
+    match attributes.key_type {
+        Type::RsaKeyPair => {
+            let mechanism = CK_MECHANISM {
+                mechanism: CKM_RSA_PKCS_KEY_PAIR_GEN,
+                pParameter: std::ptr::null_mut(),
+                ulParameterLen: 0,
+            };
```
ionut-arm, line 443: I wonder if it's worth expanding this to support for ECDSA to have parity with the TPM provider, but figuring out which attributes go where might be a pain.

src/providers/pkcs11_provider/utils.rs
hug-dev: Hmm yes you are right that is really bad and unsafe ☠️ By the time the `params` are dereferenced, the address is no longer valid.
I think the abstraction `CkMechanism` is wrong here: it should really contain a reference on the parameters (with a lifetime specifier) and probably not be `Copy`.
The problem also is that `CK_MECHANISM` or `CK_ATTRIBUTE` do not contain a lifetime as they use raw pointers so I think it's hard to use the compiler to make sure the references are valid... In that case, even if you convert `CkMechanism` to a `CK_MECHANISM`, you lose the reference lifetime and the compiler can not do checks on it when accessing it

ionut-arm: Yeah, this is a massive issue, since you'd never be sure whether the reference is valid. I'm actually kinda shocked Rust lets you do this (the conversion to pointer part) without an `unsafe` around it... Maybe we should enable [this](https://rust-lang.github.io/rust-clippy/master/index.html#as_conversions) clippy lint? Though that might be overkill.

src/providers/pkcs11_provider/utils.rs
```diff
@@ -1,14 +1,274 @@
...
+                        pParameter: p_params as CK_VOID_PTR,
+                        ulParameterLen: len,
+                    },
+                    Some(CParams::CkmRsaPkcsPssParams(params)),
+                )
+            }
+            CkMechanism::CkmRsaPkcsOaep(ref mut params) => {
+                let mut params = params.as_c_type();
+                let len = ::std::mem::size_of_val(&params);
+                //ERROR: that does not work because the address of params might change when it is
```
hug-dev, line 253: See here.

I thought that the address of Rust object would not change when they are moved which is why I return from this function the C `*_PARAMS` structure. But the address does change 😢

e2e_tests/provider_cfg/pkcs11/config.toml
```diff
@@ -23,4 +24,4 @@ key_info_manager = "on-disk-manager"
 library_path = "/usr/local/lib/softhsm/libsofthsm2.so"
 user_pin = "123456"
 # The slot_number mandatory field is going to replace the following line with a valid number
-# slot_number
+slot_number = 1013939064
```
ionut-arm, line 27: *cough cough*

src/providers/pkcs11_provider/utils.rs
ionut-arm: ok, this was a fun one... 

I modified your code and now it works, will push in the morning. The way I did it was with `Pin`. Instead of holding the C `*_PARAMS` structure directly in `CParams`, I changed it to store `Pin<Box<*_PARAMS>>`. That way, the address is kept intact.

HOWEVER... my changes still didn't work initially and I spent a lot of time trying to figure out what was wrong... only to find that [SoftHSMv2 only supports SHA1 for OAEP](https://github.com/opendnssec/SoftHSMv2/issues/474) -_- Will have to decide if we should create some new tests, or if we want to implement support for the other algos in SoftHSM.

ionut-arm: (I have to say, though, this is rather convoluted... would be nice if we didn't have to do this in our code 😉)

src/providers/pkcs11_provider/utils.rs
```diff
@@ -1,14 +1,275 @@
...
+            CkMechanismType::CkmRsaPkcsOaep => CKM_RSA_PKCS_OAEP,
+            CkMechanismType::CkmSha1 => CKM_SHA_1,
+            CkMechanismType::CkmSha256 => CKM_SHA256,
+            CkMechanismType::CkmSha384 => CKM_SHA384,
+            CkMechanismType::CkmSha512 => CKM_SHA512,
+        }
+    }
+}
+
+pub fn mech_type_to_allowed_mech_attribute(mech_type: &mut CK_MECHANISM_TYPE) -> CK_ATTRIBUTE {
```
hug-dev, line 46: Can the `CK_MECHANISM` referenced here move after this function call? I would say not because of the mutable referenced that is passed.

src/providers/pkcs11_provider/utils.rs
hug-dev: oh wow excellent idea! I finally understand what `Pin` is made for 👍

hug-dev (commented):
THanks for making it work!! I agree with all the changes 👌

e2e_tests/Cargo.toml
```diff
@@ -28,3 +28,9 @@ uuid = "0.7.4"
 rsa = "0.3.0"
 picky-asn1-x509 = "0.1.0"
 base64 = "0.12.3"
+
+[features]
+mbed-crypto-provider = []
+tpm-provider = []
+pkcs11-provider = []
+all-providers = ["pkcs11-provider","tpm-provider","mbed-crypto-provider"]
```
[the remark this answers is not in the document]
ionut-arm: Is that even possible? To change the feature list from `build.rs`

src/providers/pkcs11_provider/utils.rs
ionut-arm: I should remove that comment 😅

src/providers/pkcs11_provider/asym_encryption.rs
```diff
@@ -0,0 +1,113 @@
...
+        op: psa_asymmetric_encrypt::Operation,
+    ) -> Result<psa_asymmetric_encrypt::Result> {
+        let key_triple = KeyTriple::new(app_name, ProviderID::Pkcs11, op.key_name.clone());
+        let store_handle = self.key_info_store.read().expect("Key store lock poisoned");
+        let (key_id, key_attributes) = get_key_info(&key_triple, &*store_handle)?;
+
+        op.validate(key_attributes)?;
+
+        let mut mech = CkMechanism::try_from(Algorithm::from(op.alg))?;
+        let (mech, _params) = mech.as_c_type();
```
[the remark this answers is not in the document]
ionut-arm: Is it somewhere in The Book or in the reference manual whether these unused vars are dropped randomly or at the end of their scope, like all the other ones? Does it change with whether the build is for release or not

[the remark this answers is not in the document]
hug-dev: If those counts as [temporaries](https://doc.rust-lang.org/stable/reference/expressions.html#temporaries) then I think they would be dropped at the end of the enclosing statement.

e2e_tests/Cargo.toml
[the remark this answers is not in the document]
hug-dev: Yes, you can emit `cfg` flag options with [`rustc-cfg`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-cfgkeyvalue)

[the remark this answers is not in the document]
hug-dev: but we don't have to change that now, we can create an issue for it

src/providers/pkcs11_provider/asym_encryption.rs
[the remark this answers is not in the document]
ionut-arm: No, I think it's covered by [this](https://doc.rust-lang.org/stable/reference/statements.html#let-statements) but that's kinda the same. I think temporaries are when you create a value using some call, but don't assign it through a `let` (or other ways of assigning a name to the value)

hug-dev (approved):
Looks good to me!
Let's ship it 🕵️

PR #233, opened 2020-08-26
Add `buffer_size_limit` config option for providers

joechrisellis:
The buffer size limit can be used to cap the maximum allowed buffer size
sent in a response. Requests that ask for buffers larger than this value
will be automatically rejected.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -31,6 +31,10 @@
 # such as key names or policies
 #log_error_details = false
 
+# Decide how large (in bytes) buffers inside responses from this provider can be. Requests that ask
+# for buffers larger than this threshold will be rejected. Defaults to 1MB.
+#buffer_size_limit = 1048576
+
 # (Required) Configuration for the service IPC listener component.
 [listener]
 # (Required) Type of IPC that the service will support.
--- a/src/utils/service_builder.rs
+++ b/src/utils/service_builder.rs
@@ -53,6 +53,9 @@ const WIRE_PROTOCOL_VERSION_MAJOR: u8 = 1;
 /// Default value for the limit on the request body size (in bytes) - equal to 1MB
 const DEFAULT_BODY_LEN_LIMIT: usize = 1 << 19;
 
+/// Default value for the limit on the buffer size for response (in bytes) - equal to 1MB
+pub const DEFAULT_BUFFER_SIZE_LIMIT: usize = 1 << 20;
+
 type KeyInfoManager = Arc<RwLock<dyn ManageKeyInfo + Send + Sync>>;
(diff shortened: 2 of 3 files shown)


src/utils/service_builder.rs
```diff
@@ -53,6 +53,9 @@ const WIRE_PROTOCOL_VERSION_MAJOR: u8 = 1;
 /// Default value for the limit on the request body size (in bytes) - equal to 1MB
 const DEFAULT_BODY_LEN_LIMIT: usize = 1 << 19;
```
hug-dev, line 54: We should probably fix this one as well 👀

config.toml
```diff
@@ -64,10 +64,15 @@ provider_type = "MbedCrypto"
 # (Required) Name of key info manager that will support this provider.
 key_info_manager = "on-disk-manager"
 
+# (Optional) Decide how large (in bytes) buffers inside responses from this provider can be.
+# Requests that ask for buffers larger than this threshold will be rejected. Defaults to 1MB.
+#buffer_size_limit = 1048576
```
hug-dev, line 69: I think for this option it is fine if we only have one global config option declared in the `core_settings` and that work for all the service.
That way, you just have to add the option to the [`GlobalConfig`](https://docs.rs/parsec-service/0.3.0/parsec_service/utils/struct.GlobalConfig.html) and access it from there directly when you need it 💯

hug-dev (commented):
Thanks! That will be useful for generate_random. Added an alternative that might simplify the change. See what you think.

src/utils/service_builder.rs
joechrisellis: Yeah, was going to do that in a separate patch. Let me do that now...

joechrisellis: [Speedy PR](https://github.com/parallaxsecond/parsec/compare/master...joechrisellis:body_len_limit_fix?expand=1)!

config.toml
joechrisellis: I thought it might be a good idea to have a separate one for each provider in case the provider itself has specific limitations that we might want to accommodate for. Although thinking about it, I can't really come up with anything concrete off the top of my head. I'll change this.

hug-dev: I guess as all providers are running with the same allocated memory, that's fine to assume that they would all have similar limitations! Also, did not want to increase the configuration complexity for such a small feature 😋

joechrisellis: SGTM!

hug-dev (approved):
Looks good!

ionut-arm (approved):

PR #238, opened 2020-08-28
Allow software operations in PKCS11 provider

ionut-arm:
This commit modifies the PKCS11 provider so that it allows public key
operations to be performed in software, conditioned on a configuration
flag. If said flag is set, the public part of the key is exported from
PKCS11 and imported into PSA Crypto, and the operation is performed
using an implementation of PSA Crypto.

Changes have also been made to the way key metadata is handled, moving
some of the helper functions onto the `Pkcs11Provider` as methods. A key
matadata cache was also added, to speed up the operations requiring
access to key information.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,7 @@ features = ["docs"]
 default = []
 no-parsec-user-and-clients-group = []
 mbed-crypto-provider = ["psa-crypto"]
-pkcs11-provider = ["pkcs11", "picky-asn1-der", "picky-asn1", "picky-asn1-x509"]
+pkcs11-provider = ["pkcs11", "picky-asn1-der", "picky-asn1", "picky-asn1-x509", "psa-crypto"]
 tpm-provider = ["tss-esapi", "picky-asn1-der", "picky-asn1", "picky-asn1-x509"]
 all-providers = ["tpm-provider", "pkcs11-provider", "mbed-crypto-provider"]
 # The Mbed provider is not included in the docs because of 2 reasons:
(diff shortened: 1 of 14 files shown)


hug-dev:
Thanks for that change 👍 
I am going to be annoying but... could you split this PR into smaller parts 😬 ? Like one for the software fallback and one for the key info manager methods refactoring? If it's too annoying or not practical I don't mind but it would make the review easier 😋

ionut-arm:
> 
> 
> Thanks for that change 👍
> I am going to be annoying but... could you split this PR into smaller parts 😬 ? Like one for the software fallback and one for the key info manager methods refactoring? If it's too annoying or not practical I don't mind but it would make the review easier 😋

Yep, I'll try and make the key info manager method refactoring first commit and the software fallback the 2nd commit, but deffo not today.

In other news, the CI is failing for some interesting reasons - see [here](https://github.com/ionut-arm/parsec/runs/1042479619?check_suite_focus=true) (search for `CKR_TEMPLATE_INCONSISTENT`). I don't know if it's because of the key info manager thing making operations much faster or if it's because of the refactoring we did earlier this week... It seems the PKCS11 call to generate a key things the template is inconsistent for some random test (it was random and flakey locally as well).

ionut-arm:
I'm not sure why, yet, but generating keys now fails randomly with some obscure error codes (`CKR_GENERAL_ERROR` or `CKR_TEMPLATE_INCONSISTENT`) see [here](https://github.com/ionut-arm/parsec/runs/1056950940?check_suite_focus=true) line 704

PR #240, opened 2020-09-01
Improve key metadata handling

ionut-arm:
This commit improves the handling of key metadata in the PKCS11
provider, allowing the provider to cache metadata in memory instead of
always requesting it from the key info manager.

### DIFF
--- a/src/key_info_managers/mod.rs
+++ b/src/key_info_managers/mod.rs
@@ -77,6 +77,16 @@ impl KeyTriple {
     pub fn belongs_to_provider(&self, provider_id: ProviderID) -> bool {
         self.provider_id == provider_id
     }
+
+    /// Get the key name
+    pub fn key_name(&self) -> &str {
+        &self.key_name
+    }
+
+    /// Get the app name
+    pub fn app_name(&self) -> &ApplicationName {
+        &self.app_name
+    }
 }
 
 /// Converts the error string returned by the ManageKeyInfo methods to
--- a/src/providers/pkcs11_provider/key_metadata.rs
+++ b/src/providers/pkcs11_provider/key_metadata.rs
@@ -1,0 +1,109 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use super::LocalIdStore;
+use super::{KeyInfo, Pkcs11Provider};
+use crate::key_info_managers;
+use crate::key_info_managers::{KeyTriple, ManageKeyInfo};
+use log::{error, warn};
+use parsec_interface::operations::psa_key_attributes::*;
+use parsec_interface::requests::{ResponseStatus, Result};
+use std::sync::RwLock;
+
+impl Pkcs11Provider {
+    // This function returns the `RWLocks` found on the `Pkcs11Provider`
+    // in the order in which they should *always* be taken. Changing the order
+    // of locking in one method can very easily result in deadlocking.
+    pub(super) fn get_ordered_locks<'a>(
+        &'a self,
+    ) -> (
+        &'a RwLock<dyn ManageKeyInfo + Send + Sync>,
+        &'a RwLock<LocalIdStore>,
+    ) {
+        (&self.key_info_store, &self.local_ids)
+    }
+
+    /// Gets a key identifier and key attributes from the Key Info Manager.
+    pub(super) fn get_key_info(&self, key_triple: &KeyTriple) -> Result<([u8; 4], Attributes)> {
+        let locks = self.get_ordered_locks();
+        let store_handle = locks.0.read().expect("Local ID lock poisoned");
+        match store_handle.get(key_triple) {
+            Ok(Some(key_info)) => {
+                if key_info.id.len() == 4 {
+                    let mut dst = [0; 4];
+                    dst.copy_from_slice(&key_info.id);
+                    Ok((dst, key_info.attributes))
+                } else {
+                    error!("Stored Key ID is not valid.");
+                    Err(ResponseStatus::KeyInfoManagerError)
+                }
+            }
+            Ok(None) => Err(ResponseStatus::PsaErrorDoesNotExist),
+            Err(string) => Err(key_info_managers::to_response_status(string)),
+        }
+    }
+
+    pub(super) fn create_key_id(
+        &self,
+        key_triple: KeyTriple,
+        key_attributes: Attributes,
+    ) -> Result<[u8; 4]> {
+        let locks = self.get_ordered_locks();
+        let mut store_handle = locks.0.write().expect("Key store lock poisoned");
+        let mut local_ids_handle = locks.1.write().expect("Local ID lock poisoned");
+        let mut key_id = rand::random::<[u8; 4]>();
+        while local_ids_handle.contains(&key_id) {
+            key_id = rand::random::<[u8; 4]>();
+        }
+        let key_info = KeyInfo {
+            id: key_id.to_vec(),
+            attributes: key_attributes,
+        };
+        match store_handle.insert(key_triple.clone(), key_info) {
+            Ok(insert_option) => {
+                if insert_option.is_some() {
+                    if crate::utils::GlobalConfig::log_error_details() {
+                        warn!("Overwriting Key triple mapping ({})", key_triple);
+                    } else {
+                        warn!("Overwriting Key triple mapping");
+                    }
+                }
+                let _ = local_ids_handle.insert(key_id);
+                Ok(key_id)
+            }
+            Err(string) => Err(key_info_managers::to_response_status(string)),
+        }
+    }
+
+    pub(super) fn remove_key_id(&self, key_triple: &KeyTriple) -> Result<[u8; 4]> {
+        let locks = self.get_ordered_locks();
+        let mut store_handle = locks.0.write().expect("Key store lock poisoned");
+        let mut local_ids_handle = locks.1.write().expect("Local ID lock poisoned");
+        match store_handle.remove(key_triple) {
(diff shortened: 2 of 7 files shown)


src/key_info_managers/mod.rs
```diff
@@ -98,7 +108,7 @@ pub trait ManageKeyInfo {
     /// # Errors
     ///
     /// Returns an error as a String if there was a problem accessing the Key Info Manager.
-    fn get_all(&self, provider_id: ProviderID) -> Result<Vec<&KeyTriple>, String>;
+    fn get_all(&self, provider_id: ProviderID) -> Result<Vec<(KeyTriple, KeyInfo)>, String>;
```
hug-dev, line 111: What is the context of that change?

The reason we used a reference here was to mimick the different methods on the `HashMap` and here particularly [`get`](https://doc.rust-lang.org/beta/std/collections/struct.HashMap.html#method.get) which returns a reference.
I would be happy to add another method `to_mappings_vec` which has this signature or modify this one but using references instead. Like returning a `Vec<(&KeyTriple, Option<&KeyInfo>)>` for example?

src/providers/pkcs11_provider/mod.rs
```diff
@@ -41,6 +43,12 @@ const SUPPORTED_OPCODES: [Opcode; 8] = [
     Opcode::PsaAsymmetricEncrypt,
 ];
 
+#[derive(Debug)]
+struct KeyMetadata {
```
hug-dev, line 47: Could probably be moved in `key_metadata.rs`?

src/providers/pkcs11_provider/key_metadata.rs
```diff
@@ -0,0 +1,115 @@
...
+use crate::key_info_managers::KeyTriple;
+use log::{error, warn};
+use parsec_interface::operations::psa_key_attributes::*;
+use parsec_interface::requests::{ResponseStatus, Result};
+
+impl Pkcs11Provider {
+    /// Gets a key identifier and key attributes from the Key Info Manager.
+    pub(super) fn get_key_info(&self, key_triple: &KeyTriple) -> Result<([u8; 4], Attributes)> {
+        let store_handle = self.key_info_store.read().expect("Local ID lock poisoned");
+        let key_metadata_cache = self
```
hug-dev, line 14: Actually, I have just checked the `OnDiskKeyInfoManager` [implementation](https://github.com/parallaxsecond/parsec/blob/3ea3461da05bad9006572c306ccc28373d685ee9/src/key_info_managers/on_disk_manager/mod.rs#L34) and it already stores a cache of value it uses when getting them/checking if they exist. Which means that the cache might not be needed here 😬

src/providers/pkcs11_provider/key_metadata.rs
```diff
@@ -0,0 +1,115 @@
...
+use crate::key_info_managers;
+use crate::key_info_managers::KeyTriple;
+use log::{error, warn};
+use parsec_interface::operations::psa_key_attributes::*;
+use parsec_interface::requests::{ResponseStatus, Result};
+
+impl Pkcs11Provider {
+    /// Gets a key identifier and key attributes from the Key Info Manager.
+    pub(super) fn get_key_info(&self, key_triple: &KeyTriple) -> Result<([u8; 4], Attributes)> {
+        let store_handle = self.key_info_store.read().expect("Local ID lock poisoned");
```
hug-dev, line 13: This module and these methods, not needing the mutex handle, make the code much clearer and easier to understand, thanks for that.
It means as well that we are reducing the scope of when the mutex is taken/freed which means that more parallelism will be available for the operations (which is good). I also believe that the error you saw on the other PR is a data race bug which is due to that.

src/providers/pkcs11_provider/key_metadata.rs
```diff
@@ -0,0 +1,115 @@
...
+use super::{KeyInfo, KeyMetadata, Pkcs11Provider};
+use crate::key_info_managers;
+use crate::key_info_managers::KeyTriple;
+use log::{error, warn};
+use parsec_interface::operations::psa_key_attributes::*;
+use parsec_interface::requests::{ResponseStatus, Result};
+
+impl Pkcs11Provider {
+    /// Gets a key identifier and key attributes from the Key Info Manager.
+    pub(super) fn get_key_info(&self, key_triple: &KeyTriple) -> Result<([u8; 4], Attributes)> {
```
hug-dev, line 12: Could this method return a `KeyMetadata`?

src/providers/pkcs11_provider/key_metadata.rs
```diff
@@ -0,0 +1,115 @@
...
+                        attributes: key_attributes,
+                    },
+                );
+                Ok(key_id)
+            }
+            Err(string) => Err(key_info_managers::to_response_status(string)),
+        }
+    }
+
+    pub(super) fn remove_key_id(&self, key_triple: &KeyTriple, key_id: [u8; 4]) -> Result<()> {
```
hug-dev, line 88: Could this method only take `key_triple` as a parameter? It would be a bit easier to use that way

hug-dev (commented):
👨‍🎤

src/providers/pkcs11_provider/mod.rs
ionut-arm: Well if there's already a cache in the key info manager, there's no reason to have this structure - I only created it to add the public key ID in the other commit!

src/providers/pkcs11_provider/key_metadata.rs
ionut-arm: Yeah, I'll do that!

src/providers/pkcs11_provider/key_metadata.rs
ionut-arm: Do you think we should have `KeyMetadata` and use that everywhere, or just drop that struct?

src/providers/pkcs11_provider/key_metadata.rs
ionut-arm: Yeah, probably, I'll have to look into it, but I have no clue how. The PKCS11 library should be thread-safe if we use a different session for each call (which we do)

src/providers/pkcs11_provider/key_metadata.rs
ionut-arm: Lol, I didn't even check, I just assumed there was no cache.

src/providers/pkcs11_provider/key_metadata.rs
hug-dev: oh yeah maybe drop it then if it's not needed anymore

src/key_info_managers/mod.rs
ionut-arm: The context was - when initialising the PKCS11 provider with its new cache, we needed both the values and the key information. But if I remove the cache 

> The reason we used a reference here was to mimick the different methods on the HashMap and here particularly get which returns a reference.

That's quite forced, though, because `HashMap` does not have a `get_all` (it's called `keys` or `values`) and does not return a `Vec` of references (it returns an iterator). So the only thing common was the fact that it was returning references. I thought `get_all` as a name makes more sense with returning everything, but yeah, since the cache is gone we don't need it.

src/providers/pkcs11_provider/key_metadata.rs
```diff
@@ -0,0 +1,109 @@
...
+    // in the order in which they should *always* be taken. Changing the order
+    // of locking in one method can very easily result in deadlocking.
+    pub(super) fn get_ordered_locks<'a>(
+        &'a self,
+    ) -> (
+        &'a RwLock<dyn ManageKeyInfo + Send + Sync>,
+        &'a RwLock<LocalIdStore>,
+    ) {
+        (&self.key_info_store, &self.local_ids)
+    }
```
ionut-arm, line 23: I've added this method to help with the lock ordering. Unfortunately I couldn't think of a way to include the session counter mutex into this

src/providers/pkcs11_provider/key_metadata.rs
```diff
@@ -0,0 +1,109 @@
...
+        key_triple: KeyTriple,
+        key_attributes: Attributes,
+    ) -> Result<[u8; 4]> {
+        let locks = self.get_ordered_locks();
+        let mut store_handle = locks.0.write().expect("Key store lock poisoned");
+        let mut local_ids_handle = locks.1.write().expect("Local ID lock poisoned");
+        let mut key_id = rand::random::<[u8; 4]>();
+        while local_ids_handle.contains(&key_id) {
+            key_id = rand::random::<[u8; 4]>();
+        }
```
hug-dev, line 56: I don't know if that is not over-optimising but you could modify the code and create new blocks so that each locks is taken the least amount of time needed.
Like:
```diff
-match store_handle.insert(key_triple.clone(), key_info) {
+match {
+    let mut store_handle = locks.0.write().expect("Key store lock poisoned");
+    store_handle.insert(key_triple.clone(), key_info)
+} {
```
and something similar for the `locks.1`

Maybe that is over the top though and adds more complexity than needed. What do you think?

hug-dev (commented):
Big fan of this PR! A small comment I could think of but otherwise, this is good for me!
Next would be to fix the bugs I guess to make the CI pass 😱

src/providers/pkcs11_provider/key_metadata.rs
ionut-arm: Ok, discussed this but given the issues we've encountered with multithreading and SoftHSM2, we're adding a mutex around all PKCS11 operations, so this wouldn't be that helpful. Plus, it'd probably fall under "premature optimization"

hug-dev (approved):

PR #241, opened 2020-09-04
Allow software operations in PKCS11 provider

### LINKED ISSUE #223: Support dbus-parsec with NXP secureobj library
### Summary
Provide the required functionality to support the dbus-parsec integration on platforms such as NXP Layerscape 1046a. This means enhancing the existing PKCS#11 provider so that it can use non-PKCS#11 functions for public key operations.

### Detail
NXP's secure object library (`libsecureobj.o`) has a PKCS#11 front-end, so the PKCS#11 Parsec provider can already be used to connect with it on supported platforms. However, the PKCS#11 coverage on this platform is limited to signing and decryption operations with private keys. The corresponding public key operations (for encryption and verification) are absent.

Parsec needs a way of fulfilling the entire set of private and public key operations using a mixture of PKCS#11 functionality and MBed Crypto functionality. There currently isn't a Parsec provider that is architected to switch between different methods in this way. All of the existing providers are uniform in the sense that they expect all operations to be fulfilled by a single back-end API in the software stack that connects with the hardware.

PKCS#11 is explicitly an opt-in standard, so it is entirely reasonable for an implementation to only provide partial coverage of the API. Parsec might encounter other PKCS#11 libraries, not just the NXP one, where the coverage is limited in this way. Public keys by definition do not need hardware protection, so it might be a common pattern for public key operations to be absent in the library. It would be good for Parsec to work with implementations like this.

There are various options for doing this:

- Import the `rust-psa-crypto` crate directly into the PKCS#11 provider, and use it to supply the primitives that are not available in the connected PKCS#11 library. This would require making a PKCS#11 `C_GetFunctionList` call in the connected library to see which operations are supported, and then switching over to PSA Crypto calls as needed.
- Fork the PKCS#11 provider to create an NXP-specific version, which imports the `rust-psa-crypto` crate and uses it for the functions that we know are needed. (Basically the same as the option above, but with more hard-coding and without inspecting the `C_GetFunctionList` data - might be simpler to implement?)
- Create a new hybrid provider for the secure object platform, which delegates functionality to the PKCS#11 and MBed Crypto providers respectively. This possibly requires very little code because it can re-use all of the code in the existing providers. It just becomes a veneer or adaptive provider, created specifically for this purpose.

ionut-arm:
This commit adds functionality to the PKCS11 provider to allow it to
handle public key operations in software, using a PSA Crypto-compatible
library. The functionality is enabled via a config flag.

Resolves #223 


e2e_tests/provider_cfg/pkcs11/config.toml
```diff
@@ -22,5 +22,6 @@ provider_type = "Pkcs11"
 key_info_manager = "on-disk-manager"
 library_path = "/usr/local/lib/softhsm/libsofthsm2.so"
 user_pin = "123456"
+software_public_operations = true
```
hug-dev, line 25: Do you think we should keep it that way forever? It would be nice to check that public key operations executed on hardware are working as well if that is the default.
I think it's fine to have the `software_public_operations` only checked manually for now.

src/providers/pkcs11_provider/key_management.rs
```diff
@@ -62,6 +62,53 @@ impl Pkcs11Provider {
...
+        &self,
+        key_triple: &KeyTriple,
+        pub_key_id: Id,
+    ) -> Result<()> {
+        info!("Removing public key stored in PSA.");
+        unsafe { psa_crypto::operations::key_management::destroy(pub_key_id) }.map_err(|e| {
+            if crate::utils::GlobalConfig::log_error_details() {
+                error!(
+                    "Failed to remove public part of key \"{}\" from PSA Crypto.",
+                    key_triple.key_name()
```
hug-dev, line 102: I would personally say that if `key_triple` is passed as a parameter just for logging, we can just skip that and log in all cases: `Failed to remove public key from PSA Crypto.`
But feel free to keep it if you think this info is definitely needed.

hug-dev (commented):
This looks pretty good to me overall!!

e2e_tests/provider_cfg/pkcs11/config.toml
ionut-arm: Yeah, sorry, I'll change that back and add some `config` tests that check the software implementation

src/providers/pkcs11_provider/key_management.rs
ionut-arm: Yeah, I was a bit 😒 changing that just to print the key triple, but could change it back

ionut-arm: Changed!

e2e_tests/provider_cfg/pkcs11/config.toml
ionut-arm: I've added a couple of tests for the software verification. Do you think we should run all tests against both? As in, do something in `ci.sh` instead of just config tests?

ci.sh
```diff
@@ -137,7 +137,7 @@ pgrep -f target/debug/parsec >/dev/null
 if [ "$PROVIDER_NAME" = "all" ]; then
     echo "Execute all-providers tests"
     RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml all_providers
-    RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml config
+    RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml config -- --test-threads=1
```
hug-dev, line 140: Is testing in single-threaded mode needed even with the mutex we have now in PKCS11 provider?

hug-dev (commented):
I think the testing is good at this is now!

ci.sh
ionut-arm: That was a mistake in the first place, not putting it like that. If you think about it, each of those tests changes the configuration of the service and expects it to be in a certain configuration while it's running. Multiple tests at the same time make no sense.

hug-dev: oh right those are the config tests! Agree, then

hug-dev (approved):

PR #243, opened 2020-09-08
Import private key support for TPM provider

joechrisellis:


src/providers/tpm_provider/key_management.rs
```diff
@@ -203,6 +176,77 @@ impl TpmProvider {
...
+        // Validate the public and the private key.
+        validate_public_key(&public_key, &attributes)?;
+        validate_private_key(&private_key, &attributes)?;
+
+        let key_prime = private_key.primes()[0].as_unsigned_bytes_be();
+        let public_modulus = private_key.modulus().as_unsigned_bytes_be();
+
+        let (int_bytes, _) = private_key
+            .public_exponent()
+            .split_at(std::mem::size_of::<u32>());
```
hug-dev, line 219: `split_at` could panic if the public exponent is smaller than 4 bytes but I guess that's fine because of the `validate_` functions above.

src/providers/tpm_provider/utils.rs
```diff
@@ -243,6 +243,72 @@ pub fn parsec_to_tpm_signature(
...
+
+    Ok(())
+}
+
+/// Validates an RSAPrivateKey against the attributes we expect. Returns ok on success, otherwise
+/// returns an error.
+pub fn validate_private_key(private_key: &RSAPrivateKey, attributes: &Attributes) -> Result<()> {
+    // NOTE: potentially incomplete, but any errors that aren't caught here should be caught
+    //       further down the stack (i.e. in the tss crate).
+    let key_prime = &private_key.primes()[0];
```
hug-dev, line 299: Maybe it is better to use the fields of `RSAPrivateKey` directly

src/providers/tpm_provider/key_management.rs
```diff
@@ -114,11 +115,24 @@ impl TpmProvider {
...
     ) -> Result<psa_import_key::Result> {
-        if op.attributes.key_type != Type::RsaPublicKey {
-            error!("The TPM provider currently only supports importing RSA public key.");
-            return Err(ResponseStatus::PsaErrorNotSupported);
+        match op.attributes.key_type {
+            Type::RsaPublicKey => self.psa_import_key_internal_rsa_public(app_name, op),
+            Type::RsaKeyPair => self.psa_import_key_internal_rsa_keypair(app_name, op),
+            _ => {
+                error!(
+                    "The TPM provider does not support the {:?} key type.",
```
hug-dev, line 123: As far as I know, the TPM provider does not support *importing* these other key types but it might support generating keys of those types

hug-dev (commented):
Thanks for this! I like the scaffolding you did, something we need to do/coordinate on all our providers.

I left a few comments and I have a request: could you please add a e2e test importing a RSA key pair? There might already be existing tests on that which are not activated for the TPM provider.

src/providers/tpm_provider/utils.rs
```diff
@@ -243,6 +243,72 @@ pub fn parsec_to_tpm_signature(
...
+    if key_len != attributes.bits / 2 {
+        error!(
+            "The key prime is not of the expected size (expected {}, got {})",
+            attributes.bits / 2,
+            key_len,
+        );
+        return Err(ResponseStatus::PsaErrorInvalidArgument);
+    }
+    Ok(())
+}
```
ionut-arm, line 310: 💯

src/providers/tpm_provider/utils.rs
```diff
@@ -243,6 +243,72 @@ pub fn parsec_to_tpm_signature(
...
+        } else {
+            error!("The TPM Provider only supports 0x101 as public exponent for RSA public keys");
+        }
+        return Err(ResponseStatus::PsaErrorNotSupported);
+    }
+    let key_data = public_key.modulus.as_unsigned_bytes_be();
+    let len = key_data.len();
+
+    let key_bits = attributes.bits;
+    if key_bits != 0 && len * 8 != key_bits {
```
ionut-arm, line 266: I'm wondering - given that the modulus is given as a vector of bytes, the assumed size is always a multiple of 8 (i.e. `len * 8`). But the size of the modulus could theoretically be something in between, like 1245, in which case the MSBits of the last byte won't be set. Now, given that the TPM will most likely reject that anyway (I think only a few values are accepted by the spec), should we reject it from here? The log messages below wouldn't be very helpful in this (rather edge) case

src/providers/tpm_provider/utils.rs
```diff
@@ -243,6 +243,72 @@ pub fn parsec_to_tpm_signature(
...
+        } else {
+            error!("`bits` field of key attributes must be either 0 or equal to the size of the key in `data`.");
+        }
+        return Err(ResponseStatus::PsaErrorInvalidArgument);
+    }
+
+    if len != 128 && len != 256 {
+        if crate::utils::GlobalConfig::log_error_details() {
+            error!(
+                "The TPM provider only supports 1024 and 2048 bits RSA public keys ({} bits given).",
```
ionut-arm, line 282: Is this limitation in the `TransientKeyContext`? I think TPMs support one of `1024, 2048, 3072 or 4096`

ionut-arm (commented):
Thanks for the change! 🚀

tests/providers/tpm.rs
```diff
@@ -284,3 +321,61 @@ fn asym_encrypt_with_crate() {
 
     assert_eq!(&initial_plaintext[..], &plaintext[..]);
 }
+
+#[test]
+fn import_private_key() {
```
hug-dev, line 326: Thanks for adding the test!
@ionut-arm is it fine for this test to be here or should we add it as an e2e test?

hug-dev (approved):
Thanks!
It all look good to me! Just one question about the test, if we want to add it in the `e2e_tests` folder instead so that it benefits other providers as well. Probably fine for now.

tests/providers/tpm.rs
ionut-arm: Leave it there! I'll rebase my changes, though I won't finish today

ionut-arm (approved):

PR #244, opened 2020-09-08
Nightly features CI

joechrisellis:
This adds support for testing features that require a nightly Rust build.

Seems like a reasonable way to do this, for now. I don't imagine that we will have a _lot_ of Parsec features in the future that will require something like this, but this should do the trick for the peer credentials authenticator stuff for now. 😄

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -122,6 +122,14 @@ fi
 echo "Unit, doc and integration tests"
 RUST_BACKTRACE=1 cargo test $FEATURES
 
+# Run tests for Parsec features that require a nightly version of the compiler
+# to function. For example: Unix peer credentials authenticator is currently
+# only functional if we use the nightly compiler, since it depends on the
+# `peer_cred` function.
+# TODO: change this list as features trickle into the stable branch.
+echo "Unit testing peer credentials authenticator"
+RUST_BACKTRACE=1 cargo +nightly test --features=peer-credentials-authenticator
+
 # Removing any mappings left over from integration tests
 rm -rf mappings/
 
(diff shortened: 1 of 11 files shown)


joechrisellis:
We'll have to wait until (if?) the [Unix peer credentials stuff](https://github.com/rust-lang/rust/pull/75148) is in the nightly compiler for this to pass CI.

ci.sh
```diff
@@ -122,6 +122,15 @@ fi
...
 
+# Run tests for Parsec features that require a nightly version of the compiler
+# to function. For example: Unix peer credentials authenticator is currently
+# only functional if we use the nightly compiler, since it depends on the
+# `peer_cred` function.
+# TODO: change this list as features trickle into the stable branch.
+echo "Nightly features"
+NIGHTLY_FEATURES="--features=peer-credentials-authenticator"
+RUST_BACKTRACE=1 cargo test +nightly $NIGHTLY_FEATURES
+
```
hug-dev, line 133: As the changes to add this test are fairly small, I propose to add it directly in the other PR so that we can reload the CI once the feature is in `rustc` 😃 

In the CI logs I can see:
```
error[E0554]: `#![feature]` may not be used on the stable release channel
```
are you sure the nightly version is installed? Or maybe this is because the `+nightly` needs to be before `test`?

Also last thing, I think this is fine to just have a section here specific to test the peer credentials authenticator:

```
echo "Unit testing peer credential authenticator"
RUST_BACKTRACE=1 cargo +nightly test --features=peer-credentials-authenticator
```

joechrisellis:
Closing.

This PR is not needed any more. We have cherry-picked the peer credentials code from the Rust standard library into Parsec so that we can use the functionality 'early' -- the relevant commit is in this PR: https://github.com/parallaxsecond/parsec/pull/214

We have also removed the `unix-peer-credentials-authenticator` feature flag.

PR #246, opened 2020-09-10
Upgrade and clean dependencies

hug-dev:
Also adds a nightly workflow


hug-dev:
I have:
```
2020-09-14T10:35:00.9127176Z [INFO  parsec_service::providers::pkcs11_provider::key_management] Generating RSA key pair in session 2
2020-09-14T10:35:00.9128115Z [ERROR parsec_service::providers::pkcs11_provider::key_management] Generate Key Pair operation failed; Error: PKCS#11: CKR_GENERAL_ERROR (0x5)
2020-09-14T10:35:00.9129244Z [ERROR parsec_service::providers::pkcs11_provider::utils] Error converted to PsaErrorCommunicationFailure; Error: 5
```
everytime I generate a key. I believe this is something wrong I do with casting but interesting in the sense that now all the tests are single-threaded and this error seems similar to the one we had before...
Will try to see if I can reproduce locally
cc @ionut-arm

e2e_tests/provider_cfg/mbed-crypto/Dockerfile
```diff
@@ -8,10 +8,6 @@ RUN apt-get update && \
 	apt-get install -y pkg-config libssl-dev
 
 WORKDIR /tmp
-RUN wget https://github.com/ARMmbed/mbed-crypto/archive/mbedcrypto-2.0.0.tar.gz
-RUN tar xf mbedcrypto-2.0.0.tar.gz
-RUN cd mbed-crypto-mbedcrypto-2.0.0 \
-		&& make
```
hug-dev, line 14: This is no longer needed because `psa-crypto-sys` does it for us 🤓

src/providers/pkcs11_provider/utils.rs
```diff
@@ -558,7 +563,7 @@ pub fn parsec_to_pkcs11_params(
             pub_template.push(CK_ATTRIBUTE::new(CKA_TOKEN).with_bool(&CK_TRUE));
             pub_template.push(CK_ATTRIBUTE::new(CKA_PRIVATE).with_bool(&CK_FALSE));
             pub_template.push(CK_ATTRIBUTE::new(CKA_PUBLIC_EXPONENT).with_bytes(&PUBLIC_EXPONENT));
-            pub_template.push(CK_ATTRIBUTE::new(CKA_MODULUS_BITS).with_ck_ulong(&attributes.bits));
```
hug-dev, line 561: I don't know how this was working before 🔫 `attributes` is moved into this function but then freed! This reference would no longer be valid then...
I tried, after applying this change, to remove the mutex in the PKCS11 provider but the bug is still there.

ionut-arm (approved):
Looks good to me. Hopefully this removes the last memory issues in the PKCS11 provider ... 😔

PR #248, opened 2020-09-14
Drop the direct dependency on picky

puiterwijk:
The Picky crate is not actually used anywhere in the code, just the
picky-asn1* sub-crates.


hug-dev:
True! Thanks for the PR 😃 I actually removed all unused dependencies in #246 using `cargo udeps`. It is failing CI but hopefully will be merged this week!

hug-dev:
#246 is merged so the dependency on `picky` should be removed on `master` 😃

hug-dev:
Closing this as the dependency is no longer there on latest `master`!

PR #252, opened 2020-09-17
Fix sign attribute in PKCS11

ionut-arm:
This commit fixes a bug in the conversion process for key attributes in the
PKCS11 provider. Instead of CKA_SIGN_RECOVER, CKA_SIGN is to be used.


hug-dev (approved):
💯

paulhowardarm (approved):

PR #253, opened 2020-09-18
Improve E2E testing

ionut-arm:
This commit improves integration testing by adding tests for:
* cross-checking compatibility for handling key material and other
buffer encodings
* fault testing the handling of request header values
* expand stress testing to cover asymmetric encryption and
decryption, and ECC sign/verify
* checking that operations that are not in the supported opcodes
list are indeed not supported
Some other improvements:
* moving testing for ECC signatures from integration testing to E2E
testing
* change the `export` usage flags to `false` unless explicitly required
by the test


e2e_tests/tests/all_providers/cross.rs
```diff
@@ -0,0 +1,228 @@
...
+    ciphertext: Vec<u8>,
+) -> Result<Vec<u8>> {
+    client.set_provider(provider);
+    client.asymmetric_decrypt_message_with_rsapkcs1v15(key_name, ciphertext)
+}
+
+#[test]
+fn tpm_sign_cross() {
+    let key_name = String::from("tpm_sign_cross");
+    let (mut client, pub_key, signature) = setup_sign(ProviderID::Tpm, key_name.clone()).unwrap();
```
hug-dev, line 84: Small nit and maybe not good for performance but we could put this function in a `for provider in [Tpm, MbedCrypto, Pkcs11]` loop

e2e_tests/tests/all_providers/mod.rs
```diff
@@ -5,6 +5,8 @@ use parsec_client::core::interface::operations::list_providers::Uuid;
 use parsec_client::core::interface::requests::{AuthType, Opcode, ProviderID, Result};
 use std::collections::HashSet;
 
+mod cross;
```
hug-dev, line 8: Is that used?

hug-dev: Oh this might be needed so that test are executed? Even without `pub` 🤯

hug-dev (commented):
Thanks a lot for your effort! I will update parallaxsecond/parsec-book#65 with these changes

hug-dev (approved):

e2e_tests/tests/all_providers/cross.rs
ionut-arm: Hmm, yeah, I guess! My only concern was with figuring out what failed and where, but I guess we can put a `format!` in the `expect(...)`

e2e_tests/tests/all_providers/mod.rs
ionut-arm: Indeed, it's only about the tests being in _some_ module

e2e_tests/src/stress.rs
```diff
@@ -46,14 +46,18 @@ fn generate_string(size: usize) -> String {
...
 }
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, PartialEq)]
 enum Operation {
     CreateDestroyKey,
     Sign,
     Verify,
+    SignEcc,
+    VerifyEcc,
```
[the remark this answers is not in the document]
ionut-arm: Yeah, unfortunately PKCS11 doesn't support it -_- one day!!!

PR #254, opened 2020-09-18
Fix attribute conversion in PKCS11 provider

ionut-arm:
This commit fixes the attribute conversion in the PKCS11 provider,
more specifically the previously-inverted attributes for export.


hug-dev (approved):

paulhowardarm (approved):
The logic looks correct now, although inevitably this will still cause test failures on the NXP Layerscape platform because exportable private keys are not valid to begin with. However, it would be wrong to try to address that here. This is a good PR because it clearly shows the faulty logic being corrected, which is a necessary change. I will raise a separate issue for remaining failures on Layerscape once I can properly isolate them.

PR #261, opened 2020-10-08
Add list keys

joechrisellis:
Implementation for `ListKeys`. This won't immediately pass CI because it's pending a version bump on `parsec-interface-rs`.

### DIFF
--- a/e2e_tests/tests/all_providers/mod.rs
+++ b/e2e_tests/tests/all_providers/mod.rs
@@ -65,6 +65,7 @@ fn list_opcodes() {
     let _ = core_provider_opcodes.insert(Opcode::ListProviders);
     let _ = core_provider_opcodes.insert(Opcode::ListAuthenticators);
     let _ = core_provider_opcodes.insert(Opcode::ListOpcodes);
+    let _ = core_provider_opcodes.insert(Opcode::ListKeys);
 
     assert_eq!(
         client
@@ -112,4 +113,37 @@ fn sign_verify_with_provider_discovery() -> Result<()> {
     let mut client = TestClient::new();
     let key_name = String::from("sign_verify_with_provider_discovery");
     client.generate_rsa_sign_key(key_name)
+}
+
+#[test]
+fn list_keys() {
+    let mut client = TestClient::new();
+    client.set_auth("list_keys test".to_string());
+
+    let keys = client.list_keys().expect("list_keys failed");
+
+    assert!(keys.is_empty());
+
+    let key1 = String::from("list_keys1");
+    let key2 = String::from("list_keys2");
+    let key3 = String::from("list_keys3");
+
+    client.set_provider(ProviderID::MbedCrypto);
+    client.generate_rsa_sign_key(key1.clone()).unwrap();
+    client.set_provider(ProviderID::Pkcs11);
+    client.generate_rsa_sign_key(key2.clone()).unwrap();
+    client.set_provider(ProviderID::Tpm);
+    client.generate_rsa_sign_key(key3.clone()).unwrap();
+
+    let key_names: Vec<String> = client
+        .list_keys()
+        .expect("list_keys failed")
+        .into_iter()
+        .map(|k| k.name)
+        .collect();
+
+    assert_eq!(key_names.len(), 3);
+    assert!(key_names.contains(&key1));
+    assert!(key_names.contains(&key2));
+    assert!(key_names.contains(&key3));
 }
--- a/src/back/backend_handler.rs
+++ b/src/back/backend_handler.rs
@@ -8,7 +8,7 @@
 use crate::authenticators::ApplicationName;
 use crate::providers::Provide;
 use derivative::Derivative;
-use log::trace;
+use log::{error, trace};
 use parsec_interface::operations::Convert;
 use parsec_interface::operations::{NativeOperation, NativeResult};
 use parsec_interface::requests::{
@@ -53,11 +53,18 @@ impl BackEndHandler {
     /// the request.
     ///
     /// # Errors
+    /// - if the provider ID can not perform the type of operation, returns
+    /// `ResponseStatus::PsaErrorNotSupported`
     /// - if the provider ID does not match, returns `ResponseStatus::WrongProviderID`
     /// - if the content type does not match, returns `ResponseStatus::ContentTypeNotSupported`
     /// - if the accept type does not match, returns `ResponseStatus::AcceptTypeNotSupported`
     pub fn is_capable(&self, request: &Request) -> Result<()> {
         let header = &request.header;
+
+        if (self.provider_id == ProviderID::Core) != header.opcode.is_core() {
+            error!("The request's operation is not compatible with the provider targeted.");
+            return Err(ResponseStatus::PsaErrorNotSupported);
+        }
 
         if header.provider != self.provider_id {
             Err(ResponseStatus::WrongProviderID)
@@ -206,6 +213,14 @@ impl BackEndHandler {
                     .list_authenticators(op_list_authenticators));
                 trace!("list_authenticators egress");
                 self.result_to_response(NativeResult::ListAuthenticators(result), header)
+            }
+            NativeOperation::ListKeys(op_list_keys) => {
+                let app_name =
+                    unwrap_or_else_return!(app_name.ok_or(ResponseStatus::NotAuthenticated));
+                let result =
+                    unwrap_or_else_return!(self.provider.list_keys(app_name, op_list_keys));
+                trace!("list_keys egress");
+                self.result_to_response(NativeResult::ListKeys(result), header)
             }
             NativeOperation::PsaHashCompute(op_hash_compute) => {
                 let _app_name =
--- a/src/key_info_managers/mod.rs
+++ b/src/key_info_managers/mod.rs
@@ -145,3 +145,40 @@ pub trait ManageKeyInfo {
     /// Returns an error as a String if there was a problem accessing the Key Info Manager.
     fn exists(&self, key_triple: &KeyTriple) -> Result<bool, String>;
 }
+
+/// Returns a Vec of the KeyInfo objects corresponding to the given application name and
+/// provider ID.
+///
+/// # Errors
+///
+/// Returns an error as a String if there was a problem accessing the Key Info Manager.
+pub fn list_keys(
+    manager: &dyn ManageKeyInfo,
+    app_name: &ApplicationName,
+    provider_id: ProviderID,
+) -> Result<Vec<parsec_interface::operations::list_keys::KeyInfo>, String> {
+    use parsec_interface::operations::list_keys::KeyInfo;
+
+    let mut keys: Vec<KeyInfo> = Vec::new();
+    let key_triples = manager.get_all(provider_id)?;
+
+    for key_triple in key_triples {
+        if key_triple.app_name() != app_name {
+            continue;
+        }
+
+        let key_info = manager.get(key_triple)?;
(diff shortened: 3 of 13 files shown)


src/providers/core_provider/mod.rs
```diff
@@ -73,6 +85,45 @@ impl Provide for CoreProvider {
         })
     }
 
+    fn list_keys(&self, _op: list_keys::Operation) -> Result<list_keys::Result> {
```
hug-dev, line 88: This operation should only list the keys belonging to a specific `ApplicationName`, so you should also add this parameter here and decode it from the auth payload in the back end handler (as it is done for crypto operations on keys).

src/providers/mod.rs
```diff
@@ -91,13 +93,15 @@ impl ProviderConfig {
...
-    psa_asymmetric_decrypt, psa_asymmetric_encrypt, psa_destroy_key, psa_export_key,
-    psa_export_public_key, psa_generate_key, psa_generate_random, psa_hash_compare,
+    list_authenticators, list_keys, list_opcodes, list_providers, ping, psa_aead_decrypt,
+    psa_aead_encrypt, psa_asymmetric_decrypt, psa_asymmetric_encrypt, psa_destroy_key,
+    psa_export_key, psa_export_public_key, psa_generate_key, psa_generate_random, psa_hash_compare,
     psa_hash_compute, psa_import_key, psa_raw_key_agreement, psa_sign_hash, psa_verify_hash,
 };
 use parsec_interface::requests::{ResponseStatus, Result};
 
+type KeyInfoStore = Arc<RwLock<dyn ManageKeyInfo + Send + Sync>>;
```
hug-dev, line 103: I personally do not like type aliasing that much... They kind of make it one step harder to find what the type really is and offer no static checks. I have previously made review comments similar like this on other code so we are trying to not use them too much.

I see the advantage of making the interfaces easier to read but it is not worth it to me.

src/providers/mod.rs
```diff
@@ -111,6 +115,13 @@ pub trait Provide {
         Err(ResponseStatus::PsaErrorNotSupported)
     }
 
+    /// Get the Provider ID the provider.
+    fn provider_id(&self) -> ProviderID;
+
+    /// Get an optional reference to the KeyInfoStore for this provider. Providers that do not
+    /// implement a key store should return None.
+    fn get_key_info_store(&self) -> Option<&KeyInfoStore>;
+
```
hug-dev, line 124: The `Provide` trait should ideally only contain the actual operations that the providers can perform with the exception of when it would be really hard/unpractical to do otherwise like with the `describe` function above. But that one should maybe be in another trait!

My point is that, from the `CoreProvider`, you can directly get this information by adding a field containing a vector of all `ManageKeyInfo` implementation and use that trait to get the key information. You can add all of them to the `CoreProvider` during service building.

hug-dev (commented):
Thanks for adding support for this! I added a few comments of another way to do it that fits better (I think) with our current architecture.

src/providers/mod.rs
joechrisellis: That's fair enough if this is a design choice we're rolling with -- will revert. 🙂

src/providers/mod.rs
joechrisellis: Gotcha. Will change this too.

src/providers/core_provider/mod.rs
joechrisellis: ACK, will add this 😄

src/providers/mod.rs
ionut-arm: I guess it's a balance between readability and ease of use. It's kinda similar to the reason why `auto` exists in C++, to make convoluted types opaque.

> They kind of make it one step harder to find what the type really is
 
That's only true if you don't use an IDE 😛 I prefer the aliasing in cases like this where there are a lot of qualifiers and a type name ends up larger than the rest of a function signature, but I'm not that bothered about it one way or another.

src/providers/mod.rs
ionut-arm: This is where a `list_keys` implementation on every provider would've helped 🤡 `CoreProvider` would then just aggregate. 

Also, I'd rather go down the route of asking the provider to tell you which `ManageKeyInfo` it's using, because otherwise imagine having two different key info stores and switching between them at some point. The `ListKeys` would then report keys that you don't have access to - you see them because they're saved in a different key info manager, but the actual provider can't see them.

src/key_info_managers/mod.rs
```diff
@@ -144,4 +144,41 @@ pub trait ManageKeyInfo {
...
     /// Returns an error as a String if there was a problem accessing the Key Info Manager.
     fn exists(&self, key_triple: &KeyTriple) -> Result<bool, String>;
+
+    /// Returns a Vec of the KeyInfo objects corresponding to the given application name and
+    /// provider ID.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error as a String if there was a problem accessing the Key Info Manager.
+    fn list_key_info(
```
hug-dev, line 154: I think I would prefer having this method outside of the trait so that it can not be overwritten:
```rust
fn list_key_info(
    &impl ManageKeyInfo,
    app_name: &ApplicationName,
    provider_id: ProviderID,
)
```

src/providers/mod.rs
```diff
@@ -132,6 +142,16 @@ pub trait Provide {
         Err(ResponseStatus::PsaErrorNotSupported)
     }
 
+    /// Lists all keys belonging to the application.
+    fn list_keys(
```
hug-dev, line 146: I think it would be nicer to only have the `list_keys` method and:
* in all the crypto providers make it the same as what `find_keys` currently does
* in the code provider call `list_keys` on all provider

Although we need to be sure somewhere that `ListKeys` can not be called directly on the crypto providers. Probably somewhere higher on the stack.

src/back/backend_handler.rs
```diff
@@ -207,6 +207,14 @@ impl BackEndHandler {
                 trace!("list_authenticators egress");
                 self.result_to_response(NativeResult::ListAuthenticators(result), header)
             }
+            NativeOperation::ListKeys(op_list_keys) => {
+                let app_name =
```
hug-dev, line 211: The check I was talking in the other comment could be here but could also be in the `is_capable` method where we would check if the request's opcode is a Core or a Crypto operation and based on that check if the provider can do it or not.

hug-dev (commented):
That looks good! One suggestion to only need one function from the `Provide` trait.
Also could you add an end-to-end test which creates a few keys, list them and expect the correct result 🎸 ?

hug-dev:
@joechrisellis gave me permission to continue the work on his fork, so I went ahead and addressed the comments of my own review 🤓

src/back/backend_handler.rs
```diff
@@ -53,12 +53,19 @@ impl BackEndHandler {
...
     /// # Errors
+    /// - if the provider ID can not perform the type of operation, returns
+    /// `ResponseStatus::PsaErrorNotSupported`
     /// - if the provider ID does not match, returns `ResponseStatus::WrongProviderID`
     /// - if the content type does not match, returns `ResponseStatus::ContentTypeNotSupported`
     /// - if the accept type does not match, returns `ResponseStatus::AcceptTypeNotSupported`
     pub fn is_capable(&self, request: &Request) -> Result<()> {
         let header = &request.header;
 
+        if (self.provider_id == ProviderID::Core) ^ header.opcode.is_core() {
```
ionut-arm, line 64:
```suggestion
        if (self.provider_id == ProviderID::Core) != header.opcode.is_core() {
```

Seems easier to read/understand

hug-dev: I think though with `!=` it will not go into the `if` in the following counter-example:
* `self.provider_id` is `Provider::Tpm`
* `header.opcode` is `Opcode::Ping`

hug-dev: I will update, your version is definitely better

e2e_tests/tests/all_providers/mod.rs
```diff
@@ -113,3 +114,31 @@ fn sign_verify_with_provider_discovery() -> Result<()> {
...
     client.generate_rsa_sign_key(key_name)
 }
+
+#[test]
+fn list_keys() {
+    let mut client = TestClient::new();
+
+    let keys = client.list_keys().expect("list_keys failed");
+
+    assert!(keys.is_empty());
```
ionut-arm, line 124: If there are other concurrent tests here, this might fail. I think easiest way out is to just set a different identity for the client used in this test.

e2e_tests/tests/all_providers/mod.rs
```diff
@@ -113,3 +114,31 @@ fn sign_verify_with_provider_discovery() -> Result<()> {
     let key_name = String::from("sign_verify_with_provider_discovery");
     client.generate_rsa_sign_key(key_name)
 }
+
+#[test]
+fn list_keys() {
```
ionut-arm, line 119: Also, shouldn't this be run for each provider, just in case?

src/key_info_managers/mod.rs
```diff
@@ -145,3 +145,40 @@ pub trait ManageKeyInfo {
...
+
+        keys.push(KeyInfo {
+            provider_id: ProviderID::MbedCrypto,
+            name: key_triple.key_name().to_string(),
+            attributes: key_info.attributes,
+        });
+    }
+
+    Ok(keys)
+}
```
ionut-arm, line 184: You can actually include this in the `ManageKeyInfo` trait, because you only rely on methods on `manager` that are exposed through that trait. You'd just change `manager.get<all>` with `self.get<all>`

src/providers/tpm/mod.rs
```diff
@@ -1,22 +1,23 @@
-// Copyright 2019 Contributors to the Parsec project.
+//KeyInfoStore Copyright 2019 Contributors to the Parsec project.
```
ionut-arm, line 1: ?!?!

e2e_tests/tests/all_providers/mod.rs
hug-dev: My idea was that `list_keys` should return keys from all providers, that's why I put it there. I should test that by creating a key in all providers!

src/key_info_managers/mod.rs
hug-dev: The reason why I did not include it is to prevent overwriting this method. I think the result is the same?

e2e_tests/tests/all_providers/mod.rs
ionut-arm: Ah, yes, that makes more sense!

src/providers/tpm/mod.rs
hug-dev: `vim` played me like a fool once again

src/key_info_managers/mod.rs
ionut-arm: > to prevent overwriting this method

To prevent overwriting what?

> I think the result is the same?

It is, in the end, I just tend to prefer methods to functions when the function is clearly acting on a specific type.

hug-dev: Well I might be wrong but if a method is in a trait then implementation can either not implement this method and use the default implementation, or overwrite it.

ionut-arm: Ah, yes, true, but we have control over that. And it might be that that's an advantage, because for a database, for example, we might be able to build a query to get only the entries that match those requirements instead of filtering ourselves.

ionut-arm (approved):
👍

PR #262, opened 2020-10-09
Bugfix: use `UnixPeerCredentials` instead of `PeerCredentials`

joechrisellis:
Really quick bugfix. Looks like is this blocking CI [here](https://github.com/parallaxsecond/parsec/pull/261).


joechrisellis:
Ignore this -- this will be addressed later with a version bump.

Issue #245, opened 2020-09-09
Add multitenancy testing infrastructure 👩‍🔧


hug-dev:
With the various authenticators that we have and are going to have, it will be helpful to add automated testing to check that multitenancy is working securely.

As ideas for test cases:

1. The service should fail to run (under any authentication) if the socket folder has wrong permissions
2. Under direct authentication using the `parsec-clients` group, a client not in that group should not be able to connect
3. Under peer credential authentication (when ready):
    1. even a client not in `parsec-clients` should be able to connect
    2. one client can not access another client's keys

Regarding the infrastructure, these tests could be implemented in a separate GitHub Actions workflow (or more than one!), in a new top-level directory (eg `multitenancy_tests`).

hug-dev:
Some work could be re-used from #218

hug-dev:
Remainder to add the new tests added [here](https://parallaxsecond.github.io/parsec-book/parsec_service/tests/existing_tests.html)

PR #273, opened 2020-10-14
Add authentication configuration

### LINKED ISSUE #249: Parsec with multiple authenticators
With #200 and #232 we can imagine a future where we will have three possible authentication methods:

1. Direct authentication using the `parsec-cliens` group to restrict access on the socket
2. Unix Peer Credentials authentication
3. SPIFFE/SPIRE based authentication

This raises multiple questions:
* How will each one of them be activated? What is the default one?
* What is the impact on [the threat model](https://parallaxsecond.github.io/parsec-book/threat_model/threat_model.html) (both service and client) and the [recommendations for secure deployment](https://parallaxsecond.github.io/parsec-book/threat_model/secure_deployment.html)?
* What is the impact on the [permission checks](https://github.com/parallaxsecond/parsec/blob/72ebbeed565f635d37e17ad38195cd117f84edc0/src/front/domain_socket.rs#L43) that we do in the service under the `no-parsec-user-and-clients-group` feature?
* What is the impact on the [permission checks](https://github.com/parallaxsecond/parsec-client-rust/blob/6dbc751e86e2fac6c4e487d45f7f3d36f2a27ddd/src/core/ipc_handler/unix_socket.rs#L64) we do in the client under the `no-fs-permission-check` feature?

As our threat model and code assume that the deployment is secure enough for direct authentication, we can securely add more authenticators in the code base respecting the same requirements.
However, if we want to loosen those requirements when using other authenticators and remove the current restrictions for Parsec clients:
* they have to trust each other
* they have to be part of the `parsec-clients` group

The questions above will have to be answered and modifications done in the code and the documentation.

### LINKED ISSUE #270: Add authenticator configuration
Add an authenticator configuration section in the `config.toml` file. The list is prioritised.
At first the possible options will be direct auth, unix peer cred, in the future jwt svid.

hug-dev:
To be rebased on top of #261 

Fix #270 
Fix #249

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,9 +2,9 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.7.13"
+version = "0.7.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "043164d8ba5c4c3035fec9bbee8647c0261d788f3474306f93bb65901cae0e86"
+checksum = "b476ce7103678b0c6d3d395dbbae31d48ff910bd28be979ba5d48c6351131d0d"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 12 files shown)


ionut-arm (approved):
Tests go in a different issue, right?

hug-dev:
> Tests go in a different issue, right?

Yes! All multitenancy tests will be done as part of #245

PR #274, opened 2020-10-16
Remove warnings about parsec and parsec-clients

hug-dev:
Add one if the Direct Authenticator is chosen as the default one.


ionut-arm (approved):
🚀

PR #275, opened 2020-10-16
Put config tests in all_providers

hug-dev:
That will allow adding more all_providers tests easily.


ionut-arm (approved):
🧪

PR #276, opened 2020-10-16
Add multitenancy tests

### LINKED ISSUE #245: Add multitenancy testing infrastructure 👩‍🔧

hug-dev:
This PR adds multitenancy tests which execute various operations under different users and check that one user can not use the keys of another one.

Fix #245

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -135,8 +135,23 @@ sleep 5
 pgrep -f target/debug/parsec >/dev/null
 
 if [ "$PROVIDER_NAME" = "all" ]; then
-    echo "Execute all-providers tests"
+    echo "Execute all-providers normal tests"
     RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml all_providers::normal
+
+    echo "Execute all-providers multi-tenancy tests"
+    su -c "RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml --target-dir /home/parsec-client-1 all_providers::multitenancy::client1_before" parsec-client-1
+    su -c "RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml --target-dir /home/parsec-client-2 all_providers::multitenancy::client2" parsec-client-2
+    su -c "RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml --target-dir /home/parsec-client-1 all_providers::multitenancy::client1_after" parsec-client-1
+    # Change the authentication method
+    sed -i 's/^\(auth_type\s*=\s*\).*$/\1\"UnixPeerCredentials\"/' $CONFIG_PATH
+    pkill -SIGHUP parsec
+    sleep 5
+    su -c "RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml --target-dir /home/parsec-client-1 all_providers::multitenancy::client1_before" parsec-client-1
+    su -c "RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml --target-dir /home/parsec-client-2 all_providers::multitenancy::client2" parsec-client-2
+    su -c "RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml --target-dir /home/parsec-client-1 all_providers::multitenancy::client1_after" parsec-client-1
+
+    # Last test as it changes the service configuration
+    echo "Execute all-providers config tests"
     RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml all_providers::config -- --test-threads=1
 else
     # Per provider tests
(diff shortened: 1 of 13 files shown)


hug-dev:
I keep facing issues because of switching user to compile and execute the tests. I think I would have to clone the `parsec` repo under each user's home directory and run the `cargo test` command there.
Because the tests are not doing anything fancy, it might just be easier to write them in bash and use `parsec-tool`. That way they will be easier to work, understand and maintain.

hug-dev:
To use `parsec-tool` for testing, there will need to be an option there to modify the socket path, to use `/tmp/parsec.sock`

ionut-arm:
> 
> 
> I keep facing issues because of switching user to compile and execute the tests. I think I would have to clone the `parsec` repo under each user's home directory and run the `cargo test` command there.
> Because the tests are not doing anything fancy, it might just be easier to write them in bash and use `parsec-tool`. That way they will be easier to work, understand and maintain.

Is it not possible to create symlinks for every user to the "root" Parsec folder?

ci.sh
```diff
@@ -135,8 +135,23 @@ sleep 5
 pgrep -f target/debug/parsec >/dev/null
 
 if [ "$PROVIDER_NAME" = "all" ]; then
-    echo "Execute all-providers tests"
+    echo "Execute all-providers normal tests"
     RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml all_providers::normal
+
+    echo "Execute all-providers multi-tenancy tests"
+    su -c "RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml --target-dir /home/parsec-client-1 all_providers::multitenancy::client1_before" parsec-client-1
```
hug-dev, line 142: I was being a bit too dramatic. The problem was that each user was trying to write the `target` directory owned by `root`. Using the `--target-dir` option fixed the problem 🚀

e2e_tests/src/lib.rs
```diff
@@ -36,7 +35,7 @@ const TEST_TIMEOUT: Duration = Duration::from_secs(1);
 #[derive(Debug)]
 pub struct TestClient {
     basic_client: BasicClient,
-    created_keys: Option<HashSet<(String, String, ProviderID)>>,
+    created_keys: Option<HashSet<(String, Option<String>, ProviderID)>>,
```
ionut-arm, line 38: Wouldn't it be easier to extend for JWT if you just put `Authentication` here, instead of `Option<String>`?

e2e_tests/tests/all_providers/multitenancy.rs
```diff
@@ -0,0 +1,69 @@
...
+// 3. client1_after is executed as parsec-client-1
+//
+// They are executed against all possible authenticators in Parsec.
+
+#[test]
+fn client1_before() {
+    // Create one key on each provider
+    let mut client = TestClient::new();
+    client.do_not_destroy_keys();
+    client.set_direct_auth("client1".to_string());
```
ionut-arm, line 18: I'm not sure I get the purpose of these tests - shouldn't they use peer credential auth?

ionut-arm (commented):
Thanks for the work to make these run 🏃

e2e_tests/src/lib.rs
hug-dev: I think the extension would not have to touch those parts as for JWT, there will be similarly than `UnixPeerCredentials` a new variant of `Authentication` without containing any information.

I mean the code would not have to change to work for JWT in the future.

e2e_tests/tests/all_providers/multitenancy.rs
hug-dev: They do use Peer Credentials as well! This line will get ignored if the default authentication selected in the client during construction is `UnixPeerCredentials`. This directly call `set_default_auth(Some("client1"))` and will do something when direct authentication is used.
This is done so that the same test can be used for direct authenticaiton and peer cred/jwt.

ionut-arm: Hmm, then maybe the method name is a bit misleading - why not just use the same name, `set_default_auth`?

e2e_tests/src/lib.rs
ionut-arm: Yes, my thought was also that this could be used in the future when there is more than one authenticator and you create keys with them, so you can't just rely on the "default". But that's probs premature optimization.

e2e_tests/tests/all_providers/multitenancy.rs
hug-dev: Changed it to the same name!

e2e_tests/src/lib.rs
```diff
@@ -927,7 +906,9 @@ impl Drop for TestClient {
         if let Some(ref mut created_keys) = self.created_keys {
             for (key_name, auth, provider) in created_keys.clone().iter() {
                 self.set_provider(*provider);
-                self.set_auth(auth.clone());
+                if let Some(app_name) = auth {
+                    self.set_default_auth(Some(app_name.clone()));
+                }
```
ionut-arm, line 911:
```suggestion
                self.set_default_auth(auth);
```

ionut-arm (approved):
👌

PR #277, opened 2020-10-22
Remove PKCS11 single thread lock (#264)

### LINKED ISSUE #264: PKCS11 multi-threading
Currently our PKCS11 provider is forcibly kept single-threaded - all crypto or key management operations have to take a lock that ensures only one is executing at any time. This is because of an issue probably found in SoftHSM2. We [raised a point](https://github.com/opendnssec/SoftHSMv2/issues/573) with them about it and got a partial fix, but even with that fix there are still failures.
The problem is that when run in a multi-threaded way, without locking, generate and import key operations start failing (with `CKR_GENERAL_ERROR` or `CKR_TEMPLATE_INCONSISTENT`). Given that this happens pretty consistently, the CI would've failed so we made the decision to simply lock down the provider. This comes at the cost of performance of said provider, but we're not that worried about that for now.

On the SoftHSM side https://github.com/opendnssec/SoftHSMv2/pull/576 was raised to fix this and I've tested with the initial commits. The issue was still happening, but much less frequently. I've yet to test with the latest commit as well - hopefully that fixes the issue and we can then remove the locking.

ionut-arm:
This commit removes the mutex that was used to force the PKCS11 provider
to work in a single-threaded way. The issue was fixed in
https://github.com/opendnssec/SoftHSMv2/pull/576 and our Dockerfiles
have been updated to use that commit.

Fixes #264

### DIFF
--- a/e2e_tests/provider_cfg/all/Dockerfile
+++ b/e2e_tests/provider_cfg/all/Dockerfile
@@ -4,7 +4,7 @@ ENV PKG_CONFIG_PATH /usr/local/lib/pkgconfig
 
 RUN apt-get update && \
 	apt-get install -y git make gcc python3 python curl wget cmake && \
-	apt-get install -y automake autoconf libtool pkg-config libssl-dev && \
+	apt-get install -y automake autoconf libtool pkg-config libssl-dev libgcc1 && \
 	# These libraries are needed for bindgen as it uses libclang.so
 	apt-get install -y clang libclang-dev libc6-dev-i386 && \
 	# Install cargo globally to not have to install it for each user for multitenancy tests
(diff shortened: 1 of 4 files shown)


hug-dev (approved):
Nice, that's egg-cellent 🥚 
I agree to merge it once you have the CI passing 🚋

e2e_tests/provider_cfg/all/Dockerfile
```diff
@@ -17,6 +17,19 @@ RUN cd mbed-crypto-mbedcrypto-2.0.0 \
...
 
 WORKDIR /tmp
+# Download and install SoftHSM2
+RUN git clone https://github.com/opendnssec/SoftHSMv2.git \
+	&& cd SoftHSMv2 \
+	&& git reset --hard 20a53bd083a6134ce2230f80edda5dc8be0366bd
+
+RUN cd SoftHSMv2 \
+	&& sh autogen.sh \
+	|| sh autogen.sh
```
hug-dev, line 27: Would be nice to link with the bug description as a comment here for an external eye to understand why this was needed

hug-dev (approved):

e2e_tests/provider_cfg/all/Dockerfile
ionut-arm: Done

hug-dev (approved):

PR #283, opened 2020-11-06
Add a JWT-SVID authenticator

hug-dev:
The authenticators validates a JWT SPIFFE Verifiable Identity Document.
Adds a configuration option to choose this new authenticator and also
adds tests for it.


hug-dev:
This won't work until the changes are pushed in some `rust-spiffe` repository 🥇

src/authenticators/jwt_svid_authenticator/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+                    );
+                    ResponseStatus::InvalidEncoding
+                })?
+                .to_string(),
+        );
+        let audience = String::from("parsec");
+
+        let validate_response = self
+            .jwt_client
+            .validate(audience, svid, Some(Duration::new(5, 0)))
```
ionut-arm, line 67: Nice, maybe we can make that `Duration` configurable in the future (when Spire actually supports it, too)

src/authenticators/jwt_svid_authenticator/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+        let svid = Jwt::new(
+            str::from_utf8(auth.buffer.expose_secret())
+                .map_err(|e| {
+                    error!(
+                        "The authentication buffer can not be parsed into a UTF-8 string ({}).",
+                        e
+                    );
+                    ResponseStatus::InvalidEncoding
+                })?
+                .to_string(),
```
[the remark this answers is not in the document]
ionut-arm: (Asking because for some reason I thought this might be a JWT base64 encoded)

ionut-arm (commented):
Looks good!

src/authenticators/jwt_svid_authenticator/mod.rs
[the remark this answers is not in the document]
ionut-arm: Ah, I read about it now from the book PR. I think it's safe to assume `Jwt::new` would return a `Result` since it has to verify it's in the correct encoding? Tbh, maybea method like `Jwt::from_compact_serialization` would be more correct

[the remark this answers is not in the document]
hug-dev: Yes agree, ideally the `Jwt` constructor would parse the Base 64-url encoded token and verify that the format is good. I used a very simple approach here where the Rust SPIFFE library does not do anything fancy. If the token is malformed, I expect the validation step to fail and return an error.

src/authenticators/jwt_svid_authenticator/mod.rs
hug-dev: Yes! Though note that this time is not the expiration time of the token but the backoff time I believe.

ionut-arm: Ah, makes sense, got confused there!

ionut-arm (approved):

Cargo.toml
```diff
@@ -45,6 +44,8 @@ picky-asn1-x509 = { version = "0.3.2", optional = true }
 users = "0.10.0"
 libc = "0.2.77"
 anyhow = "1.0.32"
+# Using a fork until the JWT support is merged into the main rust-spiffe repository
+spiffe = { git = "https://github.com/hug-dev/rust-spiffe", branch = "refactor-jwt" }
```
ionut-arm, line 48: This means we cannot upstream to crates.io, right?

e2e_tests/provider_cfg/all/Dockerfile
```diff
@@ -55,3 +53,17 @@ RUN softhsm2-util --init-token --slot 0 --label "Parsec Tests" --pin 123456 --so
...
+USER parsec-client-2
+RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
+
+# Install Rust toolchain for root
+USER root
+RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+# Download the SPIRE server and agent
+RUN curl -s -N -L https://github.com/spiffe/spire/releases/download/v0.11.1/spire-0.11.1-linux-x86_64-glibc.tar.gz | tar xz
```
ionut-arm, line 69: We should really look at upstreaming these images 😬 It would save 10 min for the `all-providers` run, almost 25% of the time.

ci.sh
```diff
@@ -139,16 +157,31 @@ if [ "$PROVIDER_NAME" = "all" ]; then
...
+    # Needed because parsec-client-1 and 2 write to those locations owned by root
+    chmod 777 /tmp/parsec/e2e_tests
+    chmod 777 /tmp/
+    export SPIFFE_ENDPOINT_SOCKET="unix:///tmp/agent.sock"
+
+    # PATH is defined before each command for user to use their own version of the Rust toolchain
+    su -c "PATH=\"/home/parsec-client-1/.cargo/bin:${PATH}\";RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml --target-dir /home/parsec-client-1 all_providers::multitenancy::client1_before" parsec-client-1
+    su -c "PATH=\"/home/parsec-client-2/.cargo/bin:${PATH}\";RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml --target-dir /home/parsec-client-2 all_providers::multitenancy::client2" parsec-client-2
+    su -c "PATH=\"/home/parsec-client-1/.cargo/bin:${PATH}\";RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml --target-dir /home/parsec-client-1 all_providers::multitenancy::client1_after" parsec-client-1
```
ionut-arm, line 168: It's not what we've done in the past, but maybe the time has come to split this big `ci.sh` file into multiple shell scripts, one for each type of test that we want to run.

src/authenticators/jwt_svid_authenticator/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+use parsec_interface::operations::list_authenticators;
+use parsec_interface::requests::request::RequestAuth;
+use parsec_interface::requests::Result;
+use parsec_interface::requests::{AuthType, ResponseStatus};
+use parsec_interface::secrecy::ExposeSecret;
+use spiffe::svid::jwt::Jwt;
+use spiffe::workload::jwt::JWTClient;
+use std::str;
+
+/// JWT SVID authenticator
```
ionut-arm, line 18: Should this go into more detail about what the authenticator does (even if it's fairly straight forward code-wise)?

ionut-arm (commented):
Looks good!! Minor/somewhat unrelated comments below

Cargo.toml
hug-dev: Yes true, I forgot about that... I will try to get this upstreamed on crates.io before the next release and if not we can isolate it in a feature and not upstream that

ci.sh
hug-dev: True! Or re-haul the whole thing and build it in another language... Im afraid of long shell scripts

src/authenticators/jwt_svid_authenticator/mod.rs
hug-dev: I could! Although similarly than for the Unix Peer Credentials auth, the authenticator is explained in lengths in the book

ionut-arm: Yeah, I do find this double-documentation that we do around the service a bit weird

PR #284, opened 2020-11-10
Add TS provider

ionut-arm:
Create base of Trusted Service provider - it connects to the service
and can be instantiated in the service.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ picky-asn1-x509 = { version = "0.3.2", optional = true }
 users = "0.10.0"
 libc = "0.2.77"
 anyhow = "1.0.32"
+prost = { version = "0.6.1", optional = true }
 
 [dev-dependencies]
 rand = { version = "0.7.3", features = ["small_rng"] }
@@ -62,4 +63,4 @@ mbed-crypto-provider = ["psa-crypto"]
 pkcs11-provider = ["pkcs11", "picky-asn1-der", "picky-asn1", "picky-asn1-x509", "psa-crypto", "rand"]
 tpm-provider = ["tss-esapi", "picky-asn1-der", "picky-asn1", "picky-asn1-x509", "hex"]
 all-providers = ["tpm-provider", "pkcs11-provider", "mbed-crypto-provider"]
-trusted-service-provider = ["psa-crypto"]
+trusted-service-provider = ["psa-crypto", "prost"]
--- a/src/providers/trusted_service/context.rs
+++ b/src/providers/trusted_service/context.rs
@@ -1,0 +1,100 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use log::info;
+use std::ffi::{c_void, CString};
+use std::io::{Error, ErrorKind};
+use std::ptr::null_mut;
+use ts_binding::*;
+
+#[allow(
+    non_snake_case,
+    non_camel_case_types,
+    non_upper_case_globals,
+    clippy::unseparated_literal_suffix,
+    // There is an issue where long double become u128 in extern blocks. Check this issue:
+    // https://github.com/rust-lang/rust-bindgen/issues/1549
+    improper_ctypes,
+    missing_debug_implementations,
+    trivial_casts,
+    clippy::all,
+    unused,
+    unused_qualifications
+)]
+pub mod ts_binding {
+    include!(concat!(env!("OUT_DIR"), "/ts_bindings.rs"));
+}
+
+#[allow(
+    non_snake_case,
+    non_camel_case_types,
+    non_upper_case_globals,
+    clippy::unseparated_literal_suffix,
+    // There is an issue where long double become u128 in extern blocks. Check this issue:
+    // https://github.com/rust-lang/rust-bindgen/issues/1549
+    improper_ctypes,
+    missing_debug_implementations,
+    trivial_casts,
+    clippy::all,
+    unused,
+    unused_qualifications
+)]
+mod ts_protobuf {
+    include!(concat!(env!("OUT_DIR"), "/ts_crypto.rs"));
+}
+
+#[derive(Debug)]
+pub struct Context {
+    rpc_caller: *mut rpc_caller,
+    service_context: *mut service_context,
+    rpc_session_handle: *mut c_void,
+}
+
+impl Context {
+    pub fn connect() -> anyhow::Result<Self> {
+        info!("Querying for crypto Trusted Services");
+        let mut status = 0;
+        let service_context = unsafe {
+            service_locator_query(
+                CString::new("sn:tf.org:crypto:0").unwrap().into_raw(),
+                &mut status,
+            )
+        };
+        if service_context == null_mut() || status != 0 {
+            return Err(Error::new(
+                ErrorKind::Other,
+                format!(
+                    "Failed to connect to Trusted Service; status code: {}",
+                    status
+                ),
+            )
+            .into());
+        }
+
+        info!("Starting crypto Trusted Service context");
+        let mut rpc_caller = null_mut();
+        let rpc_session_handle = unsafe { service_context_open(service_context, &mut rpc_caller) };
+        if rpc_caller == null_mut() || rpc_session_handle == null_mut() {
+            return Err(
+                Error::new(ErrorKind::Other, "Failed to start Trusted Service context").into(),
+            );
+        }
+        let ctx = Context {
+            rpc_caller,
+            service_context,
+            rpc_session_handle,
+        };
+
+        Ok(ctx)
+    }
+}
+
+impl Drop for Context {
+    fn drop(&mut self) {
+        unsafe { service_context_close(self.service_context, self.rpc_session_handle) };
+
+        unsafe { service_locator_relinquish(self.service_context) };
+    }
+}
+
+unsafe impl Sync for Context {}
+unsafe impl Send for Context {}
--- a/src/providers/trusted_service/mod.rs
+++ b/src/providers/trusted_service/mod.rs
@@ -1,0 +1,107 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+//! Trusted Service provider
+//!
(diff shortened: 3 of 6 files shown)


Cargo.toml
```diff
@@ -45,6 +45,7 @@ picky-asn1-x509 = { version = "0.3.2", optional = true }
 users = "0.10.0"
 libc = "0.2.77"
 anyhow = "1.0.32"
+mbox = "0.5.0"
```
hug-dev, line 48: Maybe put it optional for the `trusted-service` feature?

src/providers/trusted_service/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+        };
+        Ok(ts_provider)
+    }
+}
+
+impl Provide for Provider {
+    fn describe(&self) -> Result<(ProviderInfo, HashSet<Opcode>)> {
+        Ok((ProviderInfo {
+            // Assigned UUID for this provider: 1c1139dc-ad7c-47dc-ad6b-db6fdb466552
+            uuid: Uuid::parse_str("1c1139dc-ad7c-47dc-ad6b-db6fdb466552").or(Err(ResponseStatus::InvalidEncoding))?,
```
hug-dev, line 64: You can directly write:
```rust
            uuid: Uuid::parse_str("1c1139dc-ad7c-47dc-ad6b-db6fdb466552")?,
```
because `ResponseStatus` implements [all](https://docs.rs/parsec-interface/0.21.0/parsec_interface/requests/enum.ResponseStatus.html#impl-From%3CError%3E-1) conversions 🌞

src/providers/trusted_service/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+            description: String::from("User space software provider, based on Mbed Crypto - the reference implementation of the PSA crypto API"),
+            vendor: String::from("Arm"),
+            version_maj: 0,
+            version_min: 1,
+            version_rev: 0,
+            id: ProviderID::MbedCrypto,
```
hug-dev, line 70: Copy 🍝 ?

src/providers/trusted_service/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+impl Provider {
+    /// Creates and initialise a new instance of Provider.
+    fn new(
+        key_info_store: Arc<RwLock<dyn ManageKeyInfo + Send + Sync>>,
+    ) -> anyhow::Result<Provider> {
+        let ts_provider = Provider {
+            key_info_store,
+            context: Context::connect()?,
+            key_attr_map: HashMap::new(),
+            id_counter: AtomicU32::new(key::PSA_KEY_ID_USER_MIN),
```
hug-dev, line 54: Does the Trusted Service actually implements PSA 1.0 then?

src/providers/trusted_service/context.rs
```diff
@@ -0,0 +1,117 @@
...
+    rpc_session_handle: Option<MBox<c_void>>,
+}
+
+impl Context {
+    pub fn connect() -> anyhow::Result<Self> {
+        info!("Querying for crypto Trusted Services");
+        let mut status = 0;
+        let service_context = unsafe {
+            service_locator_query(
+                CString::new("sn:tf.org:crypto:0").unwrap().into_raw(),
```
hug-dev, line 65: Will that string always stay the same? Does it need to be part of the configuration?

src/providers/trusted_service/context.rs
```diff
@@ -0,0 +1,117 @@
...
+                Error::new(ErrorKind::Other, "Failed to start Trusted Service context").into(),
+            );
+        }
+        let rpc_caller_box = unsafe { MBox::from_raw(rpc_caller) };
+        let service_context_box = unsafe { MBox::from_raw(service_context) };
+        let rpc_session_handle = unsafe { MBox::from_raw(rpc_session_handle) };
+        let ctx = Context {
+            rpc_caller: Some(*rpc_caller_box),
+            service_context: Some(*service_context_box),
+            rpc_session_handle: Some(rpc_session_handle),
```
hug-dev, line 94: When you dereference the `MBox` is that still keeping ownership? As in, is it not just copying the structure contents (which would be freed once this function ends)?

hug-dev (commented):
🏊 A few comments

Cargo.toml
ionut-arm: Will do

src/providers/trusted_service/mod.rs
ionut-arm: Yeh, I might have to check that one -_- I'm not exactly sure where the limits stand, might be an older version.

src/providers/trusted_service/context.rs
ionut-arm: Potentially part of the configuration, but I have no clue for now, tbh. It kinda looks static to me, but who knows

src/providers/trusted_service/context.rs
ionut-arm: Dereferencing copies the contents, I think, and keeps ownership of the initial object that came from the FFI. However, that object essentially contains some pointers. When the function ends, the `MBox` is dropped and, if I don't release them before, the pointers are dropped as well. But the FFI is the one who should be freeing them, ends up with a double free.

src/providers/trusted_service/context.rs
```diff
@@ -0,0 +1,117 @@
...
+    unused_qualifications
+)]
+pub mod ts_binding {
+    include!(concat!(env!("OUT_DIR"), "/ts_bindings.rs"));
+
+    unsafe impl Send for rpc_caller {}
+    unsafe impl Sync for rpc_caller {}
+
+    unsafe impl Send for service_context {}
+    unsafe impl Sync for service_context {}
```
ionut-arm, line 31: @hug-dev You might've missed this when reviewing - need to do this because otherwise I can't use the provider in a multithreaded environment. We do, however, need to lock down all operations using the `Context`, otherwise Bad Things Happen ™️ .

src/providers/trusted_service/context.rs
hug-dev: Ok we can check that one later.

src/providers/trusted_service/context.rs
hug-dev: ah I think I got it but why

> ends up with a double free.

?
If we copy all the contents of the `MBox` into owned structure then `MBox` will free the originals but it's fine because we made copies. For those copies, the FFI will free whatever needs to be freed and ownership will do the rest?

src/providers/trusted_service/context.rs
```diff
@@ -0,0 +1,117 @@
...
+        let mut service_context = self.service_context.take().unwrap();
+        let mut rpc_session_handle = self.rpc_session_handle.take().unwrap();
+        let mut _rpc_caller = self.rpc_caller.take().unwrap();
+
+        unsafe { service_context_close(&mut service_context, rpc_session_handle.as_mut_ptr()) };
+
+        unsafe { service_locator_relinquish(&mut service_context) };
+
+        std::mem::forget(service_context);
+        std::mem::forget(_rpc_caller);
```
hug-dev, line 114: Referring to my previous comment, why are those needed?

src/providers/trusted_service/context.rs
hug-dev: Is this `Context` going to be shared amoung threads? Are you going to put it in the provider?

Do the TS allow multi-threading? Could the `Context` itself implement `Send` and `Sync` so that we can put the multi-threading protections there?

src/providers/trusted_service/context.rs
ionut-arm: Same for why we need to remove the `MBox` - those two are essentially structures holding pointers - if they get dropped, you have double frees.

ionut-arm: Maybe I can put them in some sort of `WeakPointer` structure that doesn't drop them when the structure is dropped

src/providers/trusted_service/context.rs
ionut-arm: > Are you going to put it in the provider?

It already is in there: https://github.com/parallaxsecond/parsec/pull/284/files#diff-64b289133c05ec54716b4b88a821159125e8c48faf27c6df53e5a36a2f27596cR31

> Do the TS allow multi-threading?

The RPC caller isn't thread safe at the moment. We could try to get a new service context per thread/call, but I think that's not properly working yet either. So we'll have to lock somewhere in the stack.

> Could the Context itself implement Send and Sync so that we can put the multi-threading protections there?

I guess we could do that, and locking things inside the `Context` so the provider doesn't have to be aware of that.

src/providers/trusted_service/context.rs
ionut-arm: Yeah, I think `MBox` is clever and tries to free the pointer we give it. Decided to just drop `MBox` usage completely and store the pointers instead. I'll implement `Send` and `Sync` for `Context` and we can have it lock every call with a mutex.

src/providers/trusted_service/mod.rs
ionut-arm: The description as well 😆

src/providers/trusted_service/mod.rs
ionut-arm: The values seem to have stayed the same between the versions 🤔

src/providers/trusted_service/context.rs
ionut-arm: Didn't find anything in `std`, will just use `*mut`

src/providers/trusted_service/context.rs
```diff
@@ -0,0 +1,100 @@
...
+impl Drop for Context {
+    fn drop(&mut self) {
+        unsafe { service_context_close(self.service_context, self.rpc_session_handle) };
+
+        unsafe { service_locator_relinquish(self.service_context) };
+    }
+}
+
+unsafe impl Sync for Context {}
+unsafe impl Send for Context {}
```
hug-dev, line 100: nice, hide them at the end

hug-dev (approved):
Let's continue with it 🧑‍🔧

src/providers/trusted_service/context.rs
ionut-arm: 🙈

PR #285, opened 2020-11-16
Add generate and destroy key to TS provider

ionut-arm:
This commit adds functionality for generating and destroying
keys in the Trusted Service provider.


src/providers/trusted_service/context/key_management.rs
```diff
@@ -0,0 +1,146 @@
...
+
+impl Context {
+    pub fn generate_key(
+        &self,
+        key_attrs: Attributes,
+        id: u32,
+    ) -> parsec_interface::requests::Result<u32> {
+        let proto_req = Box::new(GenerateKeyIn {
+            attributes: Some(KeyAttributes {
+                r#type: u16::try_from(key_attrs.key_type)? as u32,
```
hug-dev, line 26: Can't you directly do `u32::try_from`?

src/providers/trusted_service/context/key_management.rs
```diff
@@ -0,0 +1,146 @@
...
+        });
+        let mut proto_resp = Box::new(GenerateKeyOut::default());
+        Context::send_request(
+            &proto_req,
+            &mut proto_resp,
+            Opcode::GenerateKey,
+            self.rpc_caller,
+        )?;
+
+        Ok(0)
```
hug-dev, line 44: What does the `u32` value returned mean?

src/providers/trusted_service/context/key_management.rs
```diff
@@ -0,0 +1,146 @@
...
+                } else {
+                    Err(e)
+                }
+            }
+        }
+    }
+
+    fn send_request<Req: Message, Resp: Message>(
+        req: &Box<Req>,
+        resp: &mut Box<Resp>,
```
hug-dev, line 104: is that the same as:
```
    fn send_request(
        req: &Box<Message>,
        resp: &mut Box<Message>,
```

hug-dev: and also maybe same as:
```
    fn send_request(
        req: &dyn Message,
        resp: &mut dyn Message,
```
why might be easier to read?

src/providers/trusted_service/context/key_management.rs
```diff
@@ -0,0 +1,146 @@
...
+    ) -> Result<(), Error> {
+        info!("Beginning call to Trusted Service");
+        let mut buf_out = null_mut();
+
+        let call_handle = unsafe { rpc_caller_begin(rpc_cl, &mut buf_out, req.encoded_len()) };
+        if call_handle == null_mut() {
+            error!("Call handle was null");
+            return Err(Error::CommunicationFailure);
+        }
+        let mut buf_out = unsafe { Vec::from_raw_parts(buf_out, 0, req.encoded_len()) };
```
hug-dev, line 116: Shouldn't length be equal to capacity?

src/providers/trusted_service/context/key_management.rs
```diff
@@ -0,0 +1,146 @@
...
+        req.encode(&mut buf_out).unwrap();
+        forget(buf_out);
```
hug-dev, line 118: hmm I don't understand this part: why do we create a vec and then forget it 😕 ?

hug-dev (commented):
Thanks for adding this! I don't fully understand how the context work, I've asked some initial questions

src/providers/trusted_service/mod.rs
```diff
@@ -3,22 +3,29 @@
 //! Trusted Service provider
 //!
 //! This provider is backed by a crypto Trusted Service deployed in TrustZone
+use super::mbed_crypto::key_management as mbed_key_management;
```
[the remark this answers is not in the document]
ionut-arm: Ok, makes sense!

src/providers/trusted_service/context/key_management.rs
ionut-arm: I don't think so - `key_type` is an enum, and we've implemented `TryFrom<Type> for psa_crypto_sys::psa_key_type_t`. I then have to go from `psa_key_type_t` to `u32`.

src/providers/trusted_service/context/key_management.rs
ionut-arm: Good spot, I think that's a remnant from the first iterations

src/providers/trusted_service/context/key_management.rs
hug-dev: ahh I see! I did not even look but we are going from a small to big type so it's all good not using `try_from`.

src/providers/trusted_service/context/key_management.rs
ionut-arm: The second one doesn't work because `dyn Message` isn't `Sized` so it won't compile.
The first one would look like this:
```rust
    fn send_request(
        req: &Box<dyn Message>,
        resp: &mut Box<dyn Message>,
```

But I'd then have to create the messages like this: `let proto_req: Box<dyn Message> = Box::from(CloseKeyIn { handle });` which means I can't access the fields inside anymore, only the methods defined on `Message`. I think the approach with `<Resp: Message, Req: Message>` just says "any type that implements `Message`", instead of "this particular type, `Box<dyn Message>`".

ionut-arm: Ah, the 2nd one works with `impl` instead of `dyn`!

src/providers/trusted_service/context/key_management.rs
ionut-arm: Same kinda thing that was happening with the context pointers - creating a `Vec` from a pointer so that I can use it on the line above to pull data into a protobuf object, then forgetting it because it's the RPC caller who ultimately owns it and frees it in `rpc_caller_end`.

src/providers/trusted_service/context/key_management.rs
ionut-arm: I'll try that again, but the weird thing is that when I tried, it was putting `req.encoded_len()` zeroes at the end, so the final length of the vector was actually twice what I wanted. 

It also doesn't help that the documentation only says about the 2nd field: `length needs to be less than or equal to capacity.`

src/providers/trusted_service/context/key_management.rs
hug-dev: Ah, so if I understand:

1. `rpc_caller_begin` allocates `req.encoded_len()` bytes at `buf_out`
2. we create a `Vec` over this location that we can pass to `req.encode`
3. `req.encode` writes `encoded_len()` bytes in the vector
4. We forget the `Vec` to prevent double-free

The RPC layer then expect the request to be there for the following operations and use it.

I think instead of creating a whole new `Vec` and forgetting it, we can just [create a slice from raw parts](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html), that way the memory management is not done by Rust and we can still (probably) pass it to `prost`.

src/providers/trusted_service/context/key_management.rs
```diff
@@ -0,0 +1,146 @@
...
+                &mut buf_in_size,
+            )
+        };
+        if status != 0 || opstatus != 0 {
+            unsafe { rpc_caller_end(rpc_cl, call_handle) };
+            error!("Error: status = {}, opstatus = {}", status, opstatus as i32);
+            return Status::from(opstatus as i32).to_result();
+        }
+        let buf_in = unsafe { Vec::from_raw_parts(buf_in, 0, buf_in_size) };
+        resp.merge(&*buf_in).unwrap();
```
hug-dev, line 140: ah ok here we write in `resp` the encoded content in `buf_in`. Could we return `resp` to make it clear this is the result?

ionut-arm: I'll try, but I'm not 100% sure it's possible

ionut-arm: Ok, made it work, it's actually nicer to work with and follows the "don't pass return arguments as params" guideline 😄

src/providers/trusted_service/context/key_management.rs
ionut-arm: Nice, cheers for the link - had to use the `_mut` one but nonetheless, works fine.

src/providers/trusted_service/context/key_management.rs
ionut-arm: Resolving this as moving to `slice` removed that issue

src/providers/trusted_service/mod.rs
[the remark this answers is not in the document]
ionut-arm: Are you ok with importing the key names directly in the module (not here, in the `key_management.rs` file)?

ionut-arm:
Fixed all the comments, and just to note - I did try this locally and the `create_and_destroy` test passed 👌

src/providers/trusted_service/mod.rs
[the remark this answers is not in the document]
hug-dev: Yes, it's ok!

hug-dev (approved):
Looks good, thanks for adressing the comments!

src/providers/trusted_service/context/key_management.rs
ionut-arm: Oh, sorry, thought I solved this, will update in a sec

PR #286, opened 2020-11-18
Add asymmetric sign and verify to TS provider

ionut-arm:
This commit adds functionality to the Trusted Service provider
allowing it to sign and verify hashes.

### DIFF
--- a/src/providers/trusted_service/asym_sign.rs
+++ b/src/providers/trusted_service/asym_sign.rs
@@ -1,0 +1,42 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use super::Provider;
+use crate::authenticators::ApplicationName;
+use crate::key_info_managers::KeyTriple;
+use crate::providers::mbed_crypto::key_management;
+use parsec_interface::operations::{psa_sign_hash, psa_verify_hash};
+use parsec_interface::requests::{ProviderID, Result};
+
+impl Provider {
+    pub(super) fn psa_sign_hash_internal(
+        &self,
+        app_name: ApplicationName,
+        op: psa_sign_hash::Operation,
+    ) -> Result<psa_sign_hash::Result> {
+        let key_triple = KeyTriple::new(app_name, ProviderID::TrustedService, op.key_name.clone());
+        let store_handle = self.key_info_store.read().expect("Key store lock poisoned");
+        let key_id = key_management::get_key_id(&key_triple, &*store_handle)?;
+
+        Ok(psa_sign_hash::Result {
+            signature: self
+                .context
+                .asym_sign(key_id, op.hash.to_vec(), op.alg)?
+                .into(),
+        })
+    }
+
+    pub(super) fn psa_verify_hash_internal(
+        &self,
+        app_name: ApplicationName,
+        op: psa_verify_hash::Operation,
+    ) -> Result<psa_verify_hash::Result> {
+        let key_triple = KeyTriple::new(app_name, ProviderID::TrustedService, op.key_name.clone());
+        let store_handle = self.key_info_store.read().expect("Key store lock poisoned");
+        let key_id = key_management::get_key_id(&key_triple, &*store_handle)?;
+
+        self.context
+            .asym_verify(key_id, op.hash.to_vec(), op.signature.to_vec(), op.alg)?;
+
+        Ok(psa_verify_hash::Result {})
+    }
+}
--- a/src/providers/trusted_service/context/asym_sign.rs
+++ b/src/providers/trusted_service/context/asym_sign.rs
@@ -1,0 +1,46 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use super::ts_protobuf::{SignHashIn, SignHashOut, VerifyHashIn};
+use super::Context;
+use log::info;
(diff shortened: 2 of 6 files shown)


src/providers/trusted_service/context/mod.rs
```diff
@@ -48,6 +49,59 @@ mod key_management;
 )]
 mod ts_protobuf {
     include!(concat!(env!("OUT_DIR"), "/ts_crypto.rs"));
+
+    pub trait OpcodeT {
```
ionut-arm, line 53: I've implemented this and the `SetHandle` below so that the whole "open key -> do operation -> close key" isn't done in each individual method (and also so that you don't have to pass the opcode along with the request - each request can only have one specific opcode anyway).

src/providers/mbed_crypto/key_management.rs
```diff
@@ -41,6 +41,18 @@ pub fn get_key_id(
     }
 }
 
+/// Get key Attributes from the Key Info Manager.
+pub fn get_key_attributes(
```
ionut-arm, line 45: Not sure why we didn't have this to begin with, I think we cache attributes separately in the Mbed Crypto provider

hug-dev: I think that is because Mbed Crypto handle the key attributes separately, we don't need to do it ourselves.

src/providers/trusted_service/context/asym_sign.rs
```diff
@@ -0,0 +1,46 @@
...
+        hash: Vec<u8>,
+        algorithm: AsymmetricSignature,
+    ) -> Result<Vec<u8>, ResponseStatus> {
+        info!("Handling GenerateKey request");
+        let mut proto_req = SignHashIn {
+            handle: 0,
+            hash,
+            alg: algorithm.try_into()?,
+        };
+        let SignHashOut { signature } = self.send_request_with_key(&mut proto_req, key_id)?;
```
hug-dev, line 23: Maybe the message can be given by ownership to `send_request_with_key` instead of reference? I guess you would not need to use the request structure afterwards.

hug-dev (commented):
Nice!! A few comments

src/providers/trusted_service/asym_sign.rs
```diff
@@ -0,0 +1,52 @@
...
+        op: psa_sign_hash::Operation,
+    ) -> Result<psa_sign_hash::Result> {
+        let key_triple = KeyTriple::new(app_name, ProviderID::TrustedService, op.key_name.clone());
+        let store_handle = self.key_info_store.read().expect("Key store lock poisoned");
+        let key_id = key_management::get_key_id(&key_triple, &*store_handle)?;
+
+        op.validate(key_management::get_key_attributes(
+            &key_triple,
+            &*store_handle,
+        )?)?;
```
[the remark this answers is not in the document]
ionut-arm: Ah, that's true, actually, but I don't know if I should rely on that

src/providers/trusted_service/context/mod.rs
```diff
@@ -152,14 +205,34 @@ impl Context {
...
+        let res = self.send_request(req);
+        let proto_req = CloseKeyIn { handle };
+        let res_close = self.send_request(&proto_req);
+        if res.is_err() {
+            res
+        } else if res_close.is_err() {
+            res_close
+        } else {
+            res
+        }
```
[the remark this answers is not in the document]
ionut-arm: yes, and to return the result of the main operation if both that and the close failed

src/providers/trusted_service/context/asym_sign.rs
ionut-arm: True, will change

src/providers/trusted_service/context/mod.rs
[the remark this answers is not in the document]
hug-dev: Cool, that is perfectly fine for me! Just so you know, I remember @samwell61 had to follow a similar approach and did something like:

```rust
res?;
res_close?;
res
```

src/providers/trusted_service/asym_sign.rs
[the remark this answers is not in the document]
hug-dev: yeah not entirely sure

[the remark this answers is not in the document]
ionut-arm: I think I will - if they claim to expose a PSA crypto API, they must enforce these conditions

ionut-arm:
Fixed according to all points raised. I also rebased on top of the latest TS version, so there are some changes to the build and `Context` connection

src/providers/trusted_service/context/mod.rs
```diff
@@ -134,39 +199,55 @@ impl Context {
...
+    ) -> Result<T, PsaError> {
+        let proto_req = OpenKeyIn { id: key_id };
+        let OpenKeyOut { handle } = self.send_request(&proto_req)?;
+        req.set_handle(handle);
+        let res = self.send_request(&req);
+        let proto_req = CloseKeyIn { handle };
+        let res_close = self.send_request(&proto_req);
+        let res = res?;
+        res_close?;
+        Ok(res)
```
hug-dev, line 242: 🙆‍♂️

hug-dev (approved):
I approve this code

src/providers/trusted_service/context/mod.rs
ionut-arm: 🆗 Google

PR #288, opened 2020-11-26
Add docs and some fixes

ionut-arm:
This commit adds documentation to the structures and methods in the new
Trusted Service provider. A few small fixes are also added, along with a
finer-grained approach to zeroizing the key data in import_key.

### DIFF
--- a/src/providers/trusted_service/context/asym_sign.rs
+++ b/src/providers/trusted_service/context/asym_sign.rs
@@ -8,13 +8,14 @@ use parsec_interface::requests::ResponseStatus;
 use std::convert::TryInto;
 
 impl Context {
-    pub fn asym_sign(
+    /// Sign a hash with an asymmetric key given its ID and the signing algorithm.
+    pub fn sign_hash(
         &self,
         key_id: u32,
         hash: Vec<u8>,
         algorithm: AsymmetricSignature,
     ) -> Result<Vec<u8>, ResponseStatus> {
-        info!("Handling GenerateKey request");
+        info!("Handling SignHash request");
         let proto_req = SignHashIn {
             handle: 0,
             hash,
@@ -25,14 +26,15 @@ impl Context {
         Ok(signature)
     }
 
-    pub fn asym_verify(
+    /// Verify a signature on a hash with an asymmetric key given its ID and the signing algorithm.
+    pub fn verify_hash(
         &self,
         key_id: u32,
         hash: Vec<u8>,
         signature: Vec<u8>,
         algorithm: AsymmetricSignature,
     ) -> Result<(), ResponseStatus> {
-        info!("Handling GenerateKey request");
+        info!("Handling VerifyHash request");
         let proto_req = VerifyHashIn {
             handle: 0,
             hash,
--- a/src/providers/trusted_service/context/key_management.rs
+++ b/src/providers/trusted_service/context/key_management.rs
@@ -13,6 +13,10 @@ use std::convert::{TryFrom, TryInto};
 use zeroize::Zeroize;
 
 impl Context {
+    /// Generate a key given its attributes and ID
+    ///
+    /// Lifetime flexibility is not supported: the `lifetime` parameter in the key
+    /// attributes is essentially ignored and replaced with `KeyLifetime::Persistent`.
     pub fn generate_key(&self, key_attrs: Attributes, id: u32) -> Result<(), ResponseStatus> {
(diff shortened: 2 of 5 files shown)


ionut-arm:
Not sure whether I should add code examples to `Context` - thoughts?

src/providers/trusted_service/context/mod.rs
```diff
@@ -171,11 +196,14 @@ impl Context {
...
 
+    // Serialize and send a request and deserialize the response back to
+    // the caller. The caller is responsible for explicitly declaring
+    // the response type if its contents are of interest.
     fn send_request<T: Message + Default>(
         &self,
         req: &(impl Message + GetOpcode),
     ) -> Result<T, PsaError> {
-        let _mutex_guard = self.call_mutex.try_lock().expect("Call mutex poisoned");
+        let _mutex_guard = self.call_mutex.lock().expect("Call mutex poisoned");
```
ionut-arm, line 206: Rookie mistake 👎

src/providers/trusted_service/context/asym_sign.rs
```diff
@@ -25,14 +26,15 @@ impl Context {
         Ok(signature)
     }
 
-    pub fn asym_verify(
```
hug-dev, line 28: Is the missing doc lint activated for this module?

src/providers/trusted_service/context/key_management.rs
```diff
@@ -35,38 +39,66 @@ impl Context {
...
+                        .permitted_algorithms
+                        .try_into()
+                        .map_err(|e| {
+                            data.zeroize();
+                            e
+                        })?,
                 }),
             }),
-            data: key_data.to_vec(),
+            data,
```
hug-dev, line 80: If you just transform it into a vector at this point then you don't need to zeroize it before when it fails right?

src/providers/trusted_service/context/mod.rs
hug-dev: Don't we want the thread to wait at this point that the mutex is free?

hug-dev (commented):
Nice docs ✍️

src/providers/trusted_service/context/mod.rs
ionut-arm: We do, but `try_lock` just returns an error if the lock is already taken. `lock` returns an error when it's poisoned

src/providers/trusted_service/context/asym_sign.rs
ionut-arm: I think not because it's not included in the feature list that we use with `clippy` (or even `cargo build`).

ionut-arm: Plus, I'm unsure whether it will complain about missing docs for things that are not public for the "library". This isn't public beyond the provider module

src/providers/trusted_service/context/key_management.rs
ionut-arm: I'm unsure what the compiler does - I don't think it necessarily does all those conversions in order. Was hoping that constructing `data` before would make it do that for sure, and hopefully all the conversions happen after that

src/providers/trusted_service/context/asym_sign.rs
hug-dev: Ahhh gotcha

src/providers/trusted_service/context/key_management.rs
hug-dev: Ah yes maybe. In that case my other suggestion would be to wrap the vector into a `Zeroizing` so that it's cleared when dropped wihtout having to zeroize on every error

src/providers/trusted_service/context/mod.rs
hug-dev: ahh yes right `lock` is the right one

src/providers/trusted_service/context/key_management.rs
ionut-arm: That's not really helping, though, because then I have to "convert" from that `Zeroizing` into a `Vec` :] the thing I'm populating here is a protobuf type, I don't have a choice as to what I fill it with.

hug-dev: ah ok fair enough then

hug-dev (approved):
Seal of Approval -> 🐳

PR #296, opened 2020-11-30
Pull in TS code and build on CI

ionut-arm:
This commit imports the TS code as a submodule and uses it to
build and test the provider on the CI.

### DIFF
--- a/e2e_tests/provider_cfg/trusted-service/Dockerfile
+++ b/e2e_tests/provider_cfg/trusted-service/Dockerfile
@@ -8,14 +8,19 @@ RUN apt-get update && \
 	apt-get install -y clang libclang-dev && \
 	# Needed for Open SSL
 	apt-get install -y pkg-config libssl-dev python3-distutils && \
-	apt-get install -y python3-distutils python3-protobuf nanopb
+	# Needed for Protobuf handling
+	apt-get install -y python3-distutils python3-protobuf nanopb && \
+	# Needed for the Trusted Services implementation
+	apt-get install -y libc++-dev libc++abi-dev 
 
 WORKDIR /tmp
-COPY trusted-services /tmp/trusted-services
-RUN cd trusted-services/deployments/ts-lib/posix-x86/ \
+RUN git clone https://git.trustedfirmware.org/TS/trusted-services.git --branch integration
+RUN cd trusted-services/deployments/libts/linux-pc/ \
 	&& cmake . \
 	&& make \
-	&& cp libts-lib.a nanopb_install/lib/libprotobuf-nanopb.a mbedcrypto_install/lib/libmbedcrypto.a /usr/local/lib/
+	&& cp libts.so nanopb_install/lib/libprotobuf-nanopb.a mbedcrypto_install/lib/libmbedcrypto.a /usr/local/lib/
+
+ENV LD_LIBRARY_PATH="/usr/local/lib"
 
 # Install Rust toolchain
 RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
--- a/e2e_tests/src/stress.rs
+++ b/e2e_tests/src/stress.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 use super::TestClient;
 use log::info;
-use parsec_client::core::interface::requests::ResponseStatus;
+use parsec_client::core::interface::requests::{Opcode, ResponseStatus};
 use rand::Rng;
 use rand::{
     distributions::{Alphanumeric, Distribution, Standard},
@@ -96,7 +96,7 @@ struct StressTestWorker {
     config: StressTestConfig,
     sign_key_name: String,
     ecc_key_name: Option<String>,
-    encrypt_key_name: String,
+    encrypt_key_name: Option<String>,
     client: TestClient,
 }
 
@@ -108,6 +108,7 @@ impl StressTestWorker {
         let auth = generate_string(10);
         info!("Worker with auth `{}` starting.", auth);
         client.set_default_auth(Some(auth));
+        let opcodes = client.list_opcodes(client.provider()).unwrap();
 
         // Create sign/verify key
         let sign_key_name = generate_string(10);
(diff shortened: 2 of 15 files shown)


e2e_tests/provider_cfg/trusted-service/Dockerfile
```diff
@@ -8,14 +8,19 @@ RUN apt-get update && \
...
-	apt-get install -y python3-distutils python3-protobuf nanopb
+	# Needed for Protobuf handling
+	apt-get install -y python3-distutils python3-protobuf nanopb && \
+	# Needed for the Trusted Services implementation
+	apt-get install -y libc++-dev libc++abi-dev 
 
 WORKDIR /tmp
-COPY trusted-services /tmp/trusted-services
-RUN cd trusted-services/deployments/ts-lib/posix-x86/ \
+RUN git clone https://git.trustedfirmware.org/TS/trusted-services.git --branch integration
```
hug-dev, line 17: Is it needed if the code is already included in Parsec as a submodule?

src/providers/trusted_service/context/error.rs
```diff
@@ -0,0 +1,203 @@
...
+use log::error;
+use psa_crypto::types::status::{Error as PsaError, Status as PsaStatus};
+use std::error::Error as ErrorTrait;
+use std::fmt;
+
+/// Wrapper over types of error that the TS Context might return
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub enum Error {
+    /// errors coming from the Crypto TS service, associated with performed operations.
+    PsaCryptoError(PsaError),
```
hug-dev, line 12: I would say the `Error` suffix is not needed for the variants 😋

src/providers/trusted_service/context/error.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+}
+
+impl From<std::convert::Infallible> for Error {
+    fn from(_: std::convert::Infallible) -> Self {
+        unimplemented!()
+    }
+}
+
+impl From<std::num::TryFromIntError> for Error {
```
hug-dev, line 61: 👏

src/providers/trusted_service/context/mod.rs
```diff
@@ -238,19 +241,15 @@ impl Context {
                 &mut resp_buf_size,
             )
         };
-        if status != 0 || opstatus != 0 {
+        Error::from_status_opstatus(status, opstatus).map_err(|e| {
```
hug-dev, line 244: Do you need the `map_err` anymore?

src/providers/trusted_service/error.rs
```diff
@@ -0,0 +1,42 @@
...
+            RpcCallerError::EndpointDoesNotExist
+            | RpcCallerError::InvalidOpcode
+            | RpcCallerError::SerializationNotSupported
+            | RpcCallerError::ResourceFailure
+            | RpcCallerError::NotReady
+            | RpcCallerError::InvalidTransaction
+            | RpcCallerError::Internal => ResponseStatus::PsaErrorCommunicationFailure,
+            RpcCallerError::InvalidRequestBody
+            | RpcCallerError::InvalidResponseBody
+            | RpcCallerError::InvalidParameter => ResponseStatus::PsaErrorInvalidArgument,
```
hug-dev, line 18: Couldn't these also be converted to `CommunicationFailure`? As in if they happen, isn't it a problem in Parsec rather than a wrong argument sent by the user?

hug-dev (commented):
Nice 👏 Good to have everything together, a few minor comments

e2e_tests/provider_cfg/trusted-service/Dockerfile
ionut-arm: Well, then it would have to be part of the CI script - I'd rather it's here, no big loss.

ionut-arm: Actually, I guess I could put these steps in a different script or in commands in the YAML CI file

src/providers/trusted_service/context/mod.rs
ionut-arm: Yes, because I need to close the call if it's not successful

src/providers/trusted_service/error.rs
ionut-arm: No, that was something that I was seeing in the test I was fixing - well, maybe I can move `InvalidResponseBody` and `InvalidParameter`, though I'm not sure what exactly the RPC client considers as "bad parameter". In that test, it was a "malformed" hash buffer that the RPC client didn't like, so it wasn't our fault.

src/providers/trusted_service/context/mod.rs
hug-dev: oh yes missed that, I read `|e| e` as closure

src/providers/trusted_service/error.rs
hug-dev: > a "malformed" hash buffer that the RPC client didn't like

malformed for PSA Crypto as well? As in if it is legal for the PSA Crypto interface but not for the RPC client maybe we should not return `InvalidArgument` and this is a bug in the client?

ionut-arm: Maybe, I'll look into it

src/providers/trusted_service/context/error.rs
ionut-arm: Fixed locally - will upstream in a bit

src/providers/trusted_service/error.rs
ionut-arm: ah, I see what's happening - in that test we rely on the fact that a hash of 255 bytes won't be accepted as valid (which makes sense, a call to PSA crypto directly returns `InvalidArgument`), but the RPC client that talks to the TS needs to deserialize the call and has a limit of how large the hash can be. If the hash is longer than 64 bytes, it returns "invalid request body" so that one probably makes sense to convert to `PsaErrorInvalidArgument`. I'll move `RpcCallerError::ResponseBody` and `RpcCallerError::InvalidParameter` to `CommunicationError`.

e2e_tests/src/stress.rs
```diff
@@ -132,9 +133,22 @@ impl StressTestWorker {
 
         // Create asym encrypt/decrypt key
         let encrypt_key_name = generate_string(10);
-        client
-            .generate_rsa_encryption_keys_rsapkcs1v15crypt(encrypt_key_name.clone())
-            .expect("Failed to generate asym encr key");
+        let res = client.generate_rsa_encryption_keys_rsapkcs1v15crypt(encrypt_key_name.clone());
+
+        if !(res.is_ok() || res == Err(ResponseStatus::PsaErrorNotSupported)) {
```
hug-dev, line 138: image

hug-dev (approved):
Looks good, thank you!!

🍕

e2e_tests/src/stress.rs
ionut-arm: 👌

PR #298, opened 2020-12-09
Improve error handling in builder

ionut-arm:
This commit changes the way errors are handled in the service builder(s)
to always fail if an error is encountered.


hug-dev (approved):
Looks good!

PR #299, opened 2020-12-09
Switch Travis CI build to cron-only

ionut-arm:
This commit changes the settings for Travis CI builds to only allow cron
jobs to run (as we now have a very limited number of build minutes
available) and to show them as failed instead of ignoring all failures.


hug-dev (approved):
Looks good! Saving every minute!

PR #303, opened 2020-12-17
Add new parsec provider using ATECCx08 cryptochip via CryptoAuthentication Library

RobertDrazkowskiGL:
Definition of new structures and their interfaces.
### Version mismatch of psa-crypto
There is a version mismatch in a dependent library (psa-crypto) between parsec-interface and parsec on appropriate master/HEAD versions:
- parsec-interface depends on git https://github.com/parallaxsecond/rust-psa-crypto#rev18dd4dda96d8b61d2e112b9e6ad83e90fe768d78
- parsec depends on depends on official version 6.0.0 from crates.io

The solution is to

- assign a new version to psa-crypto (git/hash) as discussed elsewhere
- update the package meta-data on crates.io
- use this new version in both parsec-interface and parsec

As a workaround this PR modifies parsec/Cargo.toml to have the same psa-crypto as a parsec-interface.

### Wrong parsec-interface version number being used
The parsec-interface extension (add an CryptoAuthLib entry to providers' enumeration) was not followed by package version increase nor the package meta-data on crates.io was updated. This PR addressed an issue with parsec-interface version for parsec compilation purposes but there are still issues with e2e-tests.
#### Issues with e2e-tests
The e2e-tests depend on parsec-client that depends on parsec-interface and it is affected by the previously mentioned versioning issues. 
 
Summary TODO points 
- [x] assign a new version to psa-crypto and update crate's meta-data on crates.io
- [x] update both parsec and parsec-interface to use this new psa-crypto version
- [x] assign a new version to parsec-interface and update its metadata on crates.io
- [x] update parsec and parsec-client to use this new parsec-interface

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -131,3 +131,7 @@ key_info_manager = "on-disk-manager"
 # e.g. "str:password", or to represent a string version of a hex value, e.g. "hex:1a2b3c". If no prefix is
 # provided, the value is considered to be a string.
 #owner_hierarchy_auth = "password"
+
+[[provider]]
+provider_type = "CryptoAuthLib"
+key_info_manager = "on-disk-manager"
--- a/config.toml
+++ b/config.toml
@@ -132,6 +132,25 @@ key_info_manager = "on-disk-manager"
 # provided, the value is considered to be a string.
 #owner_hierarchy_auth = "password"
 
-[[provider]]
-provider_type = "CryptoAuthLib"
-key_info_manager = "on-disk-manager"
+# Example of a CryptoAuthLib provider configuration
+# All below parameters depend on what devices, interfaces or parameters are required or supported by 
+# "rust-cryptoauthlib" wrapper for cryptoauthlib and underlying hardware.
+#[[provider]]
+#provider_type = "CryptoAuthLib"
+#key_info_manager = "on-disk-manager"
+# (Required) ATCA device type.
+#   Supported values: "atecc508a", "atecc608a"
+#device_type = "atecc508a"
+# (Required) Interface for ATCA device
+#   Supported values: "i2c"
+#iface_type = "i2c"
+# (Required) Default wake delay for ATCA device
+#wake_delay = 1500
+# (Required) Default number of rx retries for ATCA device
+#rx_retries = 20
+# (Optional - required for i2c) i2c slave addres
+#slave_address = 0xc0
+# (Optional - required for i2c) i2c bus number
+#bus = 1
+# (Optional - required for i2c) i2c bus baud rate
+#baud = 400000
--- a/config.toml
+++ b/config.toml
@@ -138,19 +138,3 @@ key_info_manager = "on-disk-manager"
 #[[provider]]
 #provider_type = "CryptoAuthLib"
 #key_info_manager = "on-disk-manager"
-# (Required) ATCA device type.
-#   Supported values: "atecc508a", "atecc608a"
-#device_type = "atecc508a"
-# (Required) Interface for ATCA device
-#   Supported values: "i2c"
-#iface_type = "i2c"
-# (Required) Default wake delay for ATCA device
-#wake_delay = 1500
-# (Required) Default number of rx retries for ATCA device
-#rx_retries = 20
-# (Optional - required for i2c) i2c slave addres
-#slave_address = 0xc0
-# (Optional - required for i2c) i2c bus number
-#bus = 1
-# (Optional - required for i2c) i2c bus baud rate
-#baud = 400000
--- a/e2e_tests/provider_cfg/cryptoauthlib/Dockerfile
+++ b/e2e_tests/provider_cfg/cryptoauthlib/Dockerfile
@@ -1,0 +1,15 @@
+FROM ubuntu:18.04
+
+ENV PKG_CONFIG_PATH /usr/local/lib/pkgconfig
+
+RUN apt-get update && \
+	apt-get install -y git make gcc python3 python curl wget cmake && \
+	apt-get install -y automake autoconf libtool pkg-config libssl-dev && \
+	# These libraries are needed for bindgen as it uses libclang.so
+	apt-get install -y clang libclang-dev libc6-dev-i386 && \
+	# Install cargo globally to not have to install it for each user for multitenancy tests
+	apt-get install -y cargo
+
+# Add users for multitenancy tests
+RUN useradd -m parsec-client-1
+RUN useradd -m parsec-client-2
--- a/src/providers/cryptoauthlib/mod.rs
+++ b/src/providers/cryptoauthlib/mod.rs
@@ -1,0 +1,89 @@
+// Copyright 2019 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+//! Microchip CryptoAuthentication Library provider
+//!
+//! This provider is a hardware based implementation of PSA Crypto, Mbed Crypto.
+use super::Provide;
+use crate::key_info_managers::ManageKeyInfo;
+use derivative::Derivative;
+use log::trace;
+use std::collections::HashSet;
+use std::io::{Error, ErrorKind};
+use std::sync::{Arc, RwLock};
+use uuid::Uuid;
+
+use parsec_interface::operations::list_providers::ProviderInfo;
+use parsec_interface::requests::{Opcode, ProviderID, ResponseStatus, Result};
+
+const SUPPORTED_OPCODES: [Opcode; 0] = [];
+
+/// CryptoAuthLib provider structure
+#[derive(Derivative)]
+#[derivative(Debug, Clone, Copy)]
+pub struct Provider {
+    // device: rust_cryptoauthlib::AtcaDevice,
+}
+
+impl Provider {
+    /// Creates and initialise a new instance of CryptoAuthLibProvider
+    fn new(_key_info_store: Arc<RwLock<dyn ManageKeyInfo + Send + Sync>>) -> Option<Provider> {
+        Some(Provider {})
+    }
+}
+
+impl Provide for Provider {
+    fn describe(&self) -> Result<(ProviderInfo, HashSet<Opcode>)> {
+        trace!("describe ingress");
+        Ok((ProviderInfo {
+            // Assigned UUID for this provider: b8ba81e2-e9f7-4bdd-b096-a29d0019960c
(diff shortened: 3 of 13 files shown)


ionut-arm (commented):
I only had a brief look, but it seems good overall, left just one comment below

config.toml
```diff
@@ -131,3 +131,7 @@ key_info_manager = "on-disk-manager"
 # e.g. "str:password", or to represent a string version of a hex value, e.g. "hex:1a2b3c". If no prefix is
 # provided, the value is considered to be a string.
 #owner_hierarchy_auth = "password"
+
+[[provider]]
+provider_type = "CryptoAuthLib"
+key_info_manager = "on-disk-manager"
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: Thank you. I will do it.

paulhowardarm:
Thanks for raising the PR - it's really exciting to see this stuff starting to arrive! And this is a perfectly-sized first change as well - just introducing the provider without its functions.

Before we can merge this in, I think we'll need to work a bit more on getting the dependency graph consistent. The CI actions are failing in this PR, and I think it's because there is a lack of consistency between how `parsec-interface` is being consumed into `parsec` and `e2e_tests` respectively. It's worth being aware that `e2e_tests` is an independent crate, even though its source code is nested within `parsec`. Like `parsec`, `e2e_tests` needs to consume `parsec-interface`, and it does so in a slightly disguised way, via its resurfacing through `parsec-client-rust`.

Confusing, I know!

The good news is that I think we can resolve the confusion quite easily by following some basic process. I think we need to create tagged releases of everything in the dependency chain, and then consume those releases in the `Cargo.toml` files instead of having specific Git references.

If we merge this PR as it is, it will cause the CI to fail, so it's worth sorting out ahead of time.

Starting from the ground up, and taking into account our Slack conversation as well as this PR. My proposal would be as follows, although I would also like to take advice on this from other engineers in the community, since I haven't been involved with the tagging/releasing process to date, and I want to make sure that what I'm proposing here is consistent with established practice.

1. Create a new tag/release on `psa-crypto` and set it as `0.6.1` since its functionality doesn't seem to have evolved significantly since the last tag.
2. Change `parsec-interface-rs` (via separate PR) to consume `psa-crypto` at `0.6.1` rather than its current git/hash reference.
3. Create a new tag/release on `parsec-interface-rs` and set it as `0.22.0`.
4. Change `parsec-client-rust` (via separate PR) to consume `parsec-interface-rs` at its new tag level of `0.22.0`.
5. Create a new tag/release on `parsec-client-rust` and set it as `0.12.0`.
6. Update this PR with a new commit to consume `psa-crypto` at `0.6.1`, `parsec-interface-rs` at `0.22.0` and `parsec-client-rust` at `0.12.0`.

This should result in a cleanly-updated and consistent dependency graph, with everything consuming official versions rather than referencing commit hashes or head revisions. Referencing head revisions is a particular problem because it could mean that commits to other repos could actually randomly break things.

RobertDrazkowskiGL:
Yes, I fully agree, I was writing very similar proposal in PR top comment.
Is crates.io automatically updated after package version numbers are increased on github?

paulhowardarm:
I'm not sure how the tagging process links with crates.io. Let me put out a query on the community channel and we'll find out. Ideally the process would be written up somewhere.

paulhowardarm (changes requested):
As discussed in the PR and on the Slack channel, we will need to sanitise the dependency graph a little before this can be merged. We are working on a process for this, but it will involve a few PRs and releases in other repos, so please be patient while these propagate through.

RobertDrazkowskiGL:
I have just pushed a newer version with updated dependencies version numbers as well as corrected config.toml contents as requested by Ionuț.

src/providers/cryptoauthlib/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+impl Provide for Provider {
+    fn describe(&self) -> Result<(ProviderInfo, HashSet<Opcode>)> {
+        trace!("describe ingress");
+        Ok((ProviderInfo {
+            // Assigned UUID for this provider: b8ba81e2-e9f7-4bdd-b096-a29d0019960c
+            uuid: Uuid::parse_str("b8ba81e2-e9f7-4bdd-b096-a29d0019960c").or(Err(ResponseStatus::InvalidEncoding))?,
+            description: String::from("User space hardware provider, utilizing MicrochipTech CryptoAuthentication Library for ATECCx08 chips"),
+            vendor: String::from("Arm"),
```
hug-dev, line 41: Maybe the vendor could be `"Microchip Technology "`?

config.toml
```diff
@@ -131,3 +131,26 @@ key_info_manager = "on-disk-manager"
...
+# (Required) Default wake delay for ATCA device
+#wake_delay = 1500
+# (Required) Default number of rx retries for ATCA device
+#rx_retries = 20
+# (Optional - required for i2c) i2c slave addres
+#slave_address = 0xc0
+# (Optional - required for i2c) i2c bus number
+#bus = 1
+# (Optional - required for i2c) i2c bus baud rate
+#baud = 400000
```
hug-dev, line 156: Since for this PR only the `key_info_manager` field is required, you could remove that for now and add it later on!

e2e_tests/provider_cfg/cryptoauthlib/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM ubuntu:18.04
```
hug-dev, line 1: Thanks for adding the testing infrastructure for this provider as well!
I guess you will add a new CI workflow for this provider in the `.github/workflows/ci.yml` file later on? I don't think it's needed to do now, since most test will fail and the building part is handled by the "all-providers" test.

hug-dev (commented):
Thanks for the PR 🚀!

Added a few comments, overall this looks good to me to start with!
You might have to fix some lints on the CI to make it work...

About the issues you had, I guess they would be fixed by now? The `e2e-tests` crate use a `git` dependency on the Rust client (at `master`) which now contains the newest interface.

src/providers/cryptoauthlib/mod.rs
```diff
@@ -0,0 +1,89 @@
+// Copyright 2019 Contributors to the Parsec project.
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: Updated.

src/providers/cryptoauthlib/mod.rs
RobertDrazkowskiGL: As already mentioned on slack - this is a political decision at the moment it is safe to use "Arm" as a vendor.

config.toml
RobertDrazkowskiGL: Removed

e2e_tests/provider_cfg/cryptoauthlib/Dockerfile
RobertDrazkowskiGL: Yes, at the moment there is just a very basic structure. We will work on enabling tests later on.

src/providers/cryptoauthlib/mod.rs
hug-dev: Ok agree!

src/providers/cryptoauthlib/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+impl Provider {
+    /// Creates and initialise a new instance of CryptoAuthLibProvider
+    fn new(_key_info_store: Arc<RwLock<dyn ManageKeyInfo + Send + Sync>>) -> Option<Provider> {
+        Some(Provider {})
+    }
+}
+
+impl Provide for Provider {
```
ionut-arm, line 34: It looks like the `list_keys` test is failing because it attempts to list the keys for all providers in the service - thus, including yours because you've added it in the configuration - and that tries to call `list_keys` on this provider, which fails because it uses the default. I think you can just implement it to always return an empty list for now

RobertDrazkowskiGL: Thank you for the hint. I can reproduce this (it just failed for me the same way) so I can proceed.

e2e_tests/provider_cfg/cryptoauthlib/Dockerfile
RobertDrazkowskiGL: Unfortunately my comment was very wrong. If the CI tests are to be enabled for this particular provider, there must be underlying ATECC hardware available, which is not an issue for our team, but I can imaging it is a challenge on a public github. 
Is it possible on github at all?

ionut-arm: So I take it there's no way to have a simulator, or even a mock library that returns preconfigured responses? We could run a different set of tests for this particular provider in that case

hug-dev: Yes that would be a problem. Currently all of our tests are using virtualized environment: virtual TPM server and mock PKCS11 library for example. I was hoping that there would be something similar for either a virtualised ATECCx08 device that the CryptoAuth lib can communicate to or a mock CryptoAuth lib implementation that would work standalone for testing.
A quick Google did not show me anything useful :/

If it's not possible, maybe for now we can limit ourselves to local end-to-end testing (but still do build and lint tests on the CI). Of course this is not a viable option in the long run amd we will need to think on how to support this.

RobertDrazkowskiGL: Thank you for quick response.
To be honest - the cryptochip is not being accessed directly but through rust-cryptoauthlib, and there might be implemented an additional "device" that uses a software implementation (based on e.g. mbed-tls). But this is an additional, not planned effort and its purpose would be limited to testing only.

ionut-arm: Sure, that makes sense - we could merge this now and raise an issue to look into testing later, even if it's just for sanity checking. 

I think it makes sense to simply not include that provider in E2E tests, but even just unit tests - if you can manage to get some to run, maybe with mocking - would help.

RobertDrazkowskiGL:
@paulhowardarm 
You have requested a change, which I believe has already been addressed. 
The only 'github' dependency is for "parsec-client' where github is updated but crates.io points on previous version.
I hope this is not an issue.

hug-dev (approved):
Thanks!! Looks good to me 👍 

> The only 'github' dependency is for "parsec-client' where github is updated but crates.io points on previous version. I hope this is not an issue.

To me it's fine as it's only used in the `e2e_tests` crate!

paulhowardarm (approved):

PR #307, opened 2021-01-06
Disable Travis builds

ionut-arm:


hug-dev (approved):
Let's do that for now since we do not really use it anyway... Will change the settings to remove the checks as well

PR #313, opened 2021-01-11
Add a test checking ListKeys provider target

### LINKED ISSUE #310: ListKeys should only be callable on the Core provider
`ListKeys` is a core operation and should not be callable on crypto providers.
Reference: [this](https://github.com/parallaxsecond/parsec/issues/293#issuecomment-756680153), also this [on Slack](https://cloud-native.slack.com/archives/C01EARH2ZB3/p1610102057182000).

A test need to be added to make sure that `PsaErrorNotSupported` is returned when this operation is called on crypto providers.

hug-dev:
Fix #310

### DIFF
--- a/e2e_tests/tests/all_providers/normal.rs
+++ b/e2e_tests/tests/all_providers/normal.rs
@@ -1,8 +1,12 @@
 // Copyright 2019 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
+use e2e_tests::RawRequestClient;
 use e2e_tests::TestClient;
 use parsec_client::core::interface::operations::list_providers::Uuid;
-use parsec_client::core::interface::requests::{AuthType, Opcode, ProviderID, Result};
+use parsec_client::core::interface::requests::request::RawHeader;
+use parsec_client::core::interface::requests::{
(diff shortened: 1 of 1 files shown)


e2e_tests/tests/all_providers/normal.rs
```diff
@@ -147,3 +151,19 @@ fn list_keys() {
...
 }
+
+#[test]
+// See #310
+fn invalid_provider_list_keys() {
+    let mut client = RawRequestClient {};
+    let mut req_hdr = RawHeader::new();
+
+    // Always targeting the Mbed Crypto provider
+    req_hdr.provider = 0x1;
```
ionut-arm, line 162: Do you think we should try and loop here as well, to test all the providers?

ionut-arm (approved):
💯

e2e_tests/tests/all_providers/normal.rs
hug-dev: Ideally yes I guess. The thing is that it might be a bit over the top for this test to first spawn a normal test client to get all provider IDs and then use the raw client to test all provider IDs.

PR #316, opened 2021-01-12
Add admin configuration

### LINKED ISSUE #308: Implement admin logic
## In the config file

We add under the `[authenticator]` section the `admins` field:
```
admins = [ { name = "admin_1" }, { name = "admin_2" } ]
```
The type of `name` is authenticator-dependant but would map to application names that are considered as admin for that authenticator: a SPIFFE ID, a UID, a string (for direct auth).

## In the service builder

When building the authenticator structure (implementing `Authenticate`), we store the information of the admins application name in the authenticator structure.

## In the front-end handler

The front-end handler calls the `authenticate` method on the authenticator to then dispatch the request with the good application name (or none). Things here stay the same but the `ApplicationName` structure would be enhanced with a boolean field indicating if it is an `admin` application name or not (`is_admin`). It's the authenticator during the `authenticate` call that would set that boolean or not if the application name is in the list of the admin ones.

## In the back-end handler

Before calling the appropriate `Provide` method, for those that require admin privilege, `is_admin` is called to check if the application has privilege or not. If not a `AdminOperation` error should be returned (see parallaxsecond/parsec-interface-rs#93).

## Testing

Will be done in another issue, in #309

ionut-arm:
This commit adds support for authenticators to identify admin
applications. The admin app names can be configured from the config.toml
file and lead to the ApplicationName having a flag set indicating
whether the app is an admin or not.

Fixes #308

### DIFF
--- a/src/authenticators/direct_authenticator/mod.rs
+++ b/src/authenticators/direct_authenticator/mod.rs
@@ -8,8 +8,7 @@
 //! This authenticator does not offer any security value and should only be used in environments
 //! where all the clients and the service are mutually trustworthy.
 
-use super::ApplicationName;
-use super::Authenticate;
+use super::{Admin, AdminList, ApplicationName, Authenticate};
 use crate::front::listener::ConnectionMetadata;
 use log::error;
 use parsec_interface::operations::list_authenticators;
@@ -20,8 +19,19 @@ use parsec_interface::secrecy::ExposeSecret;
 use std::str;
 
 /// Direct authentication authenticator implementation
-#[derive(Copy, Clone, Debug)]
-pub struct DirectAuthenticator;
+#[derive(Clone, Debug)]
+pub struct DirectAuthenticator {
+    admins: AdminList,
+}
+
+impl DirectAuthenticator {
+    /// Create new direct authenticator
+    pub fn new(admins: Vec<Admin>) -> Self {
+        DirectAuthenticator {
+            admins: admins.into(),
+        }
+    }
+}
 
 impl Authenticate for DirectAuthenticator {
     fn describe(&self) -> Result<list_authenticators::AuthenticatorInfo> {
@@ -47,7 +57,11 @@ impl Authenticate for DirectAuthenticator {
             Err(ResponseStatus::AuthenticationError)
         } else {
             match str::from_utf8(auth.buffer.expose_secret()) {
-                Ok(str) => Ok(ApplicationName(String::from(str))),
+                Ok(str) => {
+                    let app_name = String::from(str);
+                    let is_admin = self.admins.is_admin(&app_name);
+                    Ok(ApplicationName::new(app_name, is_admin))
+                }
                 Err(_) => {
                     error!("Error parsing the authentication value as a UTF-8 string.");
                     Err(ResponseStatus::AuthenticationError)
@@ -59,14 +73,16 @@ impl Authenticate for DirectAuthenticator {
 
 #[cfg(test)]
 mod test {
-    use super::super::Authenticate;
+    use super::super::{Admin, Authenticate};
     use super::DirectAuthenticator;
     use parsec_interface::requests::request::RequestAuth;
     use parsec_interface::requests::ResponseStatus;
 
     #[test]
     fn successful_authentication() {
-        let authenticator = DirectAuthenticator {};
+        let authenticator = DirectAuthenticator {
+            admins: Default::default(),
+        };
 
         let app_name = "app_name".to_string();
         let req_auth = RequestAuth::new(app_name.clone().into_bytes());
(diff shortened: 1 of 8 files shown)


src/authenticators/mod.rs
```diff
@@ -24,7 +24,10 @@ use zeroize::Zeroize;
 
 /// String wrapper for app names
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
-pub struct ApplicationName(String);
+pub struct ApplicationName {
```
ionut-arm, line 27: Wasn't sure if I should just change this to `Application`

src/authenticators/mod.rs
```diff
@@ -75,12 +86,35 @@ impl std::fmt::Display for ApplicationName {
...
+
+/// Structure defining the properties of a service admin
+#[derive(Deserialize, Debug, Zeroize, Clone)]
+#[zeroize(drop)]
+pub struct Admin {
+    name: String,
+}
+
+impl Admin {
+    fn check_app_name(&self, app_name: &mut ApplicationName) {
```
ionut-arm, line 115: Similarly - wasn't sure what name to give to this, but it felt better to keep the place where the actual check is done and `set_admin` is called contained, so it's easier to verify

ionut-arm:
Also, wasn't sure if I should add the checks in the backend or if you'd rather do it when implementing the operations that actually need them

src/authenticators/direct_authenticator/mod.rs
```diff
@@ -20,8 +19,24 @@ use parsec_interface::secrecy::ExposeSecret;
...
+}
+
+impl DirectAuthenticator {
+    /// Create new direct authenticator
+    pub fn new(admins: Vec<Admin>) -> Self {
+        DirectAuthenticator { admins }
+    }
+
+    /// Verify if the app is an admin
+    fn admin_check(&self, app_name: &mut ApplicationName) {
```
hug-dev, line 34: I would prefer renaming this method to understand better that it modifies `app_name`, what about `set_admin_status`? And similarly for the one on `Admin`

src/authenticators/direct_authenticator/mod.rs
```diff
@@ -92,12 +116,43 @@ mod test {
...
         let conn_metadata = None;
         let status = authenticator
             .authenticate(&RequestAuth::new(Vec::new()), conn_metadata)
             .expect_err("Empty auth should have failed");
 
         assert_eq!(status, ResponseStatus::AuthenticationError);
     }
+
+    #[test]
+    fn admin_check() {
```
hug-dev, line 131: Nice tests 👌

src/authenticators/mod.rs
hug-dev: I think for this PR this is fine keeping the name to not trigger too many changes

src/authenticators/mod.rs
```diff
@@ -54,18 +57,26 @@ pub trait Authenticate {
...
     }
 
     /// Get a reference to the inner string
     pub fn get_name(&self) -> &str {
-        &self.0
+        &self.name
+    }
+
+    /// Flag current application as being an admin
+    fn set_admin(&mut self) {
```
hug-dev, line 72: What about this method taking a boolean to set the field? I feel like `set_admin(true)` is nice to understand

src/utils/service_builder.rs
```diff
@@ -398,16 +398,27 @@ fn build_authenticators(config: &AuthenticatorConfig) -> Vec<(AuthType, Authenti
...
 
     match config {
-        AuthenticatorConfig::Direct => {
-            authenticators.push((AuthType::Direct, Box::from(DirectAuthenticator {})))
-        }
-        AuthenticatorConfig::UnixPeerCredentials => authenticators.push((
+        AuthenticatorConfig::Direct { admins } => authenticators.push((
+            AuthType::Direct,
+            Box::from(DirectAuthenticator::new(
+                admins.as_ref().cloned().unwrap_or_default(),
```
hug-dev, line 404: Will the `Default` yield to an empty vector?

hug-dev (commented):
Thanks looks good! I agree with the general algorithm, just had some questions about method naming to maybe understand it better.

> Also, wasn't sure if I should add the checks in the backend or if you'd rather do it when implementing the operations that actually need them

I guess we can do it when adding support for the first admin operations.

src/utils/service_builder.rs
ionut-arm: Yes, that's the default for `Vec`

src/authenticators/mod.rs
ionut-arm: I think I'll change my approach here and just keep `ApplicationName` immutable, though that means I'll have to check whether a string is an admin name before calling the constructor.

hug-dev (approved):
It looks better like that indeed!

Just one nitpick but feel free to ignore:

```rust
Ok(ApplicationName::new(
    app_name.clone(),
    self.admins.is_admin(&app_name),
))
```
could that be changed to:
```rust
let is_admin = self.admins.is_admin(&app_name);
Ok(ApplicationName::new(
    app_name,
    is_admin,
))
```
in order to avoid the `clone`?

ionut-arm:
Good point - was aware of the "tradeoff", but didn't think of performance, though if you do a lot of operations, it might matter

hug-dev (approved):

Issue #312, opened 2021-01-08
Resolve default implementation issue for `list_keys` in `Provide`


ionut-arm:
For most of the methods on the `Provide` trait we offer a default implementation that generally just returns `PsaErrorNotSupported`. This allows developers of providers to skip the implementation of those methods in particular if they're not supported for their backend or if they haven't done the implementation work yet.

However, providing these defaults doesn't always work as expected - the `list_keys` operation, for example, is service-wide so one provider not supporting it would currently break the operation for the whole service.

There are two possible solutions:
* remove the default implementation for `list_keys`
* keep the implementation, but allow the core provider to simply skip when it sees `PsaErrorNotSupported` as the returned error

Neither of those really fixes the problem - developers should really implement that functionality at some point and there's no way for us to know if they've done that currently. So I also propose updating the `all_providers::normal::list_keys` test to loop through all providers returned by `list_providers` and to generate a key for each one that supports `generate_key` (and maybe try `import` if `generate` isn't supported?). Then we can check that they were all created.

ionut-arm:
Note that this should also be relevant to other operations that work on a service-aggregate level, such as those in #311

ionut-arm:
Also implement similar changes for `list_clients` and `delete_client`

PR #318, opened 2021-01-15
Add ListClients and DeleteClient operations

### LINKED ISSUE #309: Add a test for admin operations
After `ListClients` is implemented and the admin features (#308), this should be tested.

We could use the multitenancy test and use one of the two clients as an admin. The admin one must execute `ListClients` successfully while the other one should fail with the appropriate error. This for all authenticators supported. The infrasctruture is almost already there for that, expect that ListClients will have to be implemented first.

### LINKED ISSUE #311: Implement ListClients and DeleteClient in the core provider
Implemented similarly as other operations (before the provider level), but they would require admin privilege.

`Provide` trait method:
```
    fn list_clients(
        &self,
        op: list_clients::Operation,
    ) -> Result<list_clients::Result> {
```
In the future this would also contain as parameter the authentication type to only return clients of that type.

The operation would be implemented in the **core provider**.

For `ListClients` to do something similar than `ListKeys`: have the `list_clients` method be implemented by all providers to use the key info manager implementation they are using to return a list of client. The Core provider would call `list_clients` on all of its `Provide` implementations (which it already has) and merge them.
`ListClients` should only be callable on the Core Provider, see #310 .

For `DeleteClient`, we can use a mix of existing methods: `list_keys` on the application name given as parameter and then `psa_destroy_key` on each one of them.

# Testing

We can test this in the all-providers multitenancy tests. Making sure that the admin can see all clients and that he can delete all keys of the other one 💯

hug-dev:
Also adds multitenancy tests for those operations and for the admin
feature. Splits the ApplicationName structure into Application and
ApplicationName to only deal with key names in the Key Info Managers and
not the admin status which is not relevent there.

Fix #309 
Fix #311

### DIFF
--- a/e2e_tests/provider_cfg/all/config.toml
+++ b/e2e_tests/provider_cfg/all/config.toml
@@ -14,6 +14,7 @@ socket_path = "/tmp/parsec.sock"
 
 [authenticator]
 auth_type = "Direct"
+admins = [ { name = "list_clients test" }, { name = "1000" }, { name = "client1" }, { name = "spiffe://example.org/parsec-client-1" } ]
 #workload_endpoint="unix:///tmp/agent.sock"
 
 [[key_manager]]
--- a/e2e_tests/tests/all_providers/multitenancy.rs
+++ b/e2e_tests/tests/all_providers/multitenancy.rs
@@ -9,6 +9,8 @@ use parsec_client::core::interface::requests::{ProviderID, ResponseStatus};
 // 3. client1_after is executed as parsec-client-1
 //
 // They are executed against all possible authenticators in Parsec.
+//
+// client1 will be configured as an admin.
 
 #[test]
 fn client1_before() {
@@ -23,11 +25,15 @@ fn client1_before() {
         client.set_provider(*provider);
         client.generate_rsa_sign_key(key.clone()).unwrap();
     }
+
+    let clients = client.list_clients().unwrap();
+    assert_eq!(clients.len(), 1);
 }
 
 #[test]
 fn client2() {
     let mut client = TestClient::new();
+    client.do_not_destroy_keys();
     client.set_default_auth(Some("client2".to_string()));
 
     let key = String::from("multitenant");
@@ -49,11 +55,24 @@ fn client2() {
         client.generate_rsa_sign_key(key.clone()).unwrap();
         client.destroy_key(key.clone()).unwrap();
     }
+
+    assert_eq!(
+        client.list_clients().unwrap_err(),
+        ResponseStatus::AdminOperation
+    );
+    assert_eq!(
+        client.delete_client("toto".to_string()).unwrap_err(),
+        ResponseStatus::AdminOperation
+    );
+    client
+        .generate_rsa_sign_key("client2-key".to_string())
+        .unwrap();
 }
 
 #[test]
 fn client1_after() {
     let mut client = TestClient::new();
+    client.do_not_destroy_keys();
     client.set_default_auth(Some("client1".to_string()));
 
     // Verify all keys are still there and can be used
@@ -66,4 +85,18 @@ fn client1_after() {
         client.set_provider(*provider);
         client.destroy_key(key.clone()).unwrap();
     }
+
+    client
+        .generate_rsa_sign_key("client1-key".to_string())
+        .unwrap();
+    let mut clients = client.list_clients().unwrap();
+    assert_eq!(clients.len(), 2);
+    client.delete_client(clients.remove(0)).unwrap();
+    let mut clients = client.list_clients().unwrap();
+    assert_eq!(clients.len(), 1);
+    client.delete_client(clients.remove(0)).unwrap();
+    let clients = client.list_clients().unwrap();
+    assert_eq!(clients.len(), 0);
+    let keys = client.list_keys().unwrap();
+    assert_eq!(keys.len(), 0);
 }
--- a/src/authenticators/mod.rs
+++ b/src/authenticators/mod.rs
@@ -27,6 +27,20 @@ use zeroize::Zeroize;
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct ApplicationName {
     name: String,
+}
+
+impl Deref for ApplicationName {
+    type Target = String;
+
+    fn deref(&self) -> &Self::Target {
+        &self.name
+    }
+}
+
+/// Wrapper for a Parsec application
+#[derive(Debug, Clone)]
+pub struct Application {
+    name: ApplicationName,
     is_admin: bool,
 }
 
@@ -40,7 +54,7 @@ pub trait Authenticate {
     /// operation.
     fn describe(&self) -> Result<list_authenticators::AuthenticatorInfo>;
 
-    /// Authenticates a `RequestAuth` payload and returns the `ApplicationName` if successful. A
+    /// Authenticates a `RequestAuth` payload and returns the `Application` if successful. A
     /// optional `ConnectionMetadata` object is passed in too, since it is sometimes possible to
     /// perform authentication based on the connection's metadata (i.e. as is the case for UNIX
     /// domain sockets with Unix peer credentials).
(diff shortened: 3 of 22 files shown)


e2e_tests/provider_cfg/all/config.toml
```diff
@@ -14,6 +14,7 @@ socket_path = "/tmp/parsec.sock"
 
 [authenticator]
 auth_type = "Direct"
+admins = [ { name = "list_clients test" }, { name = "parsec-client-1" }, { name = "client1" }, { name = "spiffe://example.org/parsec-client-1" } ]
```
hug-dev, line 17: Having a list of admin names is definitely useful 👌

src/back/backend_handler.rs
```diff
@@ -215,35 +220,39 @@ impl BackEndHandler {
...
+                self.result_to_response(NativeResult::ListClients(result), header)
+            }
+            NativeOperation::DeleteClient(op_delete_client) => {
+                let result = unwrap_or_else_return!(self.provider.delete_client(op_delete_client));
+                trace!("delete_client egress");
+                self.result_to_response(NativeResult::DeleteClient(result), header)
+            }
             NativeOperation::PsaHashCompute(op_hash_compute) => {
-                let _app_name =
-                    unwrap_or_else_return!(app_name.ok_or(ResponseStatus::NotAuthenticated));
```
hug-dev, line 227: This is not needed I think?

src/providers/mod.rs
```diff
@@ -155,6 +156,18 @@ pub trait Provide {
...
+    fn list_clients(&self, _op: list_clients::Operation) -> Result<list_clients::Result> {
+        trace!("list_clients ingress");
+        Err(ResponseStatus::PsaErrorNotSupported)
+    }
+
+    /// Delete all data a client has in the service..
+    fn delete_client(&self, _op: delete_client::Operation) -> Result<delete_client::Result> {
+        trace!("delete_client ingress");
+        Err(ResponseStatus::PsaErrorNotSupported)
+    }
```
ionut-arm, line 169: As per #312 you should either not implement this or skip returning on `PsaErrorNotSupported` in `CoreProvider`, but I think you do neither

src/providers/pkcs11/mod.rs
```diff
@@ -225,6 +225,20 @@ impl Provide for Provider {
...
+    fn list_clients(&self, _op: list_clients::Operation) -> Result<list_clients::Result> {
+        let store_handle = self.key_info_store.read().expect("Key store lock poisoned");
+        Ok(list_clients::Result {
+            clients: key_info_managers::list_clients(store_handle.deref(), ProviderID::MbedCrypto)
+                .map_err(|e| {
+                    format_error!("Error occurred when fetching key information", e);
+                    ResponseStatus::KeyInfoManagerError
+                })?
+                .into_iter()
+                .map(|e| e.to_string())
```
ionut-arm, line 237: Does `e` mean `client`?

e2e_tests/tests/all_providers/multitenancy.rs
```diff
@@ -66,4 +85,18 @@ fn client1_after() {
...
+    let mut clients = client.list_clients().unwrap();
+    assert_eq!(clients.len(), 2);
+    client.delete_client(clients.remove(0)).unwrap();
+    let mut clients = client.list_clients().unwrap();
+    assert_eq!(clients.len(), 1);
+    client.delete_client(clients.remove(0)).unwrap();
+    let clients = client.list_clients().unwrap();
+    assert_eq!(clients.len(), 0);
+    let keys = client.list_keys().unwrap();
+    assert_eq!(keys.len(), 0);
```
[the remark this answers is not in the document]
hug-dev: It can not directly be done here for all authentication methods, it would involve adding one more function `client2_after` to be executed under the other clients and one more line in the `ci.sh` file. Can do it if you think it's worth it?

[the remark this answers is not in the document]
ionut-arm: Nah, leave it as is, I think there are other tests for stuff like this, right?

src/authenticators/mod.rs
```diff
@@ -52,32 +58,45 @@ pub trait Authenticate {
...
+        }
+    }
 
     /// Check whether the application is an admin
     pub fn is_admin(&self) -> bool {
         self.is_admin
     }
+
+    /// Get a reference to the inner ApplicationName string
+    pub fn get_name(&self) -> &str {
```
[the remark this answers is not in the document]
hug-dev: I did that at first but then had `.get_name().get_name()` everywhere so figured I could just do that?

src/key_info_managers/mod.rs
```diff
@@ -163,7 +163,7 @@ pub fn list_keys(
     let key_triples = manager.get_all(provider_id)?;
 
     for key_triple in key_triples {
-        if key_triple.app_name() != app_name {
+        if key_triple.app_name().get_name() != app_name.get_name() {
```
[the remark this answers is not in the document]
hug-dev: No, shouldn't be! Will revert

src/providers/core/mod.rs
```diff
@@ -93,6 +94,45 @@ impl Provide for Provider {
...
+    fn list_clients(&self, _op: list_clients::Operation) -> Result<list_clients::Result> {
+        trace!("list_clients ingress");
+
+        let mut clients: Vec<String> = Vec::new();
+        for provider in &self.prov_list {
+            let mut result = provider.list_clients(_op)?;
+            clients.append(&mut result.clients);
+        }
+        clients.sort();
+        clients.dedup();
```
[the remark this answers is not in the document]
hug-dev: Cause I would have to convert each `Vec` to a `HashSet` and then convert the `HashSet` back to a `Vec`. If there is a way to represent a set in protobuf maybe we should use that everywhere?

src/authenticators/mod.rs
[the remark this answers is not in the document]
ionut-arm: You could implement `Deref` on `ApplicationName` - it is just a "smart" wrapper around a type, after all

src/providers/core/mod.rs
```diff
@@ -93,6 +94,45 @@ impl Provide for Provider {
...
+                    ApplicationName::from_name(client.clone()),
+                    list_keys::Operation {},
+                )?
+                .keys;
+            for key in keys {
+                let key_name = key.name;
+                let _ = provider.psa_destroy_key(
+                    ApplicationName::from_name(client.clone()),
+                    psa_destroy_key::Operation { key_name },
+                )?;
```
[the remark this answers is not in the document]
hug-dev: I am not sure we thought about that. I guess we will have to take a decision and maybe put it in the books?

src/providers/mod.rs
hug-dev: If I don't do this, won't we have the same problem as explained in #312 then? As in I was thinking to do the same (bad) thing than `list_keys` and whatever solution is chosen for #312 would fix both of those things.

src/providers/pkcs11/mod.rs
hug-dev: I always use `e` as Element but could also be cliEnt, right :shipit: ? I can change to `client`

hug-dev: `app_name` is more correct actually

src/providers/mod.rs
ionut-arm: Ok, that's fair, will add a comment about it in that issue

src/providers/pkcs11/mod.rs
ionut-arm: Yeah, I was confused because I tend to use `e` only when it's an `error` :joy:

src/providers/core/mod.rs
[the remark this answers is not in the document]
ionut-arm: Will be discussed tomorrow in the weekly meeting, can leave it as is for now.

ionut-arm (approved):
:+1:

hug-dev:
> Will be discussed tomorrow in the weekly meeting, can leave it as is for now.

As in, shall we merge this for now and apply what we discuss on the meeting on Parsec and the documentation afterwards? Or shall we wait?

ionut-arm:
> 
> 
> > Will be discussed tomorrow in the weekly meeting, can leave it as is for now.
> 
> As in, shall we merge this for now and apply what we discuss on the meeting on Parsec and the documentation afterwards? Or shall we wait?

Merge this one, we can apply the changes afterwards in a separate PR

PR #321, opened 2021-01-19
Improve mandatory Provide methods

### LINKED ISSUE #312: Resolve default implementation issue for `list_keys` in `Provide`

ionut-arm:
This commit changes the way some methods on the Provide trait are
implemented. Given that in some cases the methods _must_ be implemented
by the providers (such as `list_keys`), there should be no default
implementation for them and the tests must verify that they operate
correctly.

Fixes #312

### DIFF
--- a/e2e_tests/tests/all_providers/normal.rs
+++ b/e2e_tests/tests/all_providers/normal.rs
@@ -128,16 +128,19 @@ fn list_keys() {
 
     assert!(keys.is_empty());
 
-    let key1 = String::from("list_keys1");
-    let key2 = String::from("list_keys2");
-    let key3 = String::from("list_keys3");
+    let providers = client.list_providers().expect("Failed to list providers");
+    let mut suitable_providers = vec![];
 
-    client.set_provider(ProviderID::MbedCrypto);
-    client.generate_rsa_sign_key(key1.clone()).unwrap();
-    client.set_provider(ProviderID::Pkcs11);
-    client.generate_rsa_sign_key(key2.clone()).unwrap();
-    client.set_provider(ProviderID::Tpm);
-    client.generate_rsa_sign_key(key3.clone()).unwrap();
+    for provider in providers.iter() {
+        client.set_provider(provider.id);
+        if !client.is_operation_supported(Opcode::PsaGenerateKey) {
+            continue;
+        }
+        suitable_providers.push(provider.clone());
+        client
+            .generate_rsa_sign_key(format!("list_keys_{}", provider.id))
+            .unwrap();
+    }
 
     let key_names: Vec<(String, ProviderID)> = client
         .list_keys()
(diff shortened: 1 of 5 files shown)


e2e_tests/tests/all_providers/normal.rs
```diff
@@ -128,16 +128,19 @@ fn list_keys() {
...
-    let key2 = String::from("list_keys2");
-    let key3 = String::from("list_keys3");
-
-    client.set_provider(ProviderID::MbedCrypto);
-    client.generate_rsa_sign_key(key1.clone()).unwrap();
-    client.set_provider(ProviderID::Pkcs11);
-    client.generate_rsa_sign_key(key2.clone()).unwrap();
-    client.set_provider(ProviderID::Tpm);
-    client.generate_rsa_sign_key(key3.clone()).unwrap();
+    let providers = client.list_providers().expect("Failed to list providers");
```
hug-dev, line 131: That's awesome!
This snippet shows the power of Parsec actually

e2e_tests/tests/all_providers/normal.rs
```diff
@@ -187,27 +191,47 @@ fn invalid_provider_list_clients() {
...
+    }
+
+    client.set_default_auth(Some(all_providers_user.clone()));
 
     let clients = client.list_clients().expect("list_clients failed");
-    assert!(clients.contains(&"list_clients test".to_string()));
-    client
-        .delete_client("list_clients test".to_string())
-        .unwrap();
+    assert_eq!(clients.len(), suitable_providers.len() + 1);
```
hug-dev, line 225: Might there be concurrency problems if that test is executed with other ones creating keys at the same time with other auth values?
If so, maybe we can check that `clients` contains (at least) exactly what we expect.
edit: oh it's actually just what you are doing in the next lines, do we need this `assert` line 225 then?

hug-dev (approved):
Nice! And thank you for improving the tests as well.

Should that close #312 ?

e2e_tests/tests/all_providers/normal.rs
ionut-arm: Yeah, I was thinking that we need to check we have the exact set that we're expecting but yeah, it makes sense to only look for the ones we created in this test. Will remove this assert.

hug-dev (approved):

PR #322, opened 2021-01-19
Make the authenticators their own features

puiterwijk:
This would allow not compiling in certain authenticators.

### DIFF
--- a/src/authenticators/mod.rs
+++ b/src/authenticators/mod.rs
@@ -9,10 +9,20 @@
 //! is the `RequestAuth` field of a request, which is parsed by the authenticator specified in the header.
 //! The authentication functionality is abstracted through an `Authenticate` trait.
 
+#[cfg(not(any(
+    feature = "direct-authenticator",
+    feature = "unix-peer-credentials-authenticator",
+    feature = "jwt-svid-authenticator",
+)))]
+compile_error!("Please provide in at least one authenticator");
+
+#[cfg(feature = "direct-authenticator")]
 pub mod direct_authenticator;
 
+#[cfg(feature = "unix-peer-credentials-authenticator")]
 pub mod unix_peer_credentials_authenticator;
 
+#[cfg(feature = "jwt-svid-authenticator")]
 pub mod jwt_svid_authenticator;
 
 use crate::front::listener::ConnectionMetadata;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ users = "0.11.0"
 libc = "0.2.77"
 anyhow = "1.0.32"
 # Using a fork until the JWT support is merged into the main rust-spiffe repository
-spiffe = { git = "https://github.com/hug-dev/rust-spiffe", branch = "refactor-jwt" }
+spiffe = { git = "https://github.com/hug-dev/rust-spiffe", branch = "refactor-jwt", optional = true }
 #rust-cryptoauthlib = { git = "https://github.com/RobertDrazkowskiGL/rust-cryptoauthlib.git", optional = true }
 rust-cryptoauthlib = { version = "0.1.0", optional = true }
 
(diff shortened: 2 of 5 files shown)


src/authenticators/mod.rs
```diff
@@ -149,6 +160,14 @@ impl Admin {
...
 
 impl AdminList {
+    #[cfg_attr(
+        not(any(
+            feature = "direct-authenticator",
+            feature = "unix-peer-credentials-authenticator",
+            feature = "jwt-svid-authenticator",
+        )),
+        allow(unused)
+    )]
```
ionut-arm, line 170: I'm curious - why do them like this, adding `allow(unused)`, instead of `cfg(any(...))`?

hug-dev:
Thanks a lot for adding this 💯 Having the unix peer cred auth by default is actually really good for security I think.

This is obviously good to have a script that checks for compilation of each independant feature in a systematic way. We could also do something simpler by just adding in our `ci.sh` in the "Build test" section, some `cargo build` lines:
```
cargo build --no-default-features --features "direct-authenticator"
```
and etc for all of them. I was thinking that we were already compiling the provider feature independantly for each provider tests.

src/authenticators/mod.rs
puiterwijk: Because otherwise, if you don't compile in any authenticators, it warns (and thus fails to compile) with a message that doesn't scream "You forgot to enable an authenticator" to me.
So this is basically making sure that it compiles and then just doesn't ever work on runtime, instead of the case where it fails to compile with "weird" compile errors.
(Which one to choose was something I was doubting over for a while... I'm still convinced that I'd want something like the C `#error` to make the compilation fail with a self-set error.)

puiterwijk:
> Thanks a lot for adding this 💯 Having the unix peer cred auth by default is actually really good for security I think.
> 
> This is obviously good to have a script that checks for compilation of each independant feature in a systematic way. We could also do something simpler by just adding in our `ci.sh` in the "Build test" section, some `cargo build` lines:
> 
> ```
> cargo build --no-default-features --features "direct-authenticator"
> ```
> 
> and etc for all of them. I was thinking that we were already compiling the provider feature independantly for each provider tests.

Yes. The different provider tests only enable their own providers.

src/authenticators/mod.rs
puiterwijk: And just after I mentioned that, I searched for the correct things and found https://doc.rust-lang.org/std/macro.compile_error.html

ionut-arm (approved):
👍🏻 Nice, thanks for the change! Looks like `rustfmt` wasn't too happy

crosscompile.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+# Copyright 2019 Contributors to the Parsec project.
```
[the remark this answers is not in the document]
puiterwijk: I am so bad with years! Thanks.

hug-dev (commented):
Just one question about the new script file otherwise the rest looks good!

crosscompile.sh
```diff
@@ -0,0 +1,12 @@
...
+# Copyright 2021 Contributors to the Parsec project.
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+cargo install cargo-hack
+
+# --feature-powerset instead of --each-feature would be possible, but would be very long
+# (by now, 128 jobs (2^7) when excluding all-authenticators and all-providers)
+cargo hack --each-feature --exclude-no-default-features build
```
hug-dev, line 12: Could it be possible to add those two lines directly on the `docker run` line? Like:
```diff
-docker run -v $(pwd):/tmp/parsec -w /tmp/parsec all-providers /tmp/parsec/crosscompile.sh
+docker run -v $(pwd):/tmp/parsec -w /tmp/parsec all-providers cargo install cargo-hack && cargo hack --each-feature --exclude-no-default-features build
```
or add them as a new command of the `ci.sh` script?

I am just a bit cautious about having too many testing scripts at the root, for clarity. But maybe it's fine!

hug-dev (approved):
Nice 👌

PR #324, opened 2021-01-21
Add consistency in key creation/deletion

### LINKED ISSUE #149: Make existence of key info consistent with existence of key
Currently we have to check, for each return path within provider methods, whether we have maintained consistency between the existence of the key and the existence of the key info stored in the manager. This applies to key generation, import and destruction.

My idea was that at the beginning of these methods we create (for generate/import) or retrieve (for destroy) the key mapping and feed it to a function that does the rest of the processing (all the param checks, calling the method from the interface underneath). The inner method then returns its `Result`, and based on that, before returning from the main provider method, we either keep or destroy the mapping.

### LINKED ISSUE #319: Review error handling in key destruction
When the Parsec service reports that a key destruction is successful, all is well and we have a common outcome across the providers - the key is gone from both the backend and all Parsec caches. On the other hand, when an error occurs the exact sequence of events is somewhat inconsistent. This issue is only meant for investigating the possible paths that could be taken due to errors, and where the state of the service ends up from there. If changes are required (either to the service or to the documentation of `PsaDestroyKey`), a new issue should be created.

An example of discrepancy:
* in the TPM provider we store encrypted keys in the KIM, so the backend does not have any traces of the key. Thus, all we have to do is delete the KIM entry. So we rely on the KIM to do everything for us, there's nothing else to delete, therefore consistency isn't an issue.
* in the PKCS11 provider, for asymmetric keys, we get identifiers for 2 objects and attempt to remove them sequentially. If the first operation succeeds but the 2nd fails, we're stuck, because (1) we leave our KIM entry intact (since we failed), so it appears we still have the object; (2) if we try to remove it again after the failure, now the first part will fail because the object isn't there anymore. This is on top of the fact that we don't really know whether an error code coming from PKCS11 means the object is still there or not (it might not be defined by the spec)

### LINKED ISSUE #323: Adjust implementation of DeleteClient
During our weekly call we decided `DeleteClient` should be interpreted as "make a best effort to remove all resources belonging to client X or to unlink those resources if the backend fails to remove them". This is in contrast with our previous expectation which saw `DeleteClient` as a sum of `Destroy<Resource>`, e.g. `PsaDestroyKey`, operations and which currently fails at the first "sub-operation" failure. 

The new semantics should be that all providers must do a best-effort job of removing all references and metadata to resources belonging to client X, even when removing the resource returns an error code from the backend. The only case where an error response code should be returned to the Parsec client is when a failure occurs during metadata cleanup, and thus a new service client with the same identity would potentially be able to access some old resources (or have conflicts when creating new ones with identical names).

The contract and interface will remain the same (although we will probably need a new response code specific for these errors), but the implementation will change. We also need to make sure the documentation for this operation is consistent with these semantics and sufficiently clear.

hug-dev:
on all providers.
Also make the core operations looping on all providers "best-effort" by
being infallible.

This is mainly for initial review/CI testing. This makes the assumption that [this comment](https://github.com/parallaxsecond/parsec/issues/323#issuecomment-764854039) is agreed but I can change later on.

Fix #323 
Fix #319 
Fix #149

### DIFF
--- a/src/key_info_managers/mod.rs
+++ b/src/key_info_managers/mod.rs
@@ -146,60 +146,82 @@ pub trait ManageKeyInfo {
     fn exists(&self, key_triple: &KeyTriple) -> Result<bool, String>;
 }
 
-/// Returns a Vec of the KeyInfo objects corresponding to the given application name and
-/// provider ID.
-///
-/// # Errors
-///
-/// Returns an error as a String if there was a problem accessing the Key Info Manager.
-pub fn list_keys(
-    manager: &dyn ManageKeyInfo,
-    app_name: &ApplicationName,
-    provider_id: ProviderID,
-) -> Result<Vec<parsec_interface::operations::list_keys::KeyInfo>, String> {
-    use parsec_interface::operations::list_keys::KeyInfo;
+// "+ Send + Sync" is needed for things like:
+// ```
+// let store_handle = self.key_info_store.read().expect("Key store lock poisoned");
+// let _ = store_handle.list_keys(&app_name, ProviderID::Pkcs11)?;
+// ```
+// to work because `store_handle` is a RWLockReadGuard of (dyn ManageKeyInfo + Send + Sync +
+// 'static). Did not work with only `impl dyn ManageKeyInfo` below. Maybe there is a way to
+// implement automagically the same methods on (dyn ManageKeyInfo + Send + Sync + 'static)
+// if they are implemented on dyn ManageKeyInfo but not sure.
+impl dyn ManageKeyInfo + Send + Sync {
+    /// Returns a Vec of the KeyInfo objects corresponding to the given application name and
+    /// provider ID.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error as a String if there was a problem accessing the Key Info Manager.
+    pub fn list_keys(
+        &self,
+        app_name: &ApplicationName,
+        provider_id: ProviderID,
+    ) -> Result<Vec<parsec_interface::operations::list_keys::KeyInfo>, String> {
+        use parsec_interface::operations::list_keys::KeyInfo;
 
-    let mut keys: Vec<KeyInfo> = Vec::new();
-    let key_triples = manager.get_all(provider_id)?;
+        let mut keys: Vec<KeyInfo> = Vec::new();
+        let key_triples = self.get_all(provider_id)?;
 
-    for key_triple in key_triples {
-        if key_triple.app_name() != app_name {
-            continue;
+        for key_triple in key_triples {
+            if key_triple.app_name() != app_name {
+                continue;
+            }
+
+            let key_info = self.get(key_triple)?;
+            let key_info = match key_info {
+                Some(key_info) => key_info,
+                _ => continue,
+            };
+
+            keys.push(KeyInfo {
+                provider_id: key_triple.provider_id,
+                name: key_triple.key_name().to_string(),
+                attributes: key_info.attributes,
+            });
         }
 
-        let key_info = manager.get(key_triple)?;
-        let key_info = match key_info {
-            Some(key_info) => key_info,
-            _ => continue,
-        };
-
-        keys.push(KeyInfo {
-            provider_id: key_triple.provider_id,
-            name: key_triple.key_name().to_string(),
-            attributes: key_info.attributes,
-        });
+        Ok(keys)
     }
 
-    Ok(keys)
-}
+    /// Returns a Vec of ApplicationName of clients having keys in the provider.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error as a String if there was a problem accessing the Key Info Manager.
+    pub fn list_clients(&self, provider_id: ProviderID) -> Result<Vec<ApplicationName>, String> {
+        let key_triples = self.get_all(provider_id)?;
+        let mut clients = Vec::new();
 
-/// Returns a Vec of ApplicationName of clients having keys in the provider.
-///
-/// # Errors
-///
-/// Returns an error as a String if there was a problem accessing the Key Info Manager.
-pub fn list_clients(
-    manager: &dyn ManageKeyInfo,
-    provider_id: ProviderID,
-) -> Result<Vec<ApplicationName>, String> {
-    let key_triples = manager.get_all(provider_id)?;
-    let mut clients = Vec::new();
+        for key_triple in key_triples {
+            if !clients.contains(key_triple.app_name()) {
+                let _ = clients.push(key_triple.app_name().clone());
+            }
+        }
 
-    for key_triple in key_triples {
-        if !clients.contains(key_triple.app_name()) {
-            let _ = clients.push(key_triple.app_name().clone());
+        Ok(clients)
+    }
+
+    /// Check if a key triple exists in the Key Info Manager and return a ResponseStatus
+    ///
+    /// # Errors
+    ///
+    /// Returns PsaErrorAlreadyExists if the key triple already exists or KeyInfoManagerError for
+    /// another error.
+    pub fn already_exists(&self, key_triple: &KeyTriple) -> Result<(), ResponseStatus> {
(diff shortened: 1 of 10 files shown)


src/key_info_managers/mod.rs
```diff
@@ -146,60 +146,82 @@ pub trait ManageKeyInfo {
...
+    /// Returns an error as a String if there was a problem accessing the Key Info Manager.
+    pub fn list_keys(
+        &self,
+        app_name: &ApplicationName,
+        provider_id: ProviderID,
+    ) -> Result<Vec<parsec_interface::operations::list_keys::KeyInfo>, String> {
+        use parsec_interface::operations::list_keys::KeyInfo;
+
+        let mut keys: Vec<KeyInfo> = Vec::new();
+        let key_triples = self.get_all(provider_id)?;
```
ionut-arm, line 173: Potentially dumb question, but wouldn't it make sense to extend `get_all` to take in 3 parameters: one `Option` for each of app name, provider and key name - and we filter everything in that one call, based on the combination of parameters provided. You're already iterating through all the keys in the KIM

src/key_info_managers/mod.rs
```diff
@@ -146,60 +146,82 @@ pub trait ManageKeyInfo {
...
-    for key_triple in key_triples {
-        if !clients.contains(key_triple.app_name()) {
-            let _ = clients.push(key_triple.app_name().clone());
+    /// Check if a key triple exists in the Key Info Manager and return a ResponseStatus
+    ///
+    /// # Errors
+    ///
+    /// Returns PsaErrorAlreadyExists if the key triple already exists or KeyInfoManagerError for
+    /// another error.
+    pub fn already_exists(&self, key_triple: &KeyTriple) -> Result<(), ResponseStatus> {
```
ionut-arm, line 220: I get that the doc above says how this method works, but to me `already_exists` implies that `Ok(())` would mean `yes`, not `no` - at least that's what I'd assume it does if I see it used somewhere in a PR. Maybe something like `is_unique` (or `check_unique`) matches the return type better? Or return `Result<bool, ResponseStatus>`?

src/providers/core/mod.rs
```diff
@@ -99,7 +104,12 @@ impl Provide for Provider {
 
         let mut clients: Vec<String> = Vec::new();
         for provider in &self.prov_list {
-            let mut result = provider.list_clients(_op)?;
+            let mut result = provider.list_clients(_op).unwrap_or_else(|e| {
+                error!("list_clients failed on provider with {}", e);
```
ionut-arm, line 108: Any way to include the provider ID in that error message?

ionut-arm: (same for the other methods that you changed)

src/providers/mbed_crypto/key_management.rs
```diff
@@ -99,29 +98,29 @@ impl Provider {
...
         let _guard = self
             .key_handle_mutex
             .lock()
             .expect("Grabbing key handle mutex failed");
 
         match psa_crypto_key_management::generate(key_attributes, Some(key_id)) {
-            Ok(_) => Ok(psa_generate_key::Result {}),
+            Ok(_) => {
+                // If this fails, then we will have a zombie key in the Mbed Crypto backend. This
+                // is probably not a big problem apart from memory consumption.
```
ionut-arm, line 119: It can still cause some operations to fail: if you restart the service and don't have a higher ID used (the IDs are allocated incrementally, right?) then you'll end up trying to use this one and that will fail, but the next generate/import will be fine, so it just needs a retry.

src/providers/pkcs11/key_metadata.rs
```diff
@@ -98,12 +96,9 @@ impl Provider {
         }
     }
 
-    pub(super) fn key_info_exists(&self, key_triple: &KeyTriple) -> Result<bool> {
+    pub(super) fn key_info_exists(&self, key_triple: &KeyTriple) -> Result<()> {
```
ionut-arm, line 99: Ah, I see - well, I'd still prefer `bool` or a name change or smth

ionut-arm (commented):
Cheers for the changes, it's nice to make the providers more consistent

src/key_info_managers/mod.rs
hug-dev: ah yes you are correct, now that I look at it it's kind of wrong 🤡 
I changed the return type following the same kind of convention we use in the interface:
* a method starting by `is` (an interrogation) return just a `bool`
* a method which name is a declaration returns a `Result`

I can change the name to `does_not_already_exist`

src/providers/core/mod.rs
hug-dev: I could do it with the `describe` method? I tried that rapidly but felt like it was going too far for this purpose

src/providers/mbed_crypto/key_management.rs
hug-dev: ah yes you are correct, I guess this is the problem with this method... And will be the same with PKCS11 even though we use a random key ID generator: if it happens that we create the same key ID of a zombie key then the operation will have to be retried. In the worst case, if that happens a lot, one would have to re-try the key generation several times before finding an unallocated key ID.
Should we try to delete the key then, if insert_key_id fails? Here and for the PKCS11 provider.

src/key_info_managers/mod.rs
ionut-arm: Yeah, and also because it seems like a handy way of dealing with arbitrary filters like this - just pass which combination you're interested in and you get it. (the name of the function isn't that specific either, it should be more like `get_all_per_provider`)

src/providers/core/mod.rs
ionut-arm: Yeah, was wondering about that... I'm just thinking that since the operation is unlikely to proper fail and since logs are the only thing that tells you what fails maybe having more specific messages would help

ionut-arm:
Oh, there's one more issue that I forgot to put in the issue or note here: 

In `psa_destroy_key_internal` in PKCS11 you have two destroy operations happening one after the other (because you could have two parts of a key), but currently if the first bit fails you just return - destroying the second part never happens. Now that you destroy the metadata regardless of that error code I think you should attempt to delete the 2nd part as well, regardless of what happens with the first one.  I guess if either fails you return an error

src/key_info_managers/mod.rs
hug-dev: >  (the name of the function isn't that specific either, it should be more like get_all_per_provider)

yeah agree with that one! I would say it's maybe better to change that in a subsequent PR? I can raise an issue for it

hug-dev:
> I think you should attempt to delete the 2nd part as well

Yes, makes sense!

src/key_info_managers/mod.rs
hug-dev: cc #326

hug-dev:
Addressed your comments in a new commit for easier review 🧟

src/providers/pkcs11/key_management.rs
```diff
@@ -399,45 +420,55 @@ impl Provider {
...
-        let _ = self.remove_key_id(&key_triple)?;
-
-        Ok(psa_destroy_key::Result {})
+        if let Err(e) = first_res {
+            Err(e)
+        } else if let Err(e) = second_res {
+            Err(e)
+        } else {
+            Ok(psa_destroy_key::Result {})
+        }
```
ionut-arm, line 472:
```suggestion
        first_res?;
        second_res?;
        Ok(psa_destroy_key::Result {})
```

Would this work?

ionut-arm (approved):
Thanks, looks good, one question but it's just a nit

src/providers/pkcs11/key_management.rs
hug-dev: Oh yes that's nice 👌

hug-dev:
I will also add the algorithm we follow for those things, in high level terms in the `Provide` trait for generate, import and delete so that future implementors can follow the same recipe.

PR #330, opened 2021-01-28
Trusted service provider

ionut-arm:
This PR merges back the work done to create and upstream the Trusted Service provider.

The TS provider isn't yet stable enough to be used in deployments, but we can nonetheless have the code in the `master` branch.

I'll go through the PRs that have been merged into this branch and collect the comments, just to make sure they've been addressed or raised as issues.


Cargo.toml
```diff
@@ -60,6 +65,7 @@ mbed-crypto-provider = ["psa-crypto"]
 pkcs11-provider = ["pkcs11", "picky-asn1-der", "picky-asn1", "picky-asn1-x509", "psa-crypto", "rand"]
 tpm-provider = ["tss-esapi", "picky-asn1-der", "picky-asn1", "picky-asn1-x509", "hex"]
 cryptoauthlib-provider = ["rust-cryptoauthlib"]
+trusted-service-provider = ["mbed-crypto-provider", "bindgen", "prost-build", "prost"]
```
hug-dev, line 68: Sorry, this is probably not the place for a full review but have we considered pre-generating the bindings + protobuf here?

src/providers/trusted_service/context/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+
+    set_handle_impl!(DestroyKeyIn);
+    set_handle_impl!(SignHashIn);
+    set_handle_impl!(VerifyHashIn);
+    set_handle_impl!(ExportPublicKeyIn);
+}
+
+// TODO:
+// * RPC caller error handling
+// * proper logging
```
hug-dev, line 117: Would be nice to create issues for those if still needed as well

src/providers/trusted_service/key_management.rs
```diff
@@ -0,0 +1,149 @@
...
+};
+use parsec_interface::requests::{ProviderID, ResponseStatus, Result};
+use parsec_interface::secrecy::ExposeSecret;
+
+/// Check whether any key info exists for a given key triple in the Key Info Manager
+pub fn key_info_exists(key_triple: &KeyTriple, store_handle: &dyn ManageKeyInfo) -> Result<bool> {
+    store_handle
+        .exists(key_triple)
+        .map_err(key_info_managers::to_response_status)
+}
```
hug-dev, line 22: This can be replaced with the implemented on `ManageKeyInfo` `does_not_exist` if you wished, to also be able to directly do `store_handle.does_not_exist()?;`

src/providers/trusted_service/key_management.rs
```diff
@@ -0,0 +1,149 @@
...
+
+impl Provider {
+    pub(super) fn psa_generate_key_internal(
```
hug-dev, line 25: Thanks for following the same pattern!

src/providers/trusted_service/key_management.rs
```diff
@@ -0,0 +1,149 @@
...
+                data: pub_key.into(),
+            }),
+            Err(error) => {
+                format_error!("Export key status: ", error);
+                Err(error.into())
+            }
+        }
+    }
+
+    pub(super) fn psa_destroy_key_internal(
```
hug-dev, line 125: The pattern here also was changed to:
1. get the key ID from the key name using the KIM
2. destroy the key mappings
3. try to destroy the key

So that a key with the same name can be created again, even if it remains as zombie in the TS

hug-dev (commented):
Thanks!! I had already reviewed most of the code, just a couple of comments

Cargo.toml
ionut-arm: Could do, but I'm unsure how volatile both of those are, especially in the early days of the project

src/providers/trusted_service/key_management.rs
ionut-arm: Yeah, good shout, forgot that exists now

src/providers/trusted_service/key_management.rs
ionut-arm: Dun

src/providers/trusted_service/context/mod.rs
ionut-arm: Logging has been covered, it's just error handling now and I've raise #332 for that

hug-dev (approved):
Thanks!

PR #333, opened 2021-02-02
Parsec PsaHashCompare operation implementation for CryptoAuthLib provider

akazimierskigl:
An implementation of a psa_hash_compare callback inside a cryptoauthlibrary-provider responsible for PsaHashCompare operation/opcode using CryptoAuthentication Library API calls.

### DIFF
--- a/src/providers/cryptoauthlib/hash.rs
+++ b/src/providers/cryptoauthlib/hash.rs
@@ -39,7 +39,6 @@ impl Provider {
         &self,
         op: psa_hash_compare::Operation,
     ) -> Result<psa_hash_compare::Result> {
-        let op_hash = op.hash.to_vec();
         let alg_len = op.alg.hash_length();
         // calculate input hash
         let op_compute = psa_hash_compute::Operation {
@@ -49,13 +48,13 @@ impl Provider {
         match self.psa_hash_compute_internal(op_compute) {
             Ok(psa_hash_compute::Result { hash }) => {
                 // compare hashes length
-                if op_hash.len() != alg_len || hash.as_slice().len() != alg_len {
+                if op.hash.len() != alg_len || hash.len() != alg_len {
                     let error = ResponseStatus::PsaErrorInvalidArgument;
                     format_error!("Hash length comparison failed: ", error);
                     return Err(error);
                 }
                 // compare hashes
-                if op_hash != hash.as_slice() {
+                if op.hash != hash {
                     let error = ResponseStatus::PsaErrorInvalidSignature;
                     format_error!("Hash comparison failed: ", error);
                     return Err(error);
@@ -63,11 +62,7 @@ impl Provider {
                 // return result
                 Ok(psa_hash_compare::Result)
             }
-            _ => {
-                let error = ResponseStatus::PsaErrorGenericError;
-                format_error!("Hash computation failed ", error);
-                Err(error)
-            }
+            Err(error) => Err(error),
         }
     }
 }
--- a/src/providers/cryptoauthlib/hash.rs
+++ b/src/providers/cryptoauthlib/hash.rs
@@ -1,6 +1,7 @@
 // Copyright 2021 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
 use super::Provider;
+use log::error;
 use parsec_interface::operations::psa_algorithm::Hash;
 use parsec_interface::operations::psa_hash_compare;
 use parsec_interface::operations::psa_hash_compute;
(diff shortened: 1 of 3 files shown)


src/providers/cryptoauthlib/hash.rs
```diff
@@ -33,4 +34,40 @@ impl Provider {
...
+                if op_hash != hash.as_slice() {
+                    let error = ResponseStatus::PsaErrorInvalidSignature;
+                    format_error!("Hash comparison failed: ", error);
+                    return Err(error);
+                }
+                // return result
+                Ok(psa_hash_compare::Result)
+            }
+            _ => {
+                let error = ResponseStatus::PsaErrorGenericError;
```
hug-dev, line 67: I think you could return the error from `psa_hash_compute_internal` here instead of `PsaErrorGenericError`. Just doing:
```rust
let hash = self.psa_hash_compute_internal(op_compute)?.hash;
```
is probably fine instead of the match line 49

src/providers/cryptoauthlib/hash.rs
```diff
@@ -33,4 +34,40 @@ impl Provider {
             }
         }
     }
+
+    pub(super) fn psa_hash_compare_internal(
+        &self,
+        op: psa_hash_compare::Operation,
+    ) -> Result<psa_hash_compare::Result> {
+        let op_hash = op.hash.to_vec();
```
hug-dev, line 42: Do you need to use `to_vec`? I think since `Deref` is implement on those types you might not have to convert them to plain `Vec`. The good thing with keeping them in their original form is that they will get zeroized at the end.

src/providers/cryptoauthlib/hash.rs
```diff
@@ -33,4 +34,40 @@ impl Provider {
...
+                    let error = ResponseStatus::PsaErrorInvalidArgument;
+                    format_error!("Hash length comparison failed: ", error);
+                    return Err(error);
+                }
+                // compare hashes
+                if op_hash != hash.as_slice() {
+                    let error = ResponseStatus::PsaErrorInvalidSignature;
+                    format_error!("Hash comparison failed: ", error);
+                    return Err(error);
+                }
```
hug-dev, line 62: Implementing the comparison in software is probably fine but I am wondering if the callers won't expect it to be done in the ATEC chip? Is it possible?
For example I am not sure if we care about the fact that this comparison won't be done in constant time, and if that might leak data to the original data being hashed. I am probably over-anxious

hug-dev (commented):
Thanks for the PR! A few comments

src/providers/cryptoauthlib/hash.rs
akazimierskigl: I did it a bit different but code duplication is removed and error type is taken from `psa_hash_compute_internal` result.

src/providers/cryptoauthlib/hash.rs
RobertDrazkowskiGL: Yes, this is true we violate the promise of having everything done in hardware. We are not aware of a possibility to offload this comparison to ATECC. So the conclusion is that most of hard job is done by hardware, while the "easy" tasks are left for software.
Please also note - whatever a message length is, the hash is always 32 bytes long, so the comparison is in an almost constant time, but still depending on host processor load, unfortunately.
I am not sure if I understand correctly the comment, but if ATECC could do the comparison, the message and the input hash would be transferred to it anyway. The only new information leaked with this implementation, and only locally (between the i2c ends i.e. CPU and ATECC) is the output hash calculated for the message. I hope this is acceptable.

hug-dev (approved):
Thanks for the changes!

src/providers/cryptoauthlib/hash.rs
hug-dev: Thanks :)

src/providers/cryptoauthlib/hash.rs
hug-dev: Thanks for the response. That is fine on my side!
I believe that for use-cases that care about efficiency, it's probably that the hash operations will be done in software anyway (as in, on the client side) with heavily optimised library. It is however useful to have those operations defined here, for convenience.

src/providers/cryptoauthlib/hash.rs
```diff
@@ -33,4 +34,35 @@ impl Provider {
...
+        let alg_len = op.alg.hash_length();
+        // calculate input hash
+        let op_compute = psa_hash_compute::Operation {
+            alg: op.alg,
+            input: op.input,
+        };
+        match self.psa_hash_compute_internal(op_compute) {
+            Ok(psa_hash_compute::Result { hash }) => {
+                // compare hashes length
+                if op.hash.len() != alg_len || hash.len() != alg_len {
```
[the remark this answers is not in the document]
akazimierskigl: https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_hash_compare.html

> `PsaErrorInvalidArgument`: The length of `input` or `hash` does not match the hash size for `alg`.

I was trying to cover length of `input` part in this statement. I believe it meant to be length of calculated hash from input because input message can be any size. So I think according to this page type of logged error is correct.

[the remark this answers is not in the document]
ionut-arm: The `input` part is actually about the input size - there are limits on it, usually something on the order of 2^64 (so it's more of a theoretical limit than a practical one) so I wouldn't bother checking that - the machine would be long dead by the time it's able to transfer that much data. (I did check with one of the engineers working on PSA Crypto for this one!)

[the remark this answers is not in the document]
ionut-arm: (ps: I'm happy to leave this as is, there's barely any impact from this, it's more of a nit - it's more the logging change that's relevant)

ionut-arm (approved):
Thanks for the changes! 👍🏻

PR #334, opened 2021-02-09
Create test_in_docker.md

kevingzhang:
credit: Hugues de Valon

### DIFF
--- a/e2e_tests/provider_cfg/tpm/test_in_docker.md
+++ b/e2e_tests/provider_cfg/tpm/test_in_docker.md
@@ -1,0 +1,27 @@
+  # Gist
+  Run parset cargo test in docker
+  
+  # Steps
+  - Build docker using docker build . 
+  - Start container `docker run -it --rm Your_Image_Just_Built /bin/sh`
+  - Inside docker container run the following comments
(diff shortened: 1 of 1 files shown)


hug-dev:
Thanks for recording this in documentation!!
I think ideally we want those kind of guides on the Parsec book. Part of this info is already in [this section](https://parallaxsecond.github.io/parsec-book/parsec_service/tests/index.html#testing-the-tpm-provider-using-the-software-tpm) but I agree it is not self-sufficient and hard to find.
We need to think about a better strategy documentation on "how to get start" and refactor some of our pages in the book, possibly including the PR you sent.

For now I would say, let's wait but try to think about it when are to refactor our new starter documentation.

hug-dev:
Hey @kevingzhang !

I just added a PR [here](https://github.com/parallaxsecond/parsec-book/pull/98) with a basic "getting started" guide. Now it just uses the software Mbed Crypto provider but in the future it will also feature running inside a Docker container with all providers. There will also be examples in Rust and Go clients.
I am closing this PR for now as we will add the "getting started"/"quickstart" info in that part of the book!

PR #345, opened 2021-02-22
Upgrade all dependencies to their latest version

hug-dev:
Also fix some recent clippy errors (but not on the PKCS11 provider which
might change soon).


src/front/domain_socket.rs
```diff
@@ -74,7 +75,7 @@ impl DomainSocketListener {
                 let nfd = sd_notify::SD_LISTEN_FDS_START;
                 // Safe as listen_fds gives us the information that one file descriptor was
                 // received and its value starts from SD_LISTEN_FDS_START.
-                unsafe { UnixListener::from_raw_fd(nfd) }
+                unsafe { UnixListener::from_raw_fd(nfd.try_into()?) }
```
hug-dev, line 78: See https://github.com/lnicola/sd-notify/issues/3

ionut-arm (approved):
Looks like I've duplicated some of your changes brought by clippy. 

Which also makes me think that we should make the nightly run include at least the `clippy` part and some of the E2E tests

Issue #342, opened 2021-02-17
Investigate calculating test coverage


hug-dev:
We have a variety of different tests in Parsec, it would be good to know what is our test coverage. That would help us knowing if we need to do more or less.

ionut-arm:
I'm currently looking at [`tarpaulin`](https://crates.io/crates/cargo-tarpaulin) for coverage, trying to see if/how our test framework works with it.

Seems to work by calling it like this:
```
RUST_LOG=info cargo tarpaulin --forward --command build --features="mbed-crypto-provider,direct-authenticator" --run-types bins --timeout 3600 -- -c e2e_tests/provider_cfg/mbed-crypto/config.toml
```
And then running the tests from a separate thread, followed by a `pkill parsec`.

ionut-arm:
It's worth considering whether we want to have the code coverage checks running in the CI and then having the results published to something like [Codecov](https://about.codecov.io/)

hug-dev:
Is it possible to calculate it for *all* of our tests and not just the end to end ones?

> It's worth considering whether we want to have the code coverage checks running in the CI and then having the results published to something like Codecov

Would be nice! Personally I would prefer generating the results nightly to check what is the global coverage status instead of making it a requirement to merge a PR

ionut-arm:
> Is it possible to calculate it for all of our tests and not just the end to end ones?

As in, stress tests, persistency tests etc? I think so, I have a plan for how to do it

> Personally I would prefer generating the results nightly to check what is the global coverage status instead of making it a requirement to merge a PR

Yep, I don't think it's needed per PR, we just need to add it to the nightly stuff

hug-dev:
> As in, stress tests, persistency tests etc? I think so, I have a plan for how to do it

Ah yes those but I was also wondering about unit tests?

ionut-arm:
I'll try and figure that afterwards, currently trying with just integration testing - adding unit tests would be more problematic because it'd have to combine the results from multiple tarpaulin runs

ionut-arm:
I tried to run the service under tarpaulin and then run all the integration tests for every provider, sequentially, but it really didn't like it too much... It was getting a bunch of segfaults, the first ones I saw were due to using `drop(...)` in `main.rs` when reloading the service. I think there were more issues in other parts of the code, but I dropped it - will try with `grcov` which maybe is more straight forward to get working.

hug-dev:
[...]

PR #347, opened 2021-02-24
Add coverage checking in nightly run

ionut-arm:
This commit adds a way of checking code coverage for our E2E and unit
tests. Some of the E2E tests have been omitted for now (the stress tests
and the tests for all providers, in particular).

This is a first step in solving #342

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -9,7 +9,7 @@ set -e
 cleanup () {
     echo "Shutdown Parsec and clean up"
     # Stop Parsec if running
-    if [ -n "$PARSEC_PID" ]; then kill $PARSEC_PID || true ; fi
+    stop_service
     # Stop tpm_server if running
     if [ -n "$TPM_SRV_PID" ]; then kill $TPM_SRV_PID || true; fi
     # Remove the slot_number line added earlier
@@ -37,7 +37,9 @@ where PROVIDER_NAME can be one of:
     - mbed-crypto
     - pkcs11
     - tpm
+    - trusted-service
     - all
+    - coverage
 "
 }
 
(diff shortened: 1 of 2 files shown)


ci.sh
```diff
@@ -111,32 +126,38 @@ if [ "$PROVIDER_NAME" = "trusted-service" ]; then
...
+    # Check that Parsec successfully started and is running
+    pgrep -f target/debug/parsec >/dev/null
+else
+    # Start the service
+    eval $RUN_COMMAND
+
+    wait_for_service
+fi
 
 if [ "$PROVIDER_NAME" = "all" ]; then
```
hug-dev, line 162: I understand that you modified this file as you want to re-use parts of it but I am thinking that if you actually look at what needs to be executed under tarpolin, it's not that much. I think it is mostly the normal tests and the stress tests. I don't think the persistence test would yield a lot more of code coverage.
My point being that, we could just put those same lines (normal tests and stress tests) directly in the `coverage.sh` script without touching this file at all.

Maybe later if we see that it is causing problems we can still go back but I don't see the lines duplicated changing much in the future

hug-dev (commented):
Thanks for the PR! It would be nice to have some coverage set up 👌 I made one comment that might simplify the PR

ci.sh
ionut-arm: Yep, can try - I wanted to run the persistence tests as well because those impact the startup methods for the providers which seem fairly important. It's annoying that you can't just execute parts of a bash script (or, I don't know how that would happen?). Maybe I can split `ci.sh` into multiple scripts, one for each type of test?

ionut-arm: (Oh, and the stress test doesn't get executed, it's skipped, because it doesn't add anything to coverage)

hug-dev: > I wanted to run the persistence tests as well because those impact the startup methods for the providers which seem fairly important

ah I see. I guess for now you can just copy and paste the lines, in lack for a better way. We had an issue anyway to have a better testing infrastructure, with maybe more stuff in Rust

hug-dev (approved):
Thanks that looks good!!

ci.sh
```diff
@@ -74,6 +133,9 @@ while [ "$#" -gt 0 ]; do
                 TEST_FEATURES="--features=$1-provider"
             fi
         ;;
+        coverage )
+            PROVIDER_NAME=$1
```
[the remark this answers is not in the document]
ionut-arm: Yass

ci.sh
```diff
@@ -111,6 +173,44 @@ if [ "$PROVIDER_NAME" = "trusted-service" ]; then
...
 
+if [ "$PROVIDER_NAME" = "coverage" ]; then
+    PROVIDERS="mbed-crypto tpm pkcs11"
+    # Install tarpaulin and start Parsec with it
+    cargo install cargo-tarpaulin
+
+    mkdir -p reports
+
+    for provider in $PROVIDERS; do
+        PROVIDER_NAME=$provider
```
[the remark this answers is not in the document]
ionut-arm: Yes, to decide what new files to add

PR #348, opened 2021-02-25
Improve coverage script

ionut-arm:
This commit improves the coverage part of ci.sh by excluding a number of
files that we cannot test and removing the PKCS11 provider from the
build.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -174,26 +174,32 @@ if [ "$PROVIDER_NAME" = "trusted-service" ]; then
 fi
 
 if [ "$PROVIDER_NAME" = "coverage" ]; then
-    PROVIDERS="mbed-crypto tpm pkcs11"
-    # Install tarpaulin and start Parsec with it
+    PROVIDERS="mbed-crypto tpm" # pkcs11 not supported because of a segfault when the service stops; see: https://github.com/parallaxsecond/parsec/issues/349
+    EXCLUDES="fuzz/*,e2e_tests/*,src/providers/cryptoauthlib/*,src/providers/pkcs11/*,src/providers/trusted_service/*"
+    # Install tarpaulin
     cargo install cargo-tarpaulin
 
     mkdir -p reports
 
     for provider in $PROVIDERS; do
+        # Set up run
         PROVIDER_NAME=$provider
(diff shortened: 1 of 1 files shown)


ci.sh
```diff
@@ -174,26 +174,32 @@ if [ "$PROVIDER_NAME" = "trusted-service" ]; then
 fi
 
 if [ "$PROVIDER_NAME" = "coverage" ]; then
-    PROVIDERS="mbed-crypto tpm pkcs11"
-    # Install tarpaulin and start Parsec with it
+    PROVIDERS="mbed-crypto tpm" # pkcs11 not supported because of a segfault when the service stops
```
hug-dev, line 177: Maybe it would be good to open an issue to reference that and link it here!

ci.sh
```diff
@@ -174,26 +174,32 @@ if [ "$PROVIDER_NAME" = "trusted-service" ]; then
...
     for provider in $PROVIDERS; do
+        # Set up run
         PROVIDER_NAME=$provider
         TEST_FEATURES="--features=$provider-provider"
         cp $(pwd)/e2e_tests/provider_cfg/$provider/config.toml $CONFIG_PATH
         mkdir -p reports/$provider
-        RUST_LOG=info cargo tarpaulin --out Xml --forward --command build --output-dir $(pwd)/reports/$provider --features="all-providers,all-authenticators" --run-types bins --timeout 3600 -- -c $CONFIG_PATH &
+
+        # Start service
+        RUST_LOG=info cargo tarpaulin --out Xml --forward --command build --exclude-files="$EXCLUDES" --output-dir $(pwd)/reports/$provider --features="all-providers,all-authenticators" --run-types bins --timeout 3600 -- -c $CONFIG_PATH &
```
hug-dev, line 192: Actually, does it make sense to do `all-providers` then? Should we remove the one we ignore? Are they automatically ignored if they are not compiled in?

ionut-arm: I just exclude them through the CLI option and they don't get instantiated anyway. Could change it to only compile what we need!

ci.sh
ionut-arm: Yeah, I'll open an issue!

PR #354, opened 2021-03-05
Implementation of PsaGenerateKey and PsaDestroyKey operations

RobertDrazkowskiGL:
1. Refactor for a new, thread safe(er) rust-cryptoauthlib 0.2.0
2. Introduction of a CryptoAuthLib dedicated key management infrastructure
3. Implementation of PsaGenerateKey and PsaDestroyKey as users of this infrastructure


hug-dev:
Hey 👋 
Thanks for the PR, great to see key management coming up.

I spent some time reading the code, trying to understand the logic behind, and before going too deep in reviewing specific things, I wanted to discuss the global design.

I will try to write here what I understood but please correct anything I say!

* The ATECC chip has a number of slot with a predefined configuration for each one of them, noted with [`AtcaSlot`](https://docs.rs/rust-cryptoauthlib/0.2.0/rust_cryptoauthlib/struct.AtcaSlot.html).
* This configuration is fetched during provider creation and stored in the `KeySlotStorage` structure, along with the information is a slot is occupied or not.
* The key mappings map a `KeyTriple` to a slot ID (also the index of the `KeySlotStorage`) and attributes
* When generating a key, we generate it in the first suitable slot: the first one with a compatible configuration for the attributes given.

A few questions now:
1. Do we assume that the Parsec service will be the only user of the ATECC device? Could multiple applications on the same host use the same ATECC device?
2. During the provider construction, how do you actually make sure that there is a valid key in the slot fetched from the mappings? To me, `key_validate_and_mark_busy` makes sure that the slot has still the correct configuration for the key and that the slot is free (which should always be the case during construction, unless somehow the mappings are corrupted and two slot IDs are the same?). Would it be possible that there is not a key at this slot? Could be if the `mappings` were badly managed or corrupted of if 1. is not assumed and that another application deleted the key in the slot?
3. More open to discussion, but would it make sense for the `KeySlotStorage` logic to be inside the `rust-cryptoauthlib` crate? One would call a method to generate a key with a specific type and it would return the slot where the key was created. The logic to cache the slot configs and verification would be inside the crate.

hug-dev:
> Well, I am aware of the size of all those changes, but my intention was to provide not only the infrastructure, but also the users of that infrastructure.

Oh that was not a criticism! I am ok with the size of the change too! I was just meaning that it's best to agree with the global logic before going into the tiny details 😄 

Thanks for answering!

> any "i2c" group member is (usually) allowed to write to any i2c device. If there are two unsynchronized users of the same ATECC device, we have a BIG trouble.

Ah ok. I think that would go under our threat model assumption 4: "Users with privilege rights are trusted to exercise them in a non-malicious way." Though, it seems pretty hard to enforce/verify that Parsec is the only process using it...

> There is no in-hardware information whether there is anything usable in a slot or not.

Would it be possible to check if a basic (and fast) operation using the key fails or not? As in an operation that would fail if the slot is empty? Not sure if possible or if that makes sense. Something like `atcab_get_zone_size()`?

> It is possible to implement rust-cryptoauthlib that way but it would violate two assumptions:

For 2., I was under the assumption that this is something (the `KeySlotStorage`) that would be needed for *any* user of `rust-cryptoauthlib`, not just Parsec. That in order to generate a key, you would first need to fetch the slots configuration and then find the appropriate one: what you did in Parsec could then be useful for all? But maybe I do not know all the use-cases.
For 1., I agree with you that you shouldn't go too far off the original spec but if some logic can be used by all and written in a  way that is idiomatic in Rust then I'd say it can belong in the wrapper crate!

RobertDrazkowskiGL:
> Would it be possible to check if a basic (and fast) operation using the key fails or not? As in an operation that would fail if the slot is empty? Not sure if possible or if that makes sense. Something like `atcab_get_zone_size()`?

As far as I know, the answer is no, it is not possible. The only exception I am aware of is the slot with `is_locked == true`, but this is pretty easy to check anyway. And I am afraid there are no “fast” operations over I2C bus (up to 1Mbps, but more than 7 bytes of metadata overhead per request).

> 
> > It is possible to implement rust-cryptoauthlib that way but it would violate two assumptions:
> 
> For 2., I was under the assumption that this is something (the `KeySlotStorage`) that would be needed for _any_ user of `rust-cryptoauthlib`, not just Parsec. That in order to generate a key, you would first need to fetch the slots configuration and then find the appropriate one: what you did in Parsec could then be useful for all? But maybe I do not know all the use-cases.

I think the way Parsec uses ATECC and CALib is pretty unusual, at least at the moment. As you may have already noticed, it is not designed for a multi-user environment and safe privilege separation. There is no kernel driver owning exclusively the device and offering a common abstract interface or a user space daemon dynamically managing the globally shared ATECC configuration and contents.
Let me give you an example. It is not an unusual ATECC use case to have all the slots having a defined and known semantics, equipped with keys in the factory and just locked to avoid any manipulations. The whole key management requested by Parsec would be simply a waste of resources.
Therefore I believe that this is not the task for rust-cryptoauthlib to store this kind of information.


> For 1., I agree with you that you shouldn't go too far off the original spec but if some logic can be used by all and written in a way that is idiomatic in Rust then I'd say it can belong in the wrapper crate!

In my opinion, this is not only a question of being far from spec, but also a question of a good design. The approach we try to follow is to define which information belongs to Parsec, which belongs to ATECC+CALib.

The free/busy slot management (and, less important, slot reference counting) belongs to Parsec, because only Parsec “knows” which slots are free or busy. It is not available directly but can be extracted from the key info manager. And this is where `KeySlotStorage` helps, managing these slots.

The hardware information, mostly slot configuration here, belongs to rust-cryptoauthlib and should be stored/cached/accessed using rust-cryptoauthlib.

There are things that belong to both - if the hardware slot is locked, the key is considered busy even without having a `KeyTriple->KeyInfo` mapping. If the `KeyInfo` attributes are not compatible with hardware slot configuration, it is considered “free” and corresponding `KeyTriple->KeyInfo` mapping is removed. All such cases belong to the interface between the two, i.e. to Parsec CryptoAuthLib provider. And if that needs a synchronization `KeySlotStorage` is used.

Please note, presented above “functionality ownership” is just my view and I am open for the discussion.

RobertDrazkowskiGL:
Thank you.

hug-dev (commented):
Here is my review, sorry for the load of comments 😓 
Some of them are logic questions/comments and some are more to make the code more Rust idiomatic. I won't be too annoying about that last part but could be something good to have in the future (for example using more `Result`)!

src/providers/cryptoauthlib/mod.rs
```diff
@@ -17,35 +18,138 @@ use std::collections::HashSet;
...
+        // First get the device, initialise it and communication channel with it
+        let device = match rust_cryptoauthlib::AteccDevice::new(atca_iface) {
+            Ok(x) => x,
             _ => return None,
         };
-        let cryptoauthlib_provider = Provider { device };
+        let provider_id = ProviderID::CryptoAuthLib;
+        // ATECC is useful for non-trivial usage only when its configuration is locked
+        let mut is_locked = false;
+        let err = device.configuration_is_locked(&mut is_locked);
```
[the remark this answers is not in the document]
akazimierskigl: This function will be updated to this form in upcoming release of rust-cryptoauthlib. We will make this change in parsec then.

hug-dev:
Hello 👋 Don't hesitate to ping me here or on the Slack channel when my re-review is needed. Don't want to block you.

RobertDrazkowskiGL:
Hello
Sorry for the delay - Artur took a few days off, he is back tomorrow with new corrections.

akazimierskigl:
@hug-dev 
Hello, I think we are ready for re-review. There is one test failing for PKCS 11 provider which we will try to resolve very soon. We've addressed most of comments you've posted if not I left comment with details. Update of e2e tests is integrated and pull requests which were merged to main branch on meantime are also merged to this branch.

hug-dev:
Great, thanks! I will have a look again soon. Don't worry about the PKCS11 test, most probably it is not your fault, those tests are failing often for various reasons not under our control.

e2e_tests/tests/per_provider/normal_tests/auth.rs
```diff
@@ -12,10 +12,16 @@ fn two_auths_same_key_name() -> Result<()> {
...
     client.set_default_auth(Some(auth1));
+    #[cfg(not(feature = "cryptoauthlib-provider"))]
     client.generate_rsa_sign_key(key_name.clone())?;
 
     client.set_default_auth(Some(auth2));
-    client.generate_rsa_sign_key(key_name)
+    #[cfg(not(feature = "cryptoauthlib-provider"))]
+    let result = client.generate_rsa_sign_key(key_name.clone());
+    #[cfg(feature = "cryptoauthlib-provider")]
+    let result = client.generate_ecc_key_pair_secpr1_ecdsa_sha256(key_name.clone());
```
hug-dev, line 22: To make the test useful, can you also add those two lines after line 16? Basically here we are creating two keys with the same names but different app names to check that it works correctly.

e2e_tests/tests/per_provider/normal_tests/create_destroy_key.rs
```diff
@@ -2,28 +2,37 @@
 // SPDX-License-Identifier: Apache-2.0
 use e2e_tests::TestClient;
 use parsec_client::core::interface::requests::ResponseStatus;
+use parsec_client::core::interface::requests::Result;
+#[cfg(not(feature = "cryptoauthlib-provider"))]
 use picky_asn1_x509::RSAPublicKey;
 
 #[test]
-fn create_and_destroy() {
+fn create_and_destroy() -> Result<()> {
```
hug-dev, line 10: Not returning anything and using `unwrap` is actually easier as with `unwrap` you can see exactly which line panicked but, when returning errors, you don't know which line returned it 😬

e2e_tests/tests/per_provider/normal_tests/create_destroy_key.rs
```diff
@@ -2,28 +2,37 @@
...
-    client.destroy_key(key_name).unwrap();
+    #[cfg(not(feature = "cryptoauthlib-provider"))]
+    client.generate_rsa_sign_key(key_name.clone())?;
+    #[cfg(feature = "cryptoauthlib-provider")]
+    client.generate_ecc_key_pair_secpr1_ecdsa_sha256(key_name.clone())?;
+    client.destroy_key(key_name)
 }
 
+#[cfg(not(feature = "cryptoauthlib-provider"))]
+// This test nakes no sense for calib - it does not destroy created keys
```
hug-dev, line 23: When generating a key, you can check if the mapping exist or return an error otherwise? Using the `does_not_exist` method.

e2e_tests/tests/per_provider/normal_tests/create_destroy_key.rs
```diff
@@ -37,53 +46,67 @@ fn destroy_without_create() {
...
 
     let public_key: RSAPublicKey = picky_asn1_der::from_bytes(&public_key).unwrap();
     assert_eq!(
         public_key.public_exponent.as_unsigned_bytes_be(),
         [0x01, 0x00, 0x01]
     );
+    Ok(())
 }
 
+#[cfg(not(feature = "cryptoauthlib-provider"))]
```
hug-dev, line 107: Could you do it through:
```rust
    if !client.is_operation_supported(Opcode::PsaImportKey) {
        return Ok(());
    }
```
?
In case the operation is added in the future

e2e_tests/tests/per_provider/normal_tests/create_destroy_key_calib.rs
```diff
@@ -0,0 +1,58 @@
+// Copyright 2019 Contributors to the Parsec project.
```
hug-dev, line 1: This file is no longer needed, right?

e2e_tests/tests/per_provider/normal_tests/hash.rs
```diff
@@ -11,10 +11,12 @@ const SHA_256: [u8; 32] = [
     0x0d, 0xc4, 0xbc, 0x13, 0xfd, 0x91, 0x74, 0x52, 0x92, 0x24, 0xc3, 0x8e, 0x0e, 0xe0, 0x75, 0xfa,
     0x9e, 0xd8, 0x0b, 0x78, 0x47, 0xe6, 0xae, 0xa7, 0x6a, 0xe9, 0x8c, 0xf9, 0xdd, 0xd9, 0x26, 0x69,
 ];
+#[cfg(not(feature = "cryptoauthlib-provider"))]
```
hug-dev, line 14: Are those tests failing? Are yo utesting hashing with `hash_calib`?

e2e_tests/tests/per_provider/normal_tests/hash_calib.rs
```diff
@@ -0,0 +1,64 @@
+// Copyright 2020 Contributors to the Parsec project.
```
hug-dev, line 1: Know that this file won't get executed if the module is not put in `e2e_tests/tests/per_provider/normal_tests/mod.rs` 😋 Also if those are the tests that pass on the crypto auth lib provider, I think it's still worth putting them inside the `hash.rs` file so that all providers benefit from this testing.

src/providers/cryptoauthlib/key_operations.rs
```diff
@@ -0,0 +1,85 @@
...
+impl Provider {
+    pub(super) fn psa_generate_key_internal(
+        &self,
+        app_name: ApplicationName,
+        op: psa_generate_key::Operation,
+    ) -> Result<psa_generate_key::Result> {
+        let key_name = op.key_name;
+        let key_triple = self.key_info_store.get_key_triple(app_name, key_name);
+
+        self.key_info_store.does_not_exist(&key_triple)?;
```
hug-dev, line 17: Talking about the other comment in the test file: this line should return `PsaErrorAlreadyExists` if the same key is created.

hug-dev (commented):
Alright second round! Sorry if this is taking a long time... Thanks a lot for sorting out the tests with `is_operation_supported` 👌

src/providers/cryptoauthlib/key_slot.rs
```diff
@@ -0,0 +1,231 @@
...
+    pub status: KeySlotStatus,
+    /// Hardware configuration of a slot
+    pub config: rust_cryptoauthlib::SlotConfig,
+}
+
+impl Default for AteccKeySlot {
+    fn default() -> AteccKeySlot {
+        unsafe { std::mem::zeroed() }
+    }
+}
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: No, it does not work - there is no default for SlotConfig in a released version. But rust_cryptoauthlib version 0.3.0 will provide this default.

src/providers/cryptoauthlib/key_slot.rs
```diff
@@ -0,0 +1,231 @@
...
+            Type::Hmac => !self.config.no_mac,
+            Type::Aes => self.config.key_type == rust_cryptoauthlib::KeyType::Aes,
+            Type::EccKeyPair {
+                curve_family: EccFamily::SecpR1,
+            }
+            | Type::EccPublicKey {
+                curve_family: EccFamily::SecpR1,
+            } => {
+                // There may be a problem here: P256 private key has 256 bits (32 bytes),
+                // but the uncompressed public key is 512 bits (64 bytes)
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: Our expert insists on keeping this comment for some time yet. Maybe it can be improved somehow.

src/providers/cryptoauthlib/key_slot.rs
```diff
@@ -0,0 +1,231 @@
...
+
+impl Default for AteccKeySlot {
+    fn default() -> AteccKeySlot {
+        unsafe { std::mem::zeroed() }
+    }
+}
+
+impl AteccKeySlot {
+    // Check if software key attributes are compatible with hardware slot configuration
+    pub fn key_attr_vs_config(&self, key_attr: &Attributes) -> Result<(), ResponseStatus> {
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: Yes, we have thought, and we have even started working on it. We plan to submit a PR with test improvements exclusively soon.

src/providers/cryptoauthlib/key_management.rs
```diff
@@ -0,0 +1,56 @@
...
+            Type::EccKeyPair {
+                curve_family: EccFamily::SecpR1,
+            }
+            | Type::EccPublicKey {
+                curve_family: EccFamily::SecpR1,
+            } => {
+                if attributes.bits == 256 {
+                    Ok(rust_cryptoauthlib::KeyType::P256EccKey)
+                } else {
+                    Err(ResponseStatus::PsaErrorInvalidArgument)
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: Accepted.

src/providers/cryptoauthlib/key_slot.rs
```diff
@@ -0,0 +1,221 @@
...
+                if self.status == KeySlotStatus::Free {
+                    self.status = status;
+                }
+            }
+            KeySlotStatus::Locked => {
+                if self.status == KeySlotStatus::Free || self.status == KeySlotStatus::Busy {
+                    self.status = status;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: Yes it would be. At the moment I plan to leave those checks modified for problem reporting, but the status should be updated anyway.

e2e_tests/tests/per_provider/normal_tests/auth.rs
RobertDrazkowskiGL: Accepted.

e2e_tests/tests/per_provider/normal_tests/create_destroy_key.rs
RobertDrazkowskiGL: Well, `Result<()>` will be removed.

e2e_tests/tests/per_provider/normal_tests/create_destroy_key.rs
RobertDrazkowskiGL: Legacy comment removed.

e2e_tests/tests/per_provider/normal_tests/create_destroy_key.rs
RobertDrazkowskiGL: I could, but there is more to be done - ATECC does not support RSA, it supports ECC only. I plan to leave this for the time being.

e2e_tests/tests/per_provider/normal_tests/create_destroy_key_calib.rs
RobertDrazkowskiGL: File removed.

e2e_tests/tests/per_provider/normal_tests/hash.rs
RobertDrazkowskiGL: ATECC supports SHA256 only. And hash_calib.rs willl be removed.

src/providers/cryptoauthlib/key_slot.rs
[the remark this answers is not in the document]
hug-dev: sounds good!

src/providers/cryptoauthlib/key_slot.rs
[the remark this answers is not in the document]
hug-dev: Ok that's good for me!

src/providers/cryptoauthlib/key_slot.rs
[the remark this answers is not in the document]
hug-dev: Ahh right, forgot about that. That's ok for me.

hug-dev (approved):
Thank you! That looks good to me now 😄

src/providers/cryptoauthlib/key_slot.rs
```diff
@@ -0,0 +1,219 @@
...
+                            key_attr.key_type,
+                            Type::EccPublicKey { .. } | Type::DhPublicKey { .. }
+                        )
+                }
+                _ => true,
+            }
+        }
+        if !result {
+            return false;
+        }
```
ionut-arm, line 118: Is this part needed?

src/providers/cryptoauthlib/key_slot_storage.rs
```diff
@@ -0,0 +1,110 @@
...
+    pub fn find_suitable_slot(&self, key_attr: &Attributes) -> Result<u8, ResponseStatus> {
+        let mut key_slots = self.storage.write().unwrap();
+        for slot in 0..rust_cryptoauthlib::ATCA_ATECC_SLOTS_COUNT {
+            if !key_slots[slot as usize].is_free() {
+                continue;
+            }
+            match key_slots[slot as usize].key_attr_vs_config(key_attr) {
+                Ok(_) => match key_slots[slot as usize].set_slot_status(KeySlotStatus::Busy) {
+                    Ok(()) => return Ok(slot),
+                    Err(err) => return Err(err),
```
ionut-arm, line 103: Could this be `=> continue,` as well, or is it critical to report the error?

ionut-arm (approved):
Thanks, only left a couple of comments but I'm ok with the change overall!

src/providers/cryptoauthlib/key_slot_storage.rs
RobertDrazkowskiGL: Difficult to say. This line is executed when there is an attempt to change slot status from "Locked" to "Busy". This is forbidden and this line should be unreachable. What to do when an unreachable branch is executed?

ionut-arm: There's always the [`unreachable!`](https://doc.rust-lang.org/std/macro.unreachable.html) macro! We've not used it much (because it panics if that path is ever taken), but it seems suitable here

src/providers/cryptoauthlib/key_slot.rs
RobertDrazkowskiGL: It can be removed - it just prevents further checks if something is known to be forbidden. So it is kind of an micro-optimization.

src/providers/cryptoauthlib/key_slot_storage.rs
RobertDrazkowskiGL: We have three policies to chose from
1. "ignore and go further", i.e. `=> continue`
2. "report the problem but try to continue". i.e. `=> return Err(err)`
3. "panic"

The first and the second give a chance to continue operation even though a problem with the logic is found. I would rather avoid calling panic (this way or another) in the field, but for tests it is fine. Maybe this would be the way to go? I mean something like this:
```
=> {
#[cfg(test)]
unreachable!();
#[cfg(not(test))]
return Err(err);
}
```

ionut-arm: It's ok, configuring it like that is completely overkill, was just curious about the initial reasoning about returning the error vs just continuing, I'm ok with things as they are

PR #357, opened 2021-03-15
Change CI to use published Docker image

ionut-arm:
This commit modifies the CI to use a container image published to the
Github container repository. The Dockerfile for the image is also
included.


e2e_tests/docker_image/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+# Download and install TSS 2.0
+RUN git clone https://github.com/tpm2-software/tpm2-tss.git --branch 2.3.3
+RUN cd tpm2-tss \
+	&& ./bootstrap \
+	&& ./configure \
+	&& make -j$(nproc) \
+	&& make install \
+	&& ldconfig
+
+RUN rm -rf tpm2-tss
```
hug-dev, line 25: Would it work to use the [Ubuntu package](https://packages.ubuntu.com/source/bionic/tpm2-tss) here instead?

e2e_tests/docker_image/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+RUN cd SoftHSMv2 \
+	&& sh autogen.sh \
+	# Autogen might fail because of some limitations of autoconf, see:
+	# https://bugzilla.redhat.com/show_bug.cgi?id=1826935#c3
+	|| sh autogen.sh
+RUN cd SoftHSMv2 \
+	&& ./configure --disable-gost \
+	&& make \
+	&& make install
+RUN rm -rf SoftHSMv2
```
hug-dev, line 57: There is also a [`softhsm`](https://packages.ubuntu.com/bionic/softhsm) package avaible

e2e_tests/docker_image/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+
+# Create a new token in a new slot. The slot number assigned will be random
+# and is found with the find_slot_number script.
+RUN softhsm2-util --init-token --slot 0 --label "Parsec Tests" --pin 123456 --so-pin 123456
+
+# Install Rust toolchain for all users
```
hug-dev, line 63: Nice 👌

e2e_tests/docker_image/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+ENV PKG_CONFIG_PATH /usr/local/lib/pkgconfig
+
+RUN apt update
+RUN apt install -y autoconf-archive libcmocka0 libcmocka-dev procps
+RUN apt install -y iproute2 build-essential git pkg-config gcc libtool automake libssl-dev uthash-dev doxygen libjson-c-dev
+RUN apt install -y --fix-missing wget python3 cmake clang
+RUN apt install -y gcc-multilib
+RUN apt install -y gcc-arm-linux-gnueabihf
+RUN apt install -y gcc-aarch64-linux-gnu
+RUN apt install -y libini-config-dev libcurl4-openssl-dev tpm2-tools curl libgcc1
```
hug-dev, line 12: Did you by any chance check if all of these are actually needed 😬 ?

e2e_tests/docker_image/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+COPY _exec_wrapper /usr/local/bin/
+RUN ls /opt/rust/bin | xargs -n1 -I% ln -s /usr/local/bin/_exec_wrapper /usr/local/bin/$(basename %)
+
+# Adding the targets for the cross-compilation tests
+RUN rustup target add armv7-unknown-linux-gnueabihf
+RUN rustup target add aarch64-unknown-linux-gnu
+
+# Add users for multitenancy tests
+RUN useradd -m parsec-client-1
+RUN useradd -m parsec-client-2
```
hug-dev, line 82: Maybe a:
```
RUN apt autoremove
```
or something similar would slim the container even more?

e2e_tests/docker_image/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+# users cannot modify the toolchains or install new ones.
+# See: https://github.com/rust-lang/rustup/issues/1085
+ENV RUSTUP_HOME /opt/rust
+ENV CARGO_HOME /opt/rust
+RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --no-modify-path
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+# Install the wrappers for the Rust binaries installed earlier
+COPY _exec_wrapper /usr/local/bin/
+RUN ls /opt/rust/bin | xargs -n1 -I% ln -s /usr/local/bin/_exec_wrapper /usr/local/bin/$(basename %)
```
hug-dev, line 74: What does this line actually does with `_exec_wrapper`?

hug-dev (commented):
Thanks!! That will help a lot for testing both locally and on CI.

e2e_tests/docker_image/Dockerfile
ionut-arm: Hmm, I only searched for what's directly accessible from `apt` (and it doesn't appear there), but maybe if I add the correct registry... Will try.

e2e_tests/docker_image/Dockerfile
ionut-arm: And a [`softhsm2`](https://packages.ubuntu.com/source/bionic/softhsm2) but it was throwing some weird errors when used :( so I had to revert to this one

e2e_tests/docker_image/Dockerfile
ionut-arm: Yes, I started from scratch - they're all needed 😞

e2e_tests/docker_image/Dockerfile
ionut-arm: Just tried running it inside the built container - only removed about 60 MB of stuff. Will add it, though. I might try and look up which of the dependencies are no longer needed, maybe we can slim some of that too, though I think the big ones are needed (gcc, clang...). Do we need Python for any "runtime compilation" step? 🤔

e2e_tests/docker_image/Dockerfile
ionut-arm: Searches for each file in `/opt/rust/bin` (e.g. `rustc`) and creates a symlink that points at `_exec_wrapper`. So you end up with, e.g., `/usr/local/bin/rustc -> /usr/local/bin/_exec_wrapper` which then calls the real `rustc` with all the args.

e2e_tests/docker_image/Dockerfile
ionut-arm: Oh... The thing you linked is a source package - it just contains the source that you build from. Looks like those sources were used to compile [`libsapi`](https://packages.ubuntu.com/bionic/libsapi-dev), so no `esapi`. They also mention on their [Install](https://github.com/tpm2-software/tpm2-tss/blob/master/INSTALL.md) page that there's a package available for Fedora, but there's no mention of Ubuntu/Debian.

I was wondering if we should switch to testing on Fedora, I think there are a few dependencies that we could get for free.

e2e_tests/docker_image/Dockerfile
hug-dev: ah yes true, maybe some dependencies are only needed to build things but then not needed anymore? Python is heavily used by Ubuntu, last time I uninstalled it on my system, it destroyed everything 🔥

ionut-arm: Hmm, yeah, I've been prodding here and there, but don't think there's much space I can save.. :( The libraries that I can remove are small, around 1MB.

e2e_tests/docker_image/Dockerfile
hug-dev: Worth to check but it's fine if the size is not going to be a lot smaller actually..

e2e_tests/docker_image/Dockerfile
hug-dev: oh that's cool!

hug-dev (approved):

ionut-arm:
Ok, build time dropped from 23 minutes to 12 (dominated by the `all-providers` build) 📉 

I'll try to integrate the `trusted-service-provider` build in there as well, shouldn't be too bad.

ionut-arm:
Ok, this PR is READY 🔥 

I moved the Trusted Services stuff in the Ubuntu 18 image (which clocks in at around 4 GB for the final image - about 9 with all the precursors), everything in Github Actions is now using the same one!

I've also added Trusted Service provider to coverage reporting! ☂️

e2e_tests/docker_image/Dockerfile
ionut-arm: Initially I had added this, but then things started failing because some useful libraries were being removed... So I've scrapped it again, but it's good to know that exists

hug-dev (approved):
Looks good!! Thank you 😄

hug-dev:
Would it be possible to host the image (or images in the future) in their own repo so that when a PR is merged, the image is automatically built and upstreamed?

PR #368, opened 2021-03-25
Prepare for 0.7.2

hug-dev:
The `docs` feature no longer exists on `tss-esapi`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "parsec-service"
-version = "0.7.1"
+version = "0.7.2"
 authors = ["Parsec Project Contributors"]
 description = "A language-agnostic API to secure services in a platform-agnostic way"
 license = "Apache-2.0"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -52,7 +52,7 @@ bindgen = { version = "0.57.0", optional = true }
 prost-build = { version = "0.7.0", optional = true }
 
 [package.metadata.docs.rs]
-features = ["pkcs11-provider", "tpm-provider", "tss-esapi/docs", "mbed-crypto-provider", "cryptoauthlib-provider"]
+features = ["pkcs11-provider", "tpm-provider", "mbed-crypto-provider", "cryptoauthlib-provider", "direct-authenticator"]
```
ionut-arm, line 55: No `unix-peer-credentials-authenticator`?

hug-dev: I think it will get added anyway because it is in the `default-features`

hug-dev: and that there is a `no-default-features` [option](https://docs.rs/about/metadata)

ionut-arm (approved):

PR #399, opened 2021-04-20
PSA_IMPORT_KEY introduction.

RobertDrazkowskiGL:
The next operation, with dedicated tests. The largest changes are in tests, the implementation is pretty straightforward.


src/providers/cryptoauthlib/key_slot.rs
```diff
@@ -70,21 +73,32 @@ impl AteccKeySlot {
...
-            | Type::EccPublicKey {
+            Type::EccPublicKey {
                 curve_family: EccFamily::SecpR1,
             } => {
-                // There may be a problem here: P256 private key has 256 bits (32 bytes),
-                // but the uncompressed public key is 512 bits (64 bytes)
-                key_attr.bits == 256
+                // The uncompressed public key is 512 bits (64 bytes).
+                // First few (7) slots are too short for ECC public key.
+                key_attr.bits == 512
```
hug-dev, line 99: Note from the [PsaImportKey](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_import_key.html) contract:

> The key size is always determined from the data buffer. If the key size in attributes is nonzero, it must be equal to the size from data.

Explained in a bit more detail in the [PSA API](https://armmbed.github.io/mbed-crypto/html/api/keys/management.html#psa_import_key).

What I mean is that the `bits` field does not correspond to the size of the data buffer that is imported but the size of the key, and for ECC keys that is the bit size associated with the curve. See the expected format [here](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_export_public_key.html#description).

So that if a ECC public key of a curve `secp256r1` is imported, then its `bits` field should still be 256 even if the data buffer is larger than that (512 in this case?).

src/providers/cryptoauthlib/mod.rs
```diff
@@ -199,6 +208,16 @@ impl Provider {
...
+
+    // Get the deployment specific write key. With this the keys can be stored encrypted in their slots.
+    // For ECC private keys it is obligatory, for Aes it is an option.
+    fn get_write_encrypt_key(&self) -> [u8; rust_cryptoauthlib::ATCA_KEY_SIZE] {
+        [
+            0x4D, 0x50, 0x72, 0x6F, 0x20, 0x49, 0x4F, 0x20, 0x4B, 0x65, 0x79, 0x20, 0x9E, 0x31,
+            0xBD, 0x05, 0x82, 0x58, 0x76, 0xCE, 0x37, 0x90, 0xEA, 0x77, 0x42, 0x32, 0xBB, 0x51,
+            0x81, 0x49, 0x66, 0x45,
+        ]
+    }
```
hug-dev, line 220: Sorry for my lack of knowledge on cryptoauthlib, but where does this value come from?

e2e_tests/tests/per_provider/normal_tests/import_key.rs
```diff
@@ -62,6 +67,21 @@ const KEY_PAIR_DATA: [u8; 609] = [
...
 ];
 
+#[cfg(feature = "cryptoauthlib-provider")]
+//#[allow(dead_code)]
+const PRIV_KEY_ECC: [u8; 32] = [
+    0xF5, 0xDB, 0x6B, 0xA1, 0x82, 0x22, 0xCE, 0xC1, 0x54, 0x53, 0xE5, 0x63, 0xDE, 0xC5, 0xC7, 0x94,
+    0xCD, 0x48, 0x95, 0xF2, 0x8C, 0xC2, 0x7F, 0x50, 0xC2, 0x7E, 0xC3, 0x1B, 0xAF, 0x44, 0xEA, 0x54,
+];
+#[cfg(feature = "cryptoauthlib-provider")]
+const PUB_KEY_ECC: [u8; 64] = [
```
hug-dev, line 77: Just checking, is this format the expected one as described [here](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_export_public_key.html)? If the Mbed Crypto provider supports importing public keys, would be nice to try on it as well to make sure!

e2e_tests/provider_cfg/cryptoauthlib/config_608a.toml
```diff
@@ -0,0 +1,33 @@
+[core_settings]
```
hug-dev, line 1: Do we use that file on the CI?

hug-dev (commented):
Thanks for adding support for importing keys! Left some comments.

src/providers/cryptoauthlib/mod.rs
RobertDrazkowskiGL: In my understanding, this is a completely arbitrary value. This particular one comes from rust-cryptoauthlib :-).
I think setting this write key should be adapted to a real world deployment scenario, e.g. this might be the MAC address of a network card or it can simply be stored in a configuration file, but at the moment there is just this getter.
Please note - this key is used for encrypting private keys while they are stored (imported or generated - not written) into their slots. Afterwards the private keys cannot be exported or anything like that, even encrypted. The only way to get the private key would be to (1) sniff for import messages or, (2) in case key is generated, sniff the i2c bus or to desolder the chip and try to read slot contents, which in general is pretty difficult.
Just yet another level of security.

e2e_tests/provider_cfg/cryptoauthlib/config_608a.toml
RobertDrazkowskiGL: No - it is only for our convenience. Hope it is not a big issue.

e2e_tests/tests/per_provider/normal_tests/import_key.rs
RobertDrazkowskiGL: Oh well - it is not. I will correct the format. Many thanks for spotting.

src/providers/cryptoauthlib/key_slot.rs
RobertDrazkowskiGL: Let me re-phrase:
1. if `bits` is equal 0, then only the appropriate contents of `data` determine key size (a number of bits).
2. if `bits` is larger than 0, then it must match the appropriate contents of `data` BUT for cryptoauthlib provider only 256 is acceptable, be it for private or public key.

Is that right?

src/providers/cryptoauthlib/key_operations.rs
```diff
@@ -84,4 +83,72 @@ impl Provider {
...
+            | x @ rust_cryptoauthlib::AtcaStatus::AtcaBadParam => {
+                error_status = x;
+            }
+            _ => {
+                error_status = rust_cryptoauthlib::AtcaStatus::AtcaUnimplemented;
+            }
+        }
+        match self.set_slot_status(slot_id as usize, KeySlotStatus::Free) {
+            Ok(()) => {
+                let error = ResponseStatus::PsaErrorInvalidArgument;
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: I will double check.

e2e_tests/tests/per_provider/normal_tests/asym_sign_verify.rs
```diff
@@ -141,16 +140,17 @@ fn only_verify_from_internet() -> Result<()> {
...
     if !client.is_operation_supported(Opcode::PsaGenerateKey) {
         return Ok(());
     }
     if !client.is_operation_supported(Opcode::PsaSignHash) {
         return Ok(());
     }
 
+    let mut hasher = Sha256::new();
+    hasher.update(b"Bob wrote this message.");
+    let hash = hasher.finalize().to_vec();
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: Should I put a copyright note for message contents? :-)

e2e_tests/tests/per_provider/normal_tests/import_key.rs
```diff
@@ -81,7 +101,13 @@ fn import_key() -> Result<()> {
     if !client.is_operation_supported(Opcode::PsaImportKey) {
         return Ok(());
     }
-    client.import_rsa_public_key(key_name, KEY_DATA.to_vec())
+
+    #[cfg(not(feature = "cryptoauthlib-provider"))]
+    let result = client.import_rsa_public_key(key_name, KEY_DATA.to_vec());
+    #[cfg(feature = "cryptoauthlib-provider")]
+    let result = client.import_ecc_public_secp_r1_ecdsa_sha256_key(key_name, PUB_KEY_ECC.to_vec());
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: I will look at this. Maybe not everything will be corrected, but the split will start.

e2e_tests/tests/per_provider/normal_tests/import_key.rs
```diff
@@ -323,3 +375,15 @@ fn import_key_pair() {
...
+#[test]
+fn import_ecc_private_key() {
+    let mut client = TestClient::new();
+    let key_name = String::from("import_ecc_private_key");
+    if !client.is_operation_supported(Opcode::PsaImportKey) {
+        return;
+    }
+
+    client.import_ecc_pair_secp_r1_key(key_name, PRIV_KEY_ECC.to_vec()).unwrap();
+}
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: Understood

e2e_tests/tests/per_provider/normal_tests/import_key.rs
```diff
@@ -1,14 +1,19 @@
 // Copyright 2019 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
 use e2e_tests::TestClient;
+#[allow(unused_imports)]
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: OK

src/providers/cryptoauthlib/key_operations.rs
```diff
@@ -84,4 +83,72 @@ impl Provider {
...
+            rust_cryptoauthlib::AtcaStatus::AtcaSuccess => {
+                match self
+                    .key_info_store
+                    .insert_key_info(key_triple, &slot_id, key_attributes)
+                {
+                    Ok(()) => {
+                        return Ok(psa_import_key::Result {});
+                    }
+                    Err(error) => {
+                        warn!("Insert key triple to KeyInfoManager failed. {}", error);
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: You are right. The slot should be set free. About the destroy key: I do not think it is needed. Thanks for spotting.

e2e_tests/tests/per_provider/normal_tests/asym_sign_verify.rs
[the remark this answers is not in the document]
hug-dev: Ah I see 😄

src/providers/cryptoauthlib/key_slot.rs
hug-dev: Yes that's correct! So ideally you should parse `data` to calculate the key size and if `bits` is larger than 0, make sure that the two values are equal. Since only 256 is supproted I guess you would need to check that `bits` the the key size are equal to 256.

There aren't any example of parsing ECC keys in Parsec at the moment. For the ECC key pairs, the format is explained  [here](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_export_key.html#description) and is the ASN.1 type `ECPrivateKey`:

```
   ECPrivateKey ::= SEQUENCE {
     version        INTEGER { ecPrivkeyVer1(1) } (ecPrivkeyVer1),
     privateKey     OCTET STRING,
     parameters [0] ECParameters {{ NamedCurve }} OPTIONAL,
     publicKey  [1] BIT STRING OPTIONAL
   }
```

We are using the `picky` crates to do ASN.1 parsing/serialization and before we were also defining our own ASN.1 type, like for RSA keys. See #201 for example. Maybe we can convince @CBenoit to also add support for `ECPrivateKey` in `picky-asn1-x509` or another crate?

[This](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_export_public_key.html#description) explains the representation of the ECC public key, which is not an ASN.1 structure and will need manual parsing I think.

e2e_tests/provider_cfg/cryptoauthlib/config_608a.toml
hug-dev: Ah as in you are using it for local testing? It's fine for me but I think it would be nice to put this information as comment in the beginning of the file if you decide to keep it.

RobertDrazkowskiGL: I will add such comments (there is another similar file).

src/providers/cryptoauthlib/key_slot.rs
RobertDrazkowskiGL: Thanks.  I did not find any converter of a binary private key into this format. I will try harder.

src/providers/cryptoauthlib/mod.rs
hug-dev: Ah I see. I understand that this is another level of security. I'm concerned about the confidentiality of this value though. I would think about two options:
1. We make it clear that encrypting stored keys is only done because required by the crypto auth lib but has no security value and use a dummy value for the write encrypt key (like only `0xFF`, to make it obvious it's not used.).
2. We use this value properly and make the value of that key come from the configuration file with a new field. We are already storing confidential values in the config file so there is precedent.

Is this write key also a way to prevent multiple people accessing the same ECC key on the same ATEC device, as you would need to know it if you want to use the key?

src/providers/cryptoauthlib/key_slot.rs
CBenoit: Hi, just passing by! ECC is on my wishlist for `picky` (I think ASN.1 types would live in `picky-asn1-x509` as well), but I have yet to find the time to actually look into it and implement it (or someone to open a PR).

hug-dev: Thanks a lot @CBenoit! I guess we can try by defining the new `ECPrivateKey` type ourselves and then make a PR to `picky-asn1-x509` if it works well.
@RobertDrazkowskiGL I am happy to help with that!

e2e_tests/tests/per_provider/normal_tests/import_key.rs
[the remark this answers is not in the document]
RobertDrazkowskiGL: Done.

e2e_tests/tests/per_provider/normal_tests/import_key.rs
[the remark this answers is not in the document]
RobertDrazkowskiGL: Done.

src/providers/cryptoauthlib/key_management.rs
```diff
@@ -30,11 +30,17 @@ impl Provider {
...
+                    Err(ResponseStatus::PsaErrorNotSupported)
+                }
             }
-            | Type::EccPublicKey {
+            Type::EccPublicKey {
                 curve_family: EccFamily::SecpR1,
             } => {
-                if attributes.bits == 256 {
+                if attributes.bits == 512 {
                     Ok(rust_cryptoauthlib::KeyType::P256EccKey)
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: Done.

src/providers/cryptoauthlib/key_operations.rs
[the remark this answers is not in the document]
RobertDrazkowskiGL: Corrected.

src/providers/cryptoauthlib/key_operations.rs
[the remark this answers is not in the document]
RobertDrazkowskiGL: Improved.

src/providers/cryptoauthlib/key_slot.rs
RobertDrazkowskiGL: Corrected. About parsing ASN.1 - at the moment we are doing it "manually", i.e. we are calling openssl functions for doing the hard job for us. But of course we plan to submit PRs to picky-asn1, hopefully soon.

src/providers/cryptoauthlib/mod.rs
RobertDrazkowskiGL: The need for keeping this key confidential is debatable in my opinion. I suspect it exists to keep together: the cryptochip (with keys stored inside) and the piece of hardware (computer) it is soldered into.
About the key being a protection against concurrent accesses - well, it can offer kind of (poor) protection, but if the slot is configured to be writable, its contents can be easily overwritten. And if one user configures the write key, the write key is non-permanently stored on chip and another user with the access to this particular i2c bus could use the slots as if they were not encrypted. 
I discussed how can protection be ensured and it needs carefully designed hardware, where there is a dedicated i2c bus, exclusively assigned to parsec user by means of udev settings. Then the parsec user will own everything connected via this bus, but at least nobody else will be able to "disturb".

e2e_tests/tests/per_provider/normal_tests/import_key.rs
RobertDrazkowskiGL: It was just corrected, I mean a proper DER format  is used. As noted before - since no support via picky-asn is available at the moment, the binary DER sequence is being parsed "manually" via openssl calls.

e2e_tests/provider_cfg/cryptoauthlib/config_608a.toml
RobertDrazkowskiGL: Added.

src/providers/cryptoauthlib/key_operations.rs
```diff
@@ -84,4 +87,191 @@ impl Provider {
...
+            }
+        };
+        // There is no need check op.data.len() - ler the extraction functions do this.
+        let key_data = match extract_raw_key(key_attributes.key_type, &op.data) {
+            Ok(raw_key) => raw_key,
+            Err(error) => return Err(error),
+        };
+
+        let atca_error_status = self.device.import_key(key_type, &key_data, slot_id);
+
```
hug-dev, line 119: I think here you could do a;
```rust
key_data.zeroize();
```
after the import to make sure the key data can not be read in memory after that.
Alternatively you can make the `extract_raw_key` function return a `Secret<Vec<u8>>` so that it's done automatically.

src/providers/cryptoauthlib/key_operations.rs
```diff
@@ -84,4 +87,191 @@ impl Provider {
...
+                }
+                Err(_) => Err(ResponseStatus::PsaErrorInvalidArgument),
+            }
+        }
+        _ => Err(ResponseStatus::PsaErrorNotSupported),
+    }
+}
+
+#[test]
+fn test_extract_raw_ecc_private_key() {
```
hug-dev, line 215: Thanks for the tests!

src/providers/cryptoauthlib/mod.rs
hug-dev: Alright, thanks for the explanation! The assumption is that no-one will be able to read/write on the i2c bus where the chip is.

e2e_tests/tests/per_provider/normal_tests/import_key.rs
```diff
@@ -74,18 +89,33 @@ fn example_modulus_1024() -> Vec<u8> {
...
     let mut client = TestClient::new();
     let key_name = String::from("import_key");
     if !client.is_operation_supported(Opcode::PsaImportKey) {
         return Ok(());
     }
+
     client.import_rsa_public_key(key_name, KEY_DATA.to_vec())
 }
 
+#[cfg(feature = "cryptoauthlib-provider")]
```
hug-dev, line 104: Could you please also execute this test on the Mbed Crypto provider to make sure the test data is correct with the PSA specs?
```rust
#[cfg(any(feature = "mbed-crypto-provider", feature = "cryptoauthlib-provider"))]
```

src/providers/cryptoauthlib/key_management.rs
```diff
@@ -30,11 +30,17 @@ impl Provider {
...
+                } else {
+                    Err(ResponseStatus::PsaErrorNotSupported)
+                }
             }
-            | Type::EccPublicKey {
+            Type::EccPublicKey {
                 curve_family: EccFamily::SecpR1,
             } => {
-                if attributes.bits == 256 {
+                if attributes.bits == 512 || attributes.bits == 0 {
```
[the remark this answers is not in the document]
RobertDrazkowskiGL: I have an impression neither value is actually that important. From the interface point of view I would recommend to keep 0 as "do not care at the moment" just like elsewhere and 256 as the hard requirement of the provider.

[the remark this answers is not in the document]
hug-dev: From the Parsec API point of view the two values have to be accepted so that the operation is implemented correctly on the crypto auth lib provider. That is so all providers supporting the same operation share the same behaviour.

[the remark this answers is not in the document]
RobertDrazkowskiGL: If I understand correctly this `attributes.bits == 256 || attributes.bits == 0` is the correct expression.

src/providers/cryptoauthlib/key_operations.rs
RobertDrazkowskiGL: I think changing the return value is the way to go.

e2e_tests/tests/per_provider/normal_tests/import_key.rs
RobertDrazkowskiGL: Done

src/providers/cryptoauthlib/key_management.rs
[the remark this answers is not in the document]
hug-dev: Yes that's correct!

e2e_tests/tests/per_provider/normal_tests/import_key.rs
```diff
@@ -131,18 +141,18 @@ fn create_and_import_rsa_key() -> Result<()> {
...
-    let key_name = String::from("create_and_import_key");
+    let key_name = String::from("create_and_import_ecc_key");
     if !client.is_operation_supported(Opcode::PsaImportKey) {
         return Ok(());
     }
 
     client.generate_ecc_key_pair_secpr1_ecdsa_sha256(key_name.clone())?;
     let status = client
-        .import_ecc_public_secp_r1_ecdsa_sha256_key(key_name, KEY_PAIR_ECC.to_vec())
+        .import_ecc_public_secp_r1_ecdsa_sha256_key(key_name, ECC_KEY_PAIR.to_vec())
```
hug-dev, line 155: I guess it makes no difference here as we check if the key already exists (so that the code does not go as far as parsing it), but I think it should be `ECC_PUBLIC_KEY`.

src/providers/cryptoauthlib/key_operations.rs
```diff
@@ -151,62 +151,39 @@ impl Provider {
                 // Import failed but at least slot was appropriately marked as Free
             }
             Err(error) => {
-                // Things do never get better...
+                // Things never get better...
```
hug-dev: Some improvement there 😄

e2e_tests/tests/per_provider/normal_tests/import_key.rs
```diff
@@ -406,5 +416,5 @@ fn import_ecc_private_key() {
         return;
```
hug-dev, line 416: Can you also execute this test on the Mbed Crypto provider 😄? That way we will be sure that both the formats for the public key and key pair are correct 👌

hug-dev (commented):
Thanks a lot, really great to see the test also working on the Mbed Crypto provider that will help us a lot 👌 
Just added one more request and then I am fine on my side.
I will let @ionut-arm give his review also, so that you don't have to push too many time!

e2e_tests/provider_cfg/cryptoauthlib/config_608a.toml
```diff
@@ -0,0 +1,39 @@
...
+# checks.
+allow_root = true
+
+[listener]
+listener_type = "DomainSocket"
+timeout = 200 # in milliseconds
+socket_path = "/tmp/parsec.sock"
+
+[authenticator]
+auth_type = "Direct"
```
ionut-arm, line 22: Are these files supposed to be used for manual testing/playing around with the service locally, or do you intend to present them to admins as a starting place for configuring the service with an ATECC device? If it's the former then you can disregard this comment, but otherwise I'd say `auth_type = "UnixPeerCredentials"` is a better choice.

e2e_tests/tests/per_provider/normal_tests/import_key.rs
```diff
@@ -238,6 +320,7 @@ fn check_format_import3() -> Result<()> {
     Ok(())
 }
 
+#[cfg(not(feature = "cryptoauthlib-provider"))]
 #[test]
 fn failed_imported_key_should_be_removed() -> Result<()> {
```
ionut-arm, line 320: Instead of skipping the test completely, any chance you could make `public_key` and `attributes` conditional and have it try to import an ECC key for CALib? Or even just modifying the last import call to succeed in the CAL provider

e2e_tests/tests/per_provider/normal_tests/import_key.rs
RobertDrazkowskiGL: Done. And it looks like Mbed Crypto uses just the raw private key octet string from the whole structure.

ionut-arm (commented):
Thanks for the updates - I left a few comments below, overall it looks good

e2e_tests/tests/per_provider/normal_tests/import_key.rs
RobertDrazkowskiGL: Done

e2e_tests/provider_cfg/cryptoauthlib/config_608a.toml
RobertDrazkowskiGL: So far, we do not have such far-reaching plans. These files are here to make our life easier when we run tests.

hug-dev (approved):
Thanks!

Issue #375, opened 2021-03-31
Make the slot_number field optional


hug-dev:
The PKCS 11 provider can use the [`get_all_slots`](https://docs.rs/cryptoki/0.1.1/cryptoki/struct.Pkcs11.html#method.get_all_slots) function to use the first slot returned if the `slot_number` field was not given. It will save some trouble to the admin!

ionut-arm:
What was the difference between that one and `get_slots_with_token`?

hug-dev:
oh yeah I think `get_slots_with_token` should be used.
From the documentation:

>  tokenPresent indicates whether the list obtained includes only those slots with a token present

I am not fully sure of the difference between a slot and a token but I think a token is in a slot when there is actually something to use there.

Sven-bg:
Hi @hug-dev, if you want I can work on it.
Thank you

hug-dev:
Sure! Thank you!

wiktor-k:
> I am not fully sure of the difference between a slot and a token but I think a token is in a slot when there is actually something to use there.

Indeed. My understanding is that slot = smartcard reader, token = smartcard in the reader. For general use `get_slots_with_token` is better. `get_all_slots` is useful only when learning about the actual readers (they have some metadata associated too).

PR #402, opened 2021-04-23
Added Option<Slot> to PKCS 11 Provider constructor

Sven-bg:
Fix the issue #375

### DIFF
--- a/src/providers/pkcs11/mod.rs
+++ b/src/providers/pkcs11/mod.rs
@@ -22,7 +22,8 @@ use parsec_interface::operations::{
 use parsec_interface::requests::{Opcode, ProviderId, ResponseStatus, Result};
 use parsec_interface::secrecy::{ExposeSecret, SecretString};
 use std::collections::HashSet;
-use std::convert::TryInto;
+use std::convert::From;
+use std::convert::TryFrom;
 use std::io::{Error, ErrorKind};
 use std::str::FromStr;
 use std::sync::RwLock;
@@ -383,8 +384,8 @@ impl ProviderBuilder {
     }
 
     /// Specify the slot number used
-    pub fn with_slot_number(mut self, slot_number: u64) -> ProviderBuilder {
-        self.slot_number = Some(slot_number);
+    pub fn with_slot_number(mut self, slot_number: Option<u64>) -> ProviderBuilder {
+        self.slot_number = slot_number;
 
         self
     }
@@ -427,15 +428,6 @@ impl ProviderBuilder {
             "Building a PKCS 11 provider with library \'{}\'",
             library_path
         );
-        let slot_number = self
-            .slot_number
-            .ok_or_else(|| Error::new(ErrorKind::InvalidData, "missing slot number"))?;
-        let slot = slot_number.try_into().or_else(|_| {
-            Err(Error::new(
-                ErrorKind::InvalidData,
-                "cannot convert slot value",
-            ))
-        })?;
 
         let backend = Pkcs11::new(library_path).map_err(|e| {
             format_error!("Error creating a PKCS 11 context", e);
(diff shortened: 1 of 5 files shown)


ionut-arm (commented):
Hi! 👋🏻 

Thanks for the patch, the code looks good, but the "optionality" should be propagated to service spin-up: the slot number should be optional in the [service configuration file](https://github.com/parallaxsecond/parsec/blob/main/config.toml) as well, and have it passed down like that from the `ServiceBuilder`.

I'd also recommend "testing" it, either by removing the slot number from [the config file](https://github.com/parallaxsecond/parsec/blob/main/e2e_tests/provider_cfg/pkcs11/config.toml) we usually use for end-to-end tests, or by creating a new [configuration test](https://github.com/parallaxsecond/parsec/tree/main/e2e_tests/tests/all_providers/config) (the latter is probably better).

Sven-bg:
Hi @ionut-arm if you prefer I can cancel the PR, add the requested changes and make a new one. 
About the testing, I disccused it with @hug-dev and we agreed on applying the patch and later on add some new tests to cover this specific feature, but if you prefer, I can add the tests too.
Thank you for the hints.
Adriano

ionut-arm:
No, keep the PR open and build on top of it/adjust the commits. And leaving the testing for another PR sounds alright!

src/providers/pkcs11/mod.rs
```diff
@@ -74,10 +74,23 @@ impl Provider {
...
         user_pin: Option<SecretString>,
         software_public_operations: bool,
     ) -> Option<Provider> {
+        // If slot_number is not valid take the first available with a valid token
+        let slot_number = match slot_number {
+            Some(i) => i,
+            None => {
+                let slots = backend.get_slots_with_token().ok()?;
+                if !slots.is_empty() {
+                    slots[0]
```
ionut-arm, line 87: Actually, I'm wondering - should we do it this way, grabbing "the first" token, or should we fail if `slots.len() != 1`? My worry is whether the result is deterministic - what if there are two slots and you get them arranged at random? Have to admit, I've not looked at the spec too closely.

ionut-arm: (or, what if a new slot/token is added at some later point and that one becomes the first one? your service will spin up, but you'll essentially lose a lot of keys)

hug-dev: Yes that's a very good point and I think a good suggestion. I checked the PKCS 11 docs and nothing says that the order is deterministic but also because of the other reason you indicated I think it's better to check that there is only one slot to use.

So replacing the anove with:
```diff
-if !slots.is_empty() {
+if slots.len() == 1 {
```

We should make it clear that this behaviour is used in the config file comments.

src/providers/pkcs11/mod.rs
```diff
@@ -441,11 +446,34 @@ impl ProviderBuilder {
...
+                            ErrorKind::InvalidData,
+                            "cannot convert slot value",
+                    ))
+                })?;
+                slot
+            },
+            None => {
+                let slots = backend
+                    .get_slots_with_initialized_token().ok()
+                    .ok_or_else(|| Error::new(ErrorKind::InvalidData, "failed retreiving a valid slot with a initialized token"))?;
```
hug-dev, line 463: You can use `map_err` here in order to keep the information on the error that happened:
```rust
                let slots = backend
                    .get_slots_with_initialized_token()
                    .map_err(|e| {
                        format_error!("Failed retrieving a valid slot with an initialized token", e);
                        Error::new(ErrorKind::InvalidData, "failed retrieving a valid slot with an initialized token")
                    })?;
```

hug-dev (commented):
Thanks, that will be useful! A few comments added.
You might need to `cargo fmt` before sending it again, when https://github.com/parallaxsecond/rust-cryptoki/pull/35 is merged and imported here!

Next step would be to write a test, but that's totally something we can handle ourselves!

PR #295, opened 2020-11-27
Add asymmetric encryption to TS provider

ionut-arm:
This commit adds asymmetric encryption functionality in the TS
provider. It also moves the TS protobuf code in a new module given
that it now contains a bunch of hand written code as well.

Signed-off-by: Ionut Mihalcea <ionut.mihalcea@arm.com>

PR #406, opened 2021-04-29
Update TS submodule and add asym encryption

ionut-arm:
This commit pulls in new changes to the upstream TS project and adds
asymmetric encryption support for the provider.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,7 +66,7 @@ mbed-crypto-provider = ["psa-crypto"]
 pkcs11-provider = ["cryptoki", "picky-asn1-der", "picky-asn1", "picky-asn1-x509", "psa-crypto", "rand"]
 tpm-provider = ["tss-esapi", "picky-asn1-der", "picky-asn1", "picky-asn1-x509", "hex"]
 cryptoauthlib-provider = ["rust-cryptoauthlib"]
-trusted-service-provider = ["mbed-crypto-provider", "bindgen", "prost-build", "prost"]
+trusted-service-provider = ["psa-crypto", "bindgen", "prost-build", "prost"]
 all-providers = ["tpm-provider", "pkcs11-provider", "mbed-crypto-provider", "cryptoauthlib-provider"]
 
 # Authenticators
--- a/e2e_tests/docker_image/parsec-service-test-all.Dockerfile
+++ b/e2e_tests/docker_image/parsec-service-test-all.Dockerfile
@@ -94,15 +94,18 @@ RUN cd nanopb-0.4.4-linux-x86 \
 RUN rm -rf nanopb-0.4.4-linux-x86 nanopb-0.4.4-linux-x86.tar.gz
 
 # Install mock Trusted Services
-RUN git clone https://git.trustedfirmware.org/TS/trusted-services.git --branch main \
+# Setup git config for patching dependencies
+RUN git config --global user.email "some@email.com"
+RUN git config --global user.name "Parsec Team"
+RUN git clone https://git.trustedfirmware.org/TS/trusted-services.git --branch integration \
 	&& cd trusted-services \
-	&& git reset --hard 2fc7e10c7c21e4dafbf63dc9d00dfc2a7a7fddad
+	&& git reset --hard c1cf9120e4ab0b359a27176b079769b9a7e6bb87
 # Install correct python dependencies
 RUN pip3 install -r trusted-services/requirements.txt
 RUN cd trusted-services/deployments/libts/linux-pc/ \
 	&& cmake . \
 	&& make \
-	&& cp libts.so nanopb_install/lib/libprotobuf-nanopb.a mbedcrypto_install/lib/libmbedcrypto.a /usr/local/lib/
+	&& cp libts.so nanopb_install/lib/libprotobuf-nanopb.a mbedtls_install/lib/libmbedcrypto.a /usr/local/lib/
 RUN rm -rf trusted-services
 
 # Create a new token in a new slot. The slot number assigned will be random
--- a/src/providers/trusted_service/context/mod.rs
+++ b/src/providers/trusted_service/context/mod.rs
@@ -10,7 +10,10 @@ use std::ptr::null_mut;
 use std::slice;
 use std::sync::Mutex;
 use ts_binding::*;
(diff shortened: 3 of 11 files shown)


build.rs
```diff
@@ -6,16 +6,22 @@ use std::path::{Path, PathBuf};
 
 #[cfg(feature = "trusted-service-provider")]
 fn generate_ts_bindings(ts_include_dir: String) -> Result<()> {
-    let header = ts_include_dir.clone() + "/service/locator/interface/service_locator.h";
+    let header = ts_include_dir.clone() + "/components/service/locator/interface/service_locator.h";
+    let encoding_header = ts_include_dir.clone() + "/protocols/rpc/common/packed-c/encoding.h";
```
hug-dev, line 10: There is `packed-c` in the path, what is this file for? Aren't we using protobuf?

hug-dev (approved):
Good! I guess #295 can bet put do the "Done" column after this?

build.rs
```diff
@@ -6,16 +6,22 @@ use std::path::{Path, PathBuf};
 
 #[cfg(feature = "trusted-service-provider")]
 fn generate_ts_bindings(ts_include_dir: String) -> Result<()> {
```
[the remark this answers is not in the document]
ionut-arm: It's a bit difficult because the interface isn't stable yet (as can be seen in this patch, actually). And I'd rather pull the TS-specific stuff out in its own crate, to have a proper `-sys` and everything, before generating and storing the mappings.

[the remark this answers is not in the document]
ionut-arm: Though it's true, we could just pre-generate them each time we update the submodule. not sure it's worth it.

build.rs
ionut-arm: We are, but that's the correct file nonetheless - probably a bug. I did notify the TS team.

build.rs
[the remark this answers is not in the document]
hug-dev: Agree with that, I think it would be worth just having an issue mentioning that we will create those crates at some point.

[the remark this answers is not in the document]
hug-dev: #407

ionut-arm:
Looks like the initial issue I was hoping was fixed hasn't been fixed yet, hence the test failures on asym encryption. I'll keep this open and update it once that gets fixed and upstreamed 😬

src/providers/trusted_service/key_management.rs
```diff
@@ -3,13 +3,31 @@
...
+        let _ = max_current_id.store(super::context::PSA_KEY_ID_USER_MAX, Relaxed);
+        error!(
+            "PSA max key ID limit of {} reached",
+            super::context::PSA_KEY_ID_USER_MAX
+        );
+        return Err(ResponseStatus::PsaErrorInsufficientMemory);
+    }
+
+    Ok(new_key_id)
+}
```
ionut-arm, line 30: I replaced the dependency on the Mbed Crypto provider here to decouple the two PSA Crypto APIs we're abstracting over, so we can change them independently in the future. There's a bit of code duplication, but I'm not that worried about one function.

e2e_tests/docker_image/parsec-service-test-all.Dockerfile
```diff
@@ -94,15 +94,18 @@ RUN cd nanopb-0.4.4-linux-x86 \
 RUN rm -rf nanopb-0.4.4-linux-x86 nanopb-0.4.4-linux-x86.tar.gz
 
 # Install mock Trusted Services
-RUN git clone https://git.trustedfirmware.org/TS/trusted-services.git --branch main \
+# Setup git config for patching dependencies
+RUN git config --global user.email "some@email.com"
+RUN git config --global user.name "Parsec Team"
```
hug-dev, line 99: Was it failing before without this?

src/providers/trusted_service/context/mod.rs
```diff
@@ -10,7 +10,10 @@ use std::ptr::null_mut;
 use std::slice;
 use std::sync::Mutex;
 use ts_binding::*;
-use ts_protobuf::{CloseKeyIn, GetOpcode, OpenKeyIn, OpenKeyOut, SetHandle};
+use ts_protobuf::GetOpcode;
+
+// Bindgen fails to generate this constant.
```
hug-dev, line 15: 💀

Cargo.toml
```diff
@@ -66,7 +66,7 @@ mbed-crypto-provider = ["psa-crypto"]
 pkcs11-provider = ["cryptoki", "picky-asn1-der", "picky-asn1", "picky-asn1-x509", "psa-crypto", "rand"]
 tpm-provider = ["tss-esapi", "picky-asn1-der", "picky-asn1", "picky-asn1-x509", "hex"]
 cryptoauthlib-provider = ["rust-cryptoauthlib"]
-trusted-service-provider = ["mbed-crypto-provider", "bindgen", "prost-build", "prost"]
+trusted-service-provider = ["psa-crypto", "bindgen", "prost-build", "prost"]
```
hug-dev, line 69: Where do we actually need `psa-crypto` for this provider?

hug-dev (approved):

e2e_tests/docker_image/parsec-service-test-all.Dockerfile
ionut-arm: Yes, they've added some step in their build process to apply git patches to some dependencies, I think, and this is needed :(

src/providers/trusted_service/context/mod.rs
ionut-arm: 🧞

Cargo.toml
ionut-arm: All over the place - we use the conversions implemented in `psa-crypto` to native types to fill in the protobuf objects. For example, the protobuf contract for `KeyPolicy` contains a `u32` `alg` field to which we can just convert using the PSA crate, because it expects the same values as PSA Crypto would (just in a protobuf wrapper)

hug-dev: Alright, make sense!

ionut-arm:
This PR is not intended to go into the `0.8.0` release as there seems to be an issue with the underlying library. Will be fixed and merged for the next release.

ionut-arm:
Closing this PR in favour of a rebased version, PR to pop up shortly.

PR #414, opened 2021-05-06
Create SECURITY.md

### LINKED ISSUE #398: Set up Github security policy
Disclosure of security issues should happen through a safe channel, and Github allows us to set up a policy for such disclosures. The exact content of the policy and the contact point should be decided under this issue. 

We should look up other CNCF projects and take not of how they deal with this. [Spire](https://github.com/spiffe/spire/security/policy), for example, keeps track of the versions which are still supported for security updates.

Ideally, all repos that have some significant complexity and are not intended for testing should have a security policy.

ionut-arm:
Fixes #398

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security policy
+
+Security is of paramount importance to the Parsec project. We do all we can to identify and fix
+issues, however some problems might slip through the cracks. Any efforts towards responsible
+disclosure of security problems are greatly appreciated and your contributions will be acknowledged.
+
+## Supported versions
+
+Currently only the most recent version of the Parsec service is eligible for patching. This could
+change in the future.
+
+| Version         | Supported |
+|-----------------|-----------|
+| 0.7.0           | ✅       |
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,21 @@
+# Security policy
+
+Security is of paramount importance to the Parsec project. We do all we can to identify and fix
+issues, however some problems might slip through the cracks. Any efforts towards responsible
+disclosure of security problems are greatly appreciated and your contributions will be acknowledged.
+
+## Supported versions
+
+Currently only the most recent version of the Parsec service is eligible for patching. This could
+change once a first stable version (`1.0.0`) is reached.
```
ionut-arm, line 10: Should the first stable version be `0.8.0`?

hug-dev: Ideally they are all stable, as in we should not have done any breaking changes. I guess from `0.8.0`, stabiltiy is going to be checked but that does not mean that it was not stable before.
Maybe just say "This could change in the future"?

hug-dev (approved):

PR #420, opened 2021-05-07
Replace persistence tests with key mappings tests

### LINKED ISSUE #394: Stability: Persistent state (key mappings)
Definition: Old key mappings should still be read correctly by future stable Parsec versions.
Enforcement: The way the KIM persistently stores the mappings need to be backward-compatible in regards with reading/writing.

# Enforcement Check

Mappings between KeyTriple and KeyInfo are persistently stored somewhere. New stable Parsec versions should successfully load old mappings and store current ones (of the same KeyInfoManager).

For the OnDiskKeyInfoManager:
1. base64 encoding of application names are used as directory names
2. provider ID are used as directory names
3. base64 encoding of key names are used as file names
4. key ID type is serialised/deserialised by Serde
5. the Attributes field directly come from the `psa-crypto` crate, (de)serialised by Serde

Stability:
1. The application names should remain representable as UTF-8 strings. `ListClients` and `DeleteClient` are based on that. Stable.
2. Might become a problem if provider ID switch from static numbers (1: MbedCrypto, 2: PKCS 11, 3: TPM, etc) to dynamically allocated IDs (allowing external providers to be dynamically loaded, without prior Parsec knowledge). Providers UUID could be a stable provider reference (but could the same provider be loaded twice? For example two PKCS 11 providers pointing at different PKCS 11 libraries).
3. Stable as key names are UTF-8 strings.
4. If the key ID type is not stable accross Parsec versions, then it won't be possible to load old mappings.
5. If the `Attributes` structure is modified, then deserialisiation will fail for old mappings. The protobuf version could be used for more stability.

4 and 5. If Serde modifies the way it (de)serialise data, then old mappings won't be read as well.

parallaxsecond/parsec#271: if multiple authenticators are to be supported simultaneously, that will also probably impact the mappings, specially the OnDiskKeyInfoManager. The same application name should yield different keys under different authenticators.

# Testing

Different version of the service should successfully load old mappings and store new ones.

From @ionut-arm in #386:

> My initial idea was that we could create another Docker image where, apart from having all the dependencies installed, we also create a number of keys for a number of clients. We then replace the service binary and check that everything still works.

hug-dev:
This commit modifies the Docker container to already contain Parsec keys
created with version 0.7.0. When running on CI, the Parsec service will
try to use those keys and they should still work.
To replicate the test checking that wrong key mappings are deleted, fake
mappings are also introduced.

Fix #394 

The CryptoAuthLib provider was not supporting key management operations at 0.7.0 but can be tested using next version.
The Trusted Service provider should probably be included as well. It should also probably be included in the "all-providers" tests. I can do that before this PR is merged.

Instead of the big `RUN` step, I will `COPY` a script doing the same thing inside the `Dockerfile`. Submitting this PR for initial review/testing on CI.


e2e_tests/docker_image/parsec-service-test-all.Dockerfile
```diff
@@ -105,3 +105,33 @@ RUN useradd -m parsec-client-2
...
+ENV PARSEC_SERVICE_ENDPOINT="unix:/tmp/parsec.sock"
+RUN tpm_server & \
+    sleep 5 \
+    && tpm2_startup -c \
+    && tpm2_takeownership -o tpm_pass \
+    && cd /tmp/create_keys/parsec/e2e_tests \
+    && SLOT_NUMBER=`softhsm2-util --show-slots | head -n2 | tail -n1 | cut -d " " -f 2` \
+    && find . -name "*toml" -not -name "Cargo.toml" -exec sed -i "s/^# slot_number.*$/slot_number = $SLOT_NUMBER/" {} \; \
+    && cd ../ \
+    && ./target/debug/parsec -c e2e_tests/provider_cfg/all/config.toml & \
```
hug-dev, line 124: I was lazy and I did not create the keys for the Trusted Service provider because it was not included in this config file. I will add it.

e2e_tests/tests/per_provider/key_mappings.rs
```diff
@@ -0,0 +1,47 @@
...
+                .unwrap();
+            let plaintext = client
+                .asymmetric_decrypt_message_with_rsapkcs1v15(key.name.clone(), ciphertext)
+                .unwrap();
+            assert_eq!(PLAINTEXT_MESSAGE.to_vec(), plaintext);
+        } else if key.name == "ecc" {
+            let signature = client.sign_with_ecdsa_sha256(key.name.clone(), HASH.to_vec())?;
+            client.verify_with_ecdsa_sha256(key.name.clone(), HASH.to_vec(), signature)?;
+        } else {
+            // error
```
hug-dev, line 39: I will add some more comments to explain what this does but basically if it finds any other keys than `rsa` or `ecc` it means that the fake mapping was not deleted (it's a key named `toto`)

e2e_tests/docker_image/parsec-service-test-all.Dockerfile
ionut-arm: Don't worry about it, I can do it in a different PR

hug-dev:
In the TPM provider, there is an "integrity check failed", see [here](https://github.com/parallaxsecond/parsec/pull/420/checks?check_run_id=2535290734#step:4:1625). Would that be because of the software TPM server that is stopped and restarted?

ionut-arm:
> 
> 
> In the TPM provider, there is an "integrity check failed", see [here](https://github.com/parallaxsecond/parsec/pull/420/checks?check_run_id=2535290734#step:4:1625). Would that be because of the software TPM server that is stopped and restarted?

That's because `ci.sh` does a `tpm2_startup -c -T mssim` every time you want to test the TPM provider, which clears (part of) the state of the software TPM. It ends up not recognizing previously created keys. The solution would be to run it without the `-c` flag, which does the clearing.

I'd also suggest adding a `tpm2_shutdown -T mssim` at the end of `generate-keys.sh` to ensure a smooth transition - otherwise you might have the same issue.

hug-dev:
> 
> 
> > In the TPM provider, there is an "integrity check failed", see [here](https://github.com/parallaxsecond/parsec/pull/420/checks?check_run_id=2535290734#step:4:1625). Would that be because of the software TPM server that is stopped and restarted?
> 
> That's because `ci.sh` does a `tpm2_startup -c -T mssim` every time you want to test the TPM provider, which clears (part of) the state of the software TPM. It ends up not recognizing previously created keys. The solution would be to run it without the `-c` flag, which does the clearing.
> 
> I'd also suggest adding a `tpm2_shutdown -T mssim` at the end of `generate-keys.sh` to ensure a smooth transition - otherwise you might have the same issue.

Thank! Tried that. I also removed the second `take_ownership`, hope that's ok.

e2e_tests/docker_image/generate-keys.sh
```diff
@@ -0,0 +1,58 @@
...
+SLOT_NUMBER=`softhsm2-util --show-slots | head -n2 | tail -n1 | cut -d " " -f 2`
+find . -name "*toml" -not -name "Cargo.toml" -exec sed -i "s/^# slot_number.*$/slot_number = $SLOT_NUMBER/" {} \;
+cd ../
+./target/debug/parsec -c e2e_tests/provider_cfg/all/config.toml &
+
+# We use the Parsec Tool to create one RSA and one ECC key per provider,
+# when it is possible.
+cargo install parsec-tool
+export PARSEC_SERVICE_ENDPOINT="unix:/tmp/parsec.sock"
+/opt/rust/bin/parsec-tool -p 1 create-rsa-key -k rsa
```
ionut-arm, line 35: Just out of curiosity - isn't it possible to just do `parsec-tool ...`, do you need the full path? Maybe we can add that folder, `/opt/rust/bin` to `PATH` for the whole image?

ionut-arm (approved):
Looks good!

ci.sh
```diff
@@ -234,6 +208,18 @@ RUST_BACKTRACE=1 cargo test $FEATURES
 # Removing any mappings left over from integration tests
 rm -rf mappings/
 
+# Add the Docker image's mappings in this Parsec service for the key mappings
+# test.
```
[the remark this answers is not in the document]
hug-dev: I can add a bit more information here. There is information scattered around, more in the Rust test file.

e2e_tests/docker_image/generate-keys.sh
ionut-arm: I don't mind either way - you still have to update the workflow file anyway, but I understand if you don't want to rebuild some of the image

hug-dev:
Updated the registry with the new image and made the final push ⭐

PR #431, opened 2021-05-19
[CryptoAuthLib provider] Implementation of psa_export_public_key operation.

RobertDrazkowskiGL:
Implementation of psa_export_public_key operation.
The changes may seem larger than they actually are because of code layout adjustments:
- contents of file key_operations.rs are moved to key_management.rs (now it looks like in other providers' sources), while the file itself  is removed
- after the above mentioned adjustment, the CAL provider's code needed refactoring in few places

A newer rust-cryptoauthlib is added with support for all known public key export scenarios.


Cargo.toml
```diff
@@ -41,12 +41,12 @@ picky-asn1-x509 = { version = "0.4.0", optional = true }
...
 libc = "0.2.86"
 anyhow = "1.0.38"
-rust-cryptoauthlib = { version = "0.3.0", optional = true }
+rust-cryptoauthlib = { version = "0.3.1", optional = true }
 prost = { version = "0.7.0", optional = true }
 
 [dev-dependencies]
 rand = { version = "0.8.3", features = ["small_rng"] }
-rust-cryptoauthlib = { version = "0.3.0", features=["software-backend"]}
+rust-cryptoauthlib = { version = "0.3.1", features=["software-backend"]}
```
hug-dev, line 49: Just my curiosity but is this needed if `rust-cryptoauthlib` is already a normal dependency?

e2e_tests/tests/per_provider/normal_tests/export_public_key.rs
```diff
@@ -1,17 +1,21 @@
...
 use parsec_client::core::interface::requests::Opcode;
 use parsec_client::core::interface::requests::ResponseStatus;
 use parsec_client::core::interface::requests::Result;
+#[cfg(not(feature = "cryptoauthlib-provider"))]
 use picky_asn1_x509::RSAPublicKey;
 
+#[cfg(not(feature = "cryptoauthlib-provider"))]
 #[test]
-fn export_public_key() -> Result<()> {
+fn export_rsa_public_key() -> Result<()> {
```
hug-dev, line 16: Thanks for renaming those tests 👌

hug-dev (approved):
Thanks!!

Cargo.toml
RobertDrazkowskiGL: We assume the version 0.3.1 is the first supporting both scenarios of exporting public key (first from a slot with a private key, second directly from the slot with public key). Assuming that after 0.3.1 became an official crate it started being preferred anyway, this version bump is probably overzealous.

e2e_tests/tests/per_provider/normal_tests/export_public_key.rs
RobertDrazkowskiGL: I have found this makes debugging much easier.

Cargo.toml
hug-dev: Ah sorry I realized my sentence was not clear. I meant: is it needed to have `rust-cryptoauthlib` written as a dependency in both `[dependencies]` and `[dev-dependencies]` sections of the `Cargo.toml` file?

hug-dev: Ah right I missed that! Thanks for the explanation, that makes sense

ionut-arm (approved):

PR #434, opened 2021-05-19
Enable coverage testing for TS provider

ionut-arm:
This commit enables collection of code coverage results for the TS
provider, since current changes in the way testing is done means there
is no service restarting, which was triggering the problems before.
The nightly run manual toggle also has an added field, allowing the
tests to be run against specific revision numbers.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -5,6 +5,11 @@ on:
     # Every night at midnight
     - cron: "0 0 * * *"
   workflow_dispatch:
+    inputs:
+      rev:
+        description: "Revision hash to run against"
+        required: false
+        default: ""
 
 jobs:
   dependencies:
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -5,13 +5,20 @@ on:
     # Every night at midnight
     - cron: "0 0 * * *"
   workflow_dispatch:
+    inputs:
+      rev:
+        description: "Revision hash to run against"
+        required: false
+        default: ""
```
hug-dev, line 12: Is it to manually select a specific commit to test?

hug-dev (approved):
😎

.github/workflows/nightly.yml
ionut-arm: Indeed - because if I want to manually test this change, for example, I have to wait until it gets merged.... not great.

Issue #392, opened 2021-04-14
Stability: CLI invocation


hug-dev:
Definition: Old CLI invocations of Parsec should still work with for future stable Parsec versions.
Enforcement: CLI flags should not be removed. New flags should be optional.

# Enforcement Check

The default value of the optional flags should also not change.

This can only be done through testing and code review. We should put some warnings around [that part of the code](https://github.com/parallaxsecond/parsec/blob/1fdcbb4b1ed2701b012ec7d0712114970921d729/src/bin/main.rs#L68) to also make sure it's not accidentally removed.

# Testing

We should have tests that use the CLI options making sure it works and make sure that the defaults stay the same.

hug-dev:
Testing is the best way to enforce it. Each time we add a new option, we should add a test for it. Those tests should be executed for each PR.

hug-dev:
We can separate the code handling CLI options in its own file and have some checks on the CI whenever this file is modified. Taht way we can make sure new tests are created when new CLI options are added and to prevent breaking changes to be merged. That would be on top of test of course.
We can use the code owners feature of GitHub to prevent automatic merging.

This trick can also be used for the configuration file, the OS signals and the systemd integration.

PR #441, opened 2021-05-26
Move CLI log into its own file

hug-dev:
That is to prevent breaking updates to the CLI options. The file is also
added to the GitHub code owners.


ionut-arm (approved):
💯

hug-dev:
Related with #392

Issue #408, opened 2021-04-29
Stability: Build toolchain


hug-dev:
Definition: Parsec build should not break for defined versions of the Rust compiler
Enforcement: Each commit is tested with specific supported versions of the Rust compiler.

# Enforcement Check

We could adopt the same process as in the [`tss-esapi`](https://github.com/parallaxsecond/rust-tss-esapi#minimum-supported-rust-version-msrv) crate where some versions of Rust are used to build:
- the latest `stable`
- the most recent one in Fedora
- we can get request to add requirement for more

This could be enforced in our various repos as well to catch build failures earlier.

# Testing

Regular CI tests

ionut-arm:
What about `clippy`/linting issues? I think we should reserve the right to break/ignore new lint rules indefinitely. This is particularly important if we want to maintain multiple versions in parallel (see that latest clippy lints complaining about naming rules).

hug-dev:
> I think we should reserve the right to break/ignore new lint rules indefinitely.

Yes I think so too, ultimately `clippy` only helps us writing better code but that should be fine if we ignore some rules if they break other stability rules.

hug-dev:
Reminder that this need to be added in the book as well.

hug-dev:
Also thinking that the `--features` flag should have the same behaviour than the CLI options: they should not change in a breaking way.

Issue #397, opened 2021-04-14
Stability: Dynamic libraries dependencies


hug-dev:
Definition: Linking and loading the same versions of libraries should still work in future stable versions of Parsec.
Enforcement: Stable versions of Parsec should not require newer dependencies.

# Enforcement Check

A new stable version of Parsec should not introduce a requirement on a new library dependency. The dynamic libraries dependencies should be listed (with their working versions). The only ones I can think of now are the TSS libraries.

Dynamically loaded libraries like PKCS 11 implementations should also be looked at.

# Testing

For testing, different versions of Parsec should be checked with the same set of supported dynamic libraries versions.

hug-dev:
I think build-time dependencies like `libclang` should also be considered here as distributions often have the requirement that Parsec needs to be build locally, without internet connection. So that an upgrade to Parsec also means upgrade the needed dependencies to build it.

All dependencies (other than the ones locally compiled) should be listed with their working versions. If a new Parsec version requires a new version for one of those dependencies then it could be considered as a breaking change for the environment and should be communicated.

The version of those dependencies should be clearly set and the CI should break if a new version is required.

hug-dev:
Looking at our various dependencies I noted down the important ones. There are of course other dependencies but they should be broadbly available where Parsec is running.

- cryptoki: dynamically loading a pkcs11 implementation at version 2.40
- bindgen: libclang version 6.0 or later
- pkg-config (no specific version requirement)
- parsec-interface (for targets other than x86/Aarch64): protoc (no specific version requirement)
- tss-esapi: TSS libraries version 2.3.3 or later
- psa-crypto: when dynamically linking to Mbed Crypto, Mbed Crypto from Mbed TLS version 2.25.0 or later
- psa-crypto: Python 3, C99 toolchain
- CMake
- Trusted Service library (no specific version requirement at this stage)

I think the ones we should note in the book and test are:
- version 2.40 of the PKCS11 interface (currently checked via testing through SoftHSM)
- libclang version 6.0. In our CI we use the default `libclang-dev` from Ubuntu 18.04 which [happens to be](https://packages.ubuntu.com/bionic/libclang-dev) `libclang` 6.0. We should probably pin it to `libclang-6.0-dev` instead.
- TSS libraries version 2.3.3: currently checked in CI
- Mbed TLS version 2.25.0. We compile Mbed Crypto from source at every build. Instead we should build it at version 2.25.0 on CI to make sure it still works in the future.

Issue #409, opened 2021-04-30
Unable to build parsec 0.7.2 with rust 1.43.1. Parsec 0.6.0 builds fine.


ggardet:
Unable to build parsec 0.7.2 with rust 1.43.1. Parsec 0.6.0 builds fine.

The error is:
```
[  678s]    Compiling cryptoki v0.1.0
[  692s] error[E0277]: the trait bound `std::vec::Vec<u8>: std::convert::From<[u8; 4]>` is not satisfied
[  692s]   --> src/providers/pkcs11/key_management.rs:30:47
[  692s]    |
[  692s] 30 |         let mut template = vec![Attribute::Id(key_id.to_be_bytes().into())];
[  692s]    |                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::convert::From<[u8; 4]>` is not implemented for `std::vec::Vec<u8>`
[  692s]    |
[  692s]    = help: the following implementations were found:
[  692s]              <std::vec::Vec<T> as std::convert::From<&[T]>>
[  692s]              <std::vec::Vec<T> as std::convert::From<&mut [T]>>
[  692s]              <std::vec::Vec<T> as std::convert::From<std::borrow::Cow<'a, [T]>>>
[  692s]              <std::vec::Vec<T> as std::convert::From<std::boxed::Box<[T]>>>
[  692s]            and 5 others
[  692s]    = note: required because of the requirements on the impl of `std::convert::Into<std::vec::Vec<u8>>` for `[u8; 4]`
[  692s] 
[  692s] error[E0277]: the trait bound `std::vec::Vec<u8>: std::convert::From<[u8; 4]>` is not satisfied
[  692s]    --> src/providers/pkcs11/key_management.rs:106:27
[  692s]     |
[  692s] 106 |             Attribute::Id(key_id.to_be_bytes().into()),
[  692s]     |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::convert::From<[u8; 4]>` is not implemented for `std::vec::Vec<u8>`
[  692s]     |
[  692s]     = help: the following implementations were found:
[  692s]               <std::vec::Vec<T> as std::convert::From<&[T]>>
[  692s]               <std::vec::Vec<T> as std::convert::From<&mut [T]>>
[  692s]               <std::vec::Vec<T> as std::convert::From<std::borrow::Cow<'a, [T]>>>
[...]

PR #445, opened 2021-06-03
Execute e2e tests with an old version of client

hug-dev:
This is to ensure that the Parsec service does not break for old
clients.
This commit also specifies an exact version of libclang and Mbed Crypto.
Adds build tests for specific versions of the Rust compiler.

See #397 #408

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -56,6 +56,9 @@ wait_for_service() {
     done
 
     sleep 5
+
+    # Check that Parsec successfully started and is running
+    pgrep -f target/release/parsec >/dev/null
 }
 
 stop_service() {
@@ -77,14 +80,23 @@ run_normal_tests() {
     RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml normal_tests
 }
 
+run_old_e2e_tests() {
+    # The old client fails if ListProviders returns a ProviderID it does not know.
+    # See https://github.com/parallaxsecond/parsec-interface-rs/issues/111
+    if [ "$PROVIDER_NAME" = "pkcs11" ] || [ "$PROVIDER_NAME" = "mbed-crypto" ] || [ "$PROVIDER_NAME" = "tpm" ]; then
+        echo "Execute old end-to-end normal tests"
+        # The version of the Parsec client used in those old tests expect the socket to be at
+        # /tmp/parsec/parsec.sock. This can not be created in the Dockerfile as this is where
+        # the repository is checked out.
+        ln -s /tmp/parsec.sock /tmp/parsec/parsec.sock
+        RUST_BACKTRACE=1 cargo test --manifest-path /tmp/old_e2e_tests/Cargo.toml normal_tests
+    fi
+}
+
 run_key_mappings_tests() {
     echo "Execute key mappings tests"
     RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml key_mappings
 }
-
-# During end-to-end tests, Parsec is configured with the socket in /tmp/
-# Individual tests might change that, but set the default after.
-export PARSEC_SERVICE_ENDPOINT="unix:/tmp/parsec.sock"
 
 # Parse arguments
 NO_CARGO_CLEAN=
@@ -156,6 +168,15 @@ if [ "$PROVIDER_NAME" = "trusted-service" ] || [ "$PROVIDER_NAME" = "coverage" ]
     git submodule update --init
 fi
 
+if [ "$PROVIDER_NAME" = "mbed-crypto" ]; then
+    # With those variables defined, dynamic linking will be attempted to build the
+    # Mbed Crypto provider. The Mbed Crypto library was installed in the container.
+    # Defining those variables during the trusted-service provider testing leads to a
+    # linking problem as it also used Mbed Crypto.
+    export MBEDTLS_LIB_DIR="/tmp/mbedtls/library"
+    export MBEDTLS_INCLUDE_DIR="/tmp/mbedtls/include"
+fi
+
 if [ "$PROVIDER_NAME" = "coverage" ]; then
     PROVIDERS="mbed-crypto tpm pkcs11" # trusted-service not supported because of a segfault when the service stops; see: https://github.com/parallaxsecond/parsec/issues/349
     EXCLUDES="fuzz/*,e2e_tests/*,src/providers/cryptoauthlib/*,src/providers/trusted_service/*"
@@ -183,6 +204,7 @@ if [ "$PROVIDER_NAME" = "coverage" ]; then
 
         # Run tests
         run_normal_tests
+        run_old_e2e_tests
         run_key_mappings_tests
         stop_service
     done
@@ -195,6 +217,40 @@ if [ "$PROVIDER_NAME" = "coverage" ]; then
 fi
 
 echo "Build test"
+
+if [ "$PROVIDER_NAME" = "all" ]; then
+    # We test that everything in the service still builds with the current Rust stable
+    # and an old Rust compiler.
+    # The old Rust compiler version is found by manually checking the oldest Rust version of all
+    # Linux distributions that we support:
+    # - Fedora 33 and more recent releases
+    # - RHEL-8
+    # - openSUSE Tumbleweed
+    # - openSUSE Leap 15.3
+    # The oldest is currently in openSUSE Leap 15.3 and is 1.43.0.
+    rustup update
+
+    rustup toolchain install 1.43.0
+    RUST_BACKTRACE=1 cargo +1.43.0 check --release $FEATURES
+
+    # Latest stable
+    rustup toolchain install stable
+    RUST_BACKTRACE=1 cargo +stable check --release $FEATURES
+
+    # We test that each feature still exist.
+    RUST_BACKTRACE=1 cargo check
+    RUST_BACKTRACE=1 cargo check --features="mbed-crypto-provider"
+    RUST_BACKTRACE=1 cargo check --features="pkcs11-provider"
+    RUST_BACKTRACE=1 cargo check --features="tpm-provider"
+    RUST_BACKTRACE=1 cargo check --features="cryptoauthlib-provider"
+    # To be added when trusted-service is added to all-providers feature
+    #RUST_BACKTRACE=1 cargo check --features="trusted-service-provider"
+    RUST_BACKTRACE=1 cargo check --features="all-providers"
+    RUST_BACKTRACE=1 cargo check --features="direct-authenticator"
+    RUST_BACKTRACE=1 cargo check --features="unix-peer-credentials-authenticator"
+    RUST_BACKTRACE=1 cargo check --features="all-authenticators"
+fi
+
 RUST_BACKTRACE=1 cargo build $FEATURES
 
 echo "Static checks"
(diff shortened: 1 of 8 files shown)


e2e_tests/docker_image/import-old-e2e-tests.sh
```diff
@@ -0,0 +1,18 @@
...
+# These commands are made to import the oldest version of the end-to-end tests to check that they
+# still work with the current version of the Parsec service.
+
+set -xeuf -o pipefail
+
+git clone https://github.com/parallaxsecond/parsec.git
+cd parsec
+# This commit is the oldest one which still works with current Parsec version.
+# It works with the Rust client version 0.6.0
+git checkout 2fee72fc64871472edf141906bf7f55bd59a2f8d
```
ionut-arm, line 15: Maybe we should tag that commit somehow - `client-stability-anchor` or smth? ⚓

ci.sh
```diff
@@ -77,15 +77,20 @@ run_normal_tests() {
     RUST_BACKTRACE=1 cargo test $TEST_FEATURES --manifest-path ./e2e_tests/Cargo.toml normal_tests
 }
 
+run_old_e2e_tests() {
+    echo "Execute old end-to-end normal tests"
+    # The version of the Parsec client used in those old tests expect the socket to be at
+    # /tmp/parsec/parsec.sock. This can not be created in the Dockerfile as this is where
+    # the repository is checked out.
+    ln -s /tmp/parsec.sock /tmp/parsec/parsec.sock
+    RUST_BACKTRACE=1 cargo test --manifest-path /tmp/old_e2e_tests/Cargo.toml normal_tests
```
ionut-arm, line 86: Could you maybe raise an issue to identify a subset of tests that we could run (or create new ones just for this)? It's fine for now to run all of them, but really most of our tests verify functionality in the service, we probably don't need all of them here. The stability of that service functionality is tested in other places anyway.

ionut-arm (approved):
LGTM.

I'm wondering if it would be a good idea to move some of the tests we do per-PR into a weekly run. It seems overkill to do all this testing for each PR, especially since the project is moving somewhat slowly and even if something is broken for a few days it doesn't really matter - people would be using released versions anyway.

e2e_tests/docker_image/import-old-e2e-tests.sh
hug-dev: It's a good idea but I'm worried that the tag would be on the Parsec repository, next to all the release tags :/ Ideally it should be on the Rust client that is used

ci.sh
hug-dev: Yes, I understand your point. I mainly execute `normal_tests` for simplicity here instead of naming the subset of tests, since it takes less than two seconds to execute all of them anyway.

ionut-arm: Hmm, this might sound strange, but: can you compile them in the Dockerfile after copying? 😬

e2e_tests/docker_image/import-old-e2e-tests.sh
ionut-arm: Yeah, fair, though it's only a tag, not a Release.

hug-dev:
> I'm wondering if it would be a good idea to move some of the tests we do per-PR into a weekly run. It seems overkill to do all this testing for each PR, especially since the project is moving somewhat slowly and even if something is broken for a few days it doesn't really matter - people would be using released versions anyway.

Agree, that would be better to do the build tests in nightly for example. I think the tests using the old client are important to do per PR to make sure you do break stability when you import a new version of the interface.
I did that for simplicity as well, as doing it nightly would mean adding a new CLI option to the script somehow 😬

ci.sh
hug-dev: Yeah I guess I could! For the per provider tests, the Rust compiler should not change at "run time"

ci.sh
```diff
@@ -274,8 +326,10 @@ else
         echo "Start Parsec for stress tests"
         # Change the log level for the stress tests because logging is limited on the
         # CI servers.
-        RUST_LOG=error RUST_BACKTRACE=1 cargo run $FEATURES -- --config $CONFIG_PATH &
-        sleep 5
+        RUST_LOG=error RUST_BACKTRACE=1 cargo run --release $FEATURES -- --config $CONFIG_PATH &
+        wait_for_service
+        # Check that Parsec successfully started and is running
+        pgrep -f target/release/parsec >/dev/null
```
ionut-arm, line 332: Should this be included in `wait_for_service`?

ci.sh
```diff
@@ -195,6 +214,38 @@ if [ "$PROVIDER_NAME" = "coverage" ]; then
...
+    rustup toolchain install 1.46.0
+    RUST_BACKTRACE=1 cargo +1.46.0 check --release $FEATURES
+
+    # Latest version in openSUSE Leap 15.3
+    rustup toolchain install 1.43.1
+    RUST_BACKTRACE=1 cargo +1.43.1 check --release $FEATURES
+
+    # Latest stable
+    rustup toolchain install stable
+    RUST_BACKTRACE=1 cargo +stable check --release $FEATURES
```
ionut-arm, line 233: Are we sure we want to do this? We could run into problems like the one with `picky-asn1`. It's not like someone using Fedora or openSUSE couldn't get the latest Rust compiler with `rustup`...

src/providers/tpm/utils.rs
```diff
@@ -105,19 +105,23 @@ pub fn parsec_to_tpm_params(attributes: Attributes) -> Result<KeyParams> {
                 x @ 1024 | x @ 2048 | x @ 3072 | x @ 4096 => x.try_into().unwrap(), // will not fail on the matched values
                 _ => return Err(ResponseStatus::PsaErrorInvalidArgument),
             };
-            if attributes.is_encrypt_permitted() || attributes.is_decrypt_permitted() {
-                Ok(KeyParams::RsaEncrypt {
+            match attributes.policy.permitted_algorithms {
+                Algorithm::AsymmetricSignature(_) => Ok(KeyParams::RsaSign {
```
ionut-arm, line 109: I'm not sure what would happen if this was `Algorithm::AsymmetricSignature(AsymmetricSignature::Ecdsa { ... }`. You'll definitely get an error somewhere, but I'm not sure it'll be `PsaErrorInvalidParameter`. Do we have those sort of tests? Maybe we should 😬

ionut-arm (approved):
Overall looks good to me, a few comments below but nothing major.

ci.sh
hug-dev: Yes!

src/providers/tpm/utils.rs
hug-dev: What about:
```rust
Algorithm::AsymmetricSignature(alg) if alg.is_rsa() => ...
```
it's already in [`psa-crypto`](https://docs.rs/psa-crypto/0.8.0/psa_crypto/types/algorithm/enum.AsymmetricSignature.html#method.is_rsa_alg) 🤯 

We have [one test](https://github.com/parallaxsecond/parsec/blob/b6b73160498e52cbfb3527f5120036c729f91920/e2e_tests/tests/per_provider/normal_tests/key_attributes.rs#L16) which is a bit like that but is currently ignored 🛩️ Should maybe add more at some point.

ci.sh
hug-dev: This is basically to avoid things like https://github.com/parallaxsecond/parsec/issues/409 happening again 😕 I think people there might just want to use the version available without internet access.

src/providers/tpm/utils.rs
ionut-arm: Yeah, that's pretty neat!!

ci.sh
hug-dev: @ggardet @nullr0ute

In this PR we add build tests on CI with specific versions of the Rust compiler we need to track to make sure that Parsec compiles on Fedora and openSUSE.

I wanted to ask you where can we find the latest version of Rust that we should test?

For openSUSE, I used the comment on the issue linked above, is there a place where we can check where the Rust version has been updated?
For Fedora, I used the "base" version of Fedora 33 [here](https://rpms.remirepo.net/rpmphp/zoom.php?rpm=rust), not sure if it's a good ressource. Would the "updates" one be fine?

In other PRs we might introduce changes that would break on old versions of Rust so we would like to make sure that we track the *latest* version of Rust that you need to package Parsec in the distributions!

nullr0ute: So Fedora pretty actively updates the rust toolchain in all actively maintained releases, for example we have 1.52.1 across all releases currently. RHEL-8 currently has 1.49.0.

To test this in a container with Fedora you can do:
FROM fedora:latest
RUN dnf install -y cargo

For RHEL you can do:
FROM redhat/ubi8
RUN yum install -y cargo

And that will give you the latest stable containers with rust/cargo installed

ggardet: For openSUSE, you can track:
* Tumbleweed (the rolling release): https://build.opensuse.org/package/show/openSUSE:Factory/rust which is at `1.51.0` atm
* Leap 15.3 (the current stable release): https://build.opensuse.org/package/show/openSUSE:Leap:15.3/rust which is at `1.43.0`

You can of course use containers for openSUSE Tumbleweed or openSUSE Leap 15.3 for x86_64 and/or aarch64.

ionut-arm: I think the question isn't "what release of the compiler is available, packaged, for the distributions?", but "what release of the compiler is used for generating Rust-based packages?" The numbers could very well be the same, but I'm not sure the initial phrasing of "we need to track to make sure that Parsec compiles on Fedora and openSUSE" was accurate.

hug-dev: As the Rust compiler is meant to be stable, I think it's enough to try the compilation with the oldest Rust version used to compile Parsec in the distributions we support. In that case I think it would be `1.43.0`.

We can manually update that from time to time.

src/providers/tpm/utils.rs
```diff
@@ -105,19 +105,23 @@ pub fn parsec_to_tpm_params(attributes: Attributes) -> Result<KeyParams> {
...
-                })
-            } else {
-                Err(ResponseStatus::PsaErrorNotSupported)
+                }),
+                alg => {
+                    error!(
+                        "Permitted algorithm {:?} not supported with RSA key pair.",
+                        alg
+                    );
+                    Err(ResponseStatus::PsaErrorNotSupported)
```
ionut-arm, line 123: I know this wasn't really supposed to be part of the PR anyway, but this should actually be `PsaErrorInvalidArgument` 😬 

> Implementations should return this error code when an enumeration parameter such as a key type, algorithm, etc. is not recognized. If a combination of parameters is recognized and identified as not valid, return PSA_ERROR_INVALID_ARGUMENT instead.

ci.sh
```diff
@@ -195,6 +217,40 @@ if [ "$PROVIDER_NAME" = "coverage" ]; then
 fi
 
 echo "Build test"
+
+if [ "$PROVIDER_NAME" = "all" ]; then
```
ionut-arm, line 221: Could this be split into a different CI job? It seems to almost double the `all-providers` job to 17m, whereas the others finish in under 10 mins (not taking the Docker image build time into consideration, of course)

hug-dev: I think this is a good idea but will need some more logic in the `ci.sh` script to be able to isolate that code with a new parameter. At the moment, CI execution time is not a big problem so I think this is fine creating an issue for this and doing it as part of another PR?

hug-dev: Created #447

hug-dev:
Published the new image in the GitHub registry.

ionut-arm (approved):
🚀

Issue #421, opened 2021-05-10
Support ECC signing keys in the PKCS#11 provider


paulhowardarm:
### Summary
Add ECC signing key support to the PKCS#11 provider for feature parity with TPM, Mbed, Trusted Service and CryptoAuthLib providers.

### Details
ECC keys to be supported by the following API operations

- `PsaImportKey`
- `PsaGenerateKey`
- `PsaDestroyKey`
- `PsaExportPublicKey`
- `PsaSignHash`
- `PsaVerifyHash`

Curve family support may be subject to some restrictions - aim for feature parity with TPM 2.0 for now.

ionut-arm:
#446 covered the following:

* `PsaGenerateKey`
* `PsaDestroyKey`
* `PsaSignHash`
* `PsaVerifyHash`

The other two will be done in a new PR along with any changes to the shims that allow the (public key) operations to be passed to Mbed Crypto if missing

ionut-arm:
#452 covered:

* `PsaImportKey`
* `PsaExportPublicKey`
* fixing some bugs in sign and verify
* testing on the Nitrokey HSM

Support for ECC keys should now be fully available, at least at the same level as the TPM provider, mostly aimed at SECP R1 curves.

PR #446, opened 2021-06-03
Add ECC functionality to PKCS11 prov

ionut-arm:
This commit adds support for elliptic curves in the PKCS11 provider.
Namely, elliptic curve key pairs can be created and used to sign and
verify signatures. This covers a subset of elliptic curve families and
specific curves within them.

First part of implementing #421


src/providers/tpm/utils.rs
```diff
@@ -172,7 +172,7 @@ fn convert_curve_to_tpm(key_attributes: Attributes) -> Result<EccCurve> {
             224 => Ok(EccCurve::NistP224),
             256 => Ok(EccCurve::NistP256),
             384 => Ok(EccCurve::NistP384),
-            512 => Ok(EccCurve::NistP521),
+            521 => Ok(EccCurve::NistP521),
```
ionut-arm, line 175: This is (was?) a bug, don't know if it's ok to fix it in this PR.

src/providers/pkcs11/key_management.rs
```diff
@@ -92,11 +92,6 @@ impl Provider {
             return Err(ResponseStatus::PsaErrorInvalidArgument);
         }
 
-        if op.attributes.key_type != Type::RsaKeyPair {
-            error!("The PKCS11 provider currently only supports creating RSA key pairs.");
-            return Err(ResponseStatus::PsaErrorNotSupported);
-        }
-
```
hug-dev, line 99: 🌟

src/providers/pkcs11/key_management.rs
```diff
@@ -133,6 +128,17 @@ impl Provider {
                 ));
                 Ok(Mechanism::RsaPkcsKeyPairGen)
             }
+            Type::EccKeyPair { curve_family } => {
+                pub_template.push(Attribute::Private(false.into()));
```
hug-dev, line 132: Maybe the `Private` attribute can be set to false in both cases before the `match` in the public template.

src/providers/pkcs11/key_management.rs
```diff
@@ -133,6 +128,17 @@ impl Provider {
...
+                pub_template.push(Attribute::EcParams(
+                    picky_asn1_der::to_vec(&utils::ec_params(curve_family, key_attributes.bits)?)
+                        .map_err(|e| {
+                        error!("Failed to generate EC parameters: {}", e);
+                        ResponseStatus::PsaErrorGenericError
+                    })?, // This is a constant, so should be safe to unwrap
```
hug-dev, line 138: > `// This is a constant, so should be safe to unwrap`

can delete now?

src/providers/pkcs11/utils.rs
```diff
@@ -119,3 +125,14 @@ pub fn digest_info(alg: AsymmetricSignature, hash: Vec<u8>) -> Result<Vec<u8>> {
...
+pub fn ec_params(ecc_family: EccFamily, bits: usize) -> Result<EcParameters> {
+    Ok(EcParameters::NamedCurve(match (ecc_family, bits) {
+        (EccFamily::SecpR1, 192) => ObjectIdentifierAsn1(oids::secp192r1()),
+        (EccFamily::SecpR1, 224) => ObjectIdentifierAsn1(oids::secp224r1()),
+        (EccFamily::SecpR1, 256) => ObjectIdentifierAsn1(oids::secp256r1()),
+        (EccFamily::SecpR1, 384) => ObjectIdentifierAsn1(oids::secp384r1()),
+        (EccFamily::SecpR1, 521) => ObjectIdentifierAsn1(oids::secp521r1()),
+        _ => return Err(ResponseStatus::PsaErrorNotSupported),
+    }))
+}
```
hug-dev, line 138: Pretty neat 👌

hug-dev (approved):
Looks good to me! Very nice to see all those tests executing on the PKCS11 provider.

Might be worth to update the [operations coverage](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/service_api_coverage.html) once this is merged.

ionut-arm:
The Cross-compilation Docker image might need to be updated, as the Rust compiler version used for cross-compilation is throwing [an error](https://github.com/parallaxsecond/parsec/runs/2745530469?check_suite_focus=true#step:3:503) due to a newly stabilized feature being used in `picky-asn1`. We'll need to decide whether we can expect the compilation issue to vanish when distros are repackaged (because a newer Rustc will be used), or if we instead either: try to get around the problem somehow (either not using the new picky version or finding a way to not include the breaking line?), or we ask our friends at Devolutions to revert that line.

Cargo.toml
```diff
@@ -26,9 +26,9 @@ toml = "0.5.8"
 serde = { version = "1.0.123", features = ["derive"] }
 env_logger = "0.8.3"
 log = { version = "0.4.14", features = ["serde"] }
-cryptoki = { git = "https://github.com/parallaxsecond/rust-cryptoki", rev = "da990ddf7475d0f561033018750e4099020bd475", optional = true, features = ["psa-crypto-conversions"] }
-picky-asn1-der = { version = "0.2.4", optional = true }
-picky-asn1 = { version = "0.3.1", optional = true }
+cryptoki = { git = "https://github.com/parallaxsecond/rust-cryptoki", rev = "916f9bbb208ba8a671c593a36313f09c60cd0255", optional = true, features = ["psa-crypto-conversions"] }
+picky-asn1-der = { version = "<=0.2.4", optional = true }
+picky-asn1 = { version = ">=0.3.1, <=0.3.1", optional = true }
```
hug-dev, line 31: 🤯

e2e_tests/tests/per_provider/normal_tests/asym_sign_verify.rs
```diff
@@ -789,11 +800,6 @@ fn verify_ecc_with_ring() {
     pk.verify(message, &signature).unwrap();
 }
 
-#[cfg(any(
-    feature = "mbed-crypto-provider",
-    feature = "tpm-provider",
-    feature = "cryptoauthlib-provider"
-))]
```
hug-dev, line 796: So nice that it works on all providers now ⭐

src/providers/pkcs11/utils.rs
```diff
@@ -119,3 +126,25 @@ pub fn digest_info(alg: AsymmetricSignature, hash: Vec<u8>) -> Result<Vec<u8>> {
     // should not fail - if it does, there's some error in our stack
     .map_err(|_| ResponseStatus::PsaErrorGenericError)
 }
+
+pub fn ec_params(ecc_family: EccFamily, bits: usize) -> Result<ECParameters> {
+    Ok(ECParameters::NamedCurve(match (ecc_family, bits) {
+        // The following "unwrap()" should be ok, as they cover constant conversions
+        (EccFamily::SecpR1, 192) => {
+            ObjectIdentifierAsn1(String::from("1.2.840.10045.3.1.1").try_into().unwrap())
```
hug-dev, line 134: Isn't there in some of the picky crates, those defined as constants or methods?

hug-dev (approved):

src/providers/pkcs11/utils.rs
ionut-arm: There is - in the ones we can't use because of compiler limitations 😬 I added them recently, so they're in the most recent version only.

Cargo.toml
ionut-arm: Yeah, for some reason just using `<=0.3.1` didn't work....

hug-dev: It's the same but maybe `=0.3.1` would work

src/providers/pkcs11/utils.rs
hug-dev: ah right, that was it! No worries then.

PR #448, opened 2021-06-18
Randomly select the shutdown signal

hug-dev:
This is to test that both signals work in the future. Also adds the main
function to the CODEOWNERS to make sure that the OS interactions stay
stable.


ionut-arm (approved):
Looks good, feel free to merge when the lints are fixed!

PR #451, opened 2021-06-24
Add a way to allow providers to fail instantiation

### LINKED ISSUE #401: Allow providers to be optional or conditional depending on platform feature availability
### Summary
Currently, the Parsec service fails to start up if it cannot instantiate all of the providers specified in configuration. This behaviour is by design. This request is to create a new feature where the service can run with only a subset of the configured providers. This allows configurations to be more portable across different hardware environments.

### Details
In version `0.6.0` of Parsec, the service was tolerant of failed provider instantiations. It would still start successfully as long as at least one provider could be instantiated at start-up time. Any failures were logged and handled. This behaviour could lead to situations where the service was running with a different configuration from what was intended. This was deemed a security problem, and addressed as https://github.com/parallaxsecond/parsec/issues/297 

In version `0.7.0` of Parsec, the service only starts successfully if all configured providers can be instantiated.

There are some legitimate use cases for permitting the service to run with only a subset of the configured providers. For example, one might wish to specify both the TPM provider and the Mbed provider simultaneously, but to only use the TPM where it is available, and fall back on using Mbed (or another back-end) where the TPM is absent. This could be achieved with `0.6.0` but is not possible with `0.7.0` due to the stricter start-up process. No such fall-back mechanism is possible.

It is possible to re-introduce this fall-back mechanism, but it must be done with care and it must be made intentional. We want to minimise the risk that the service could end up running with an unintended configuration or behaviour.

To make this intentional and explicit, Parsec would need to recognise new settings in the configuration file - possibly this could be an optional boolean flag on each provider, defaulting to the strict behaviour of `0.7.0`, but allowing the config author to mark one or more providers as `optional`. An "optional" provider is defined as one that will not block the start-up of the service if it cannot be instantiated.

A careful definition of "cannot be instantiated" is also needed, because it is important to be able to distinguish between an optional provider and a badly-configured one, or one that fails unexpectedly due to some platform or environment error.

It might be the case that not all providers can support the new flag. The Mbed provider probably shouldn't, because that would always be the foundational default case that should always start successfully.

Providers that support the new flag could possibly do so via a preflight check mechanism of some kind. Rather than attempt to start the provider and catch any errors, the provider could inspect the state of the system first. The TPM provider, for example, could check whether the configured TCTI channel is present on the system. 

The PKCS11 provider could possibly check for the presence of the PKCS11 library on disk.

Neither of these preflight checks are bulletproof, because it is conceivable that there is a misconfiguration in either TCTI settings or the PKCS11 library path. But such checks might be good enough.

It remains the case that providers should be specified in preferential order in the configuration. By the time the service is running, Parsec's default provider (for self-configuring clients) will be whichever was the first to be successfully instantiated.

### Definition of Done
Having agreed and implemented the overall mechanism, it should at least be possible to create a Parsec configuration file that specifies both TPM (with `device` TCTI) and Mbed providers, and demonstrate that the TPM provider runs on the TPM-enabled platform, but only the Mbed provider runs on non-TPM platforms. This would restore the specific use case that was identified, which was possible (albeit via risky/buggy behaviour) in `0.6.0`.

hug-dev:
A new optional flag is added to set if a provider can fail to start.

Fix #401

Note that a shortcut is taken here, from #401

> A careful definition of "cannot be instantiated" is also needed, because it is important to be able to distinguish between an optional provider and a badly-configured one, or one that fails unexpectedly due to some platform or environment error.

Here we allow all kind of failures and do not differentiate, mainly for simplicity.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1949,7 +1949,7 @@ dependencies = [
 [[package]]
 name = "tss-esapi"
 version = "5.0.1"
-source = "git+https://github.com/parallaxsecond/rust-tss-esapi?rev=56c487a101dc85e17560416d71f0fc2eb81739a6#56c487a101dc85e17560416d71f0fc2eb81739a6"
+source = "git+https://github.com/parallaxsecond/rust-tss-esapi?rev=1f68655e278b0319c080b9804a7bf3f6e11ff721#1f68655e278b0319c080b9804a7bf3f6e11ff721"
 dependencies = [
  "bitfield",
  "enumflags2",
@@ -1967,7 +1967,7 @@ dependencies = [
 [[package]]
 name = "tss-esapi-sys"
 version = "0.1.1"
-source = "git+https://github.com/parallaxsecond/rust-tss-esapi?rev=56c487a101dc85e17560416d71f0fc2eb81739a6#56c487a101dc85e17560416d71f0fc2eb81739a6"
+source = "git+https://github.com/parallaxsecond/rust-tss-esapi?rev=1f68655e278b0319c080b9804a7bf3f6e11ff721#1f68655e278b0319c080b9804a7bf3f6e11ff721"
 dependencies = [
  "pkg-config",
  "target-lexicon",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ log = { version = "0.4.14", features = ["serde"] }
 cryptoki = { git = "https://github.com/parallaxsecond/rust-cryptoki", rev = "916f9bbb208ba8a671c593a36313f09c60cd0255", optional = true, features = ["psa-crypto-conversions"] }
 picky-asn1-der = { version = "<=0.2.4", optional = true }
 picky-asn1 = { version = ">=0.3.1, <=0.3.1", optional = true }
-tss-esapi = { git = "https://github.com/parallaxsecond/rust-tss-esapi", rev = "56c487a101dc85e17560416d71f0fc2eb81739a6", optional = true }
+tss-esapi = { git = "https://github.com/parallaxsecond/rust-tss-esapi", rev = "1f68655e278b0319c080b9804a7bf3f6e11ff721", optional = true }
 bincode = "1.3.1"
 structopt = "0.3.21"
 derivative = "2.2.0"
--- a/config.toml
+++ b/config.toml
@@ -149,6 +149,8 @@ key_info_manager = "on-disk-manager"
 # e.g. "str:password", or to represent a string version of a hex value, e.g. "hex:1a2b3c". If no prefix is
 # provided, the value is considered to be a string.
 #owner_hierarchy_auth = "password"
+# (Optional) Allows the service to still start without this provider if there is no TPM on the system. The priority list of providers will be as if this provider was commented out.
+#skip_if_no_tpm = false
 
 # Example of a CryptoAuthLib provider configuration
 # All below parameters depend on what devices, interfaces or parameters are required or supported by
--- a/e2e_tests/tests/all_providers/config/mod.rs
+++ b/e2e_tests/tests/all_providers/config/mod.rs
@@ -159,6 +159,25 @@ fn pkcs11_encrypt_software() {
 }
 
 #[test]
+fn no_tpm_support() {
+    set_config("no_tpm_support.toml");
+    // The service should still start, without the TPM provider.
+    reload_service();
+
+    let mut client = TestClient::new();
+    let providers = client.list_providers().unwrap();
+    let uuids: Vec<Uuid> = providers.iter().map(|p| p.uuid).collect();
+    assert_eq!(
+        uuids,
+        vec![
+            Uuid::parse_str("1c1139dc-ad7c-47dc-ad6b-db6fdb466552").unwrap(), // Mbed crypto provider
+            Uuid::parse_str("30e39502-eba6-4d60-a4af-c518b7f5e38f").unwrap(), // Pkcs11 provider
+            Uuid::parse_str("47049873-2a43-4845-9d72-831eab668784").unwrap(), // Core provider
+        ]
+    );
+}
+
+#[test]
 fn various_fields() {
     set_config("various_field_check.toml");
     reload_service();
--- a/e2e_tests/tests/all_providers/config/tomls/no_tpm_support.toml
+++ b/e2e_tests/tests/all_providers/config/tomls/no_tpm_support.toml
@@ -1,0 +1,37 @@
+[core_settings]
+# The CI already timestamps the logs
+log_timestamp = false
+log_error_details = true
+
+[listener]
+listener_type = "DomainSocket"
+timeout = 200 # in milliseconds
+socket_path = "/tmp/parsec.sock"
+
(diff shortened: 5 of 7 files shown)


hug-dev:
I think the difference between "badly-configured" and "failing unexpectedely due to some platform or environment error" is pretty hard to determine in all cases and I am wondering if it is worth it.

ionut-arm:
> 
> 
> I think the difference between "badly-configured" and "failing unexpectedely due to some platform or environment error" is pretty hard to determine in all cases and I am wondering if it is worth it.

I think it's worth at least checking for a few (maybe one or two) cases, even if it's something opinionated. We can document which failures we continue from.

For TPM for example maybe we can check if the TCTI can connect and continue otherwise, and if it can connect, whether the owner authentication value is correct. Everything else and we fail to start.

hug-dev:
> For TPM for example maybe we can check if the TCTI can connect and continue otherwise, and if it can connect, whether the owner authentication value is correct. Everything else and we fail to start.

I guess the only place we could check for those things is when building the TransientKeyContext [here](https://github.com/parallaxsecond/parsec/blob/094f8c119446fa778dfc0dbc0af6102003c47cd0/src/providers/tpm/mod.rs#L319) by checking the different return values?
That would also mean changing the TPM provider builder to add the "conditional" field information and also the return type of the `build` method to a `Result<Option<Provider>>` so that `Ok(None)` can be returned when we are in the allowed failure cases?

I am wondering if all this logic is worth it (code change + documentation + tests) and if it really adds value to be able to have a subset of allowed failure rather than allowing everything.

ionut-arm:
> 
> I guess the only place we could check for those things is when building the TransientKeyContext [here](https://github.com/parallaxsecond/parsec/blob/094f8c119446fa778dfc0dbc0af6102003c47cd0/src/providers/tpm/mod.rs#L319) by checking the different return values?
> That would also mean changing the TPM provider builder to add the "conditional" field information and also the return type of the `build` method to a `Result<Option<Provider>>` so that `Ok(None)` can be returned when we are in the allowed failure cases?

I was thinking of just adding another method to the builder, something like `check_valid() -> Result<()>`. Shouldn't change much of the other flow, but you'd need to probably use the non-abstracted TSS interface.

> 
> I am wondering if all this logic is worth it (code change + documentation + tests) and if it really adds value to be able to have a subset of allowed failure rather than allowing everything.

I guess that's what the initial issue was about. @paulhowardarm can weigh in on that, and maybe the RH folks? Since they asked for the feature it might be worth asking for more clarifications.

paulhowardarm:
I think we need to look at this again - the current PR looks like a simplification too far, because it's just catching errors, which was not the intention. Getting your TPM password wrong is a configuration error. That is not the kind of error that I was expecting we would continue from. That kind of error should still be fatal. The intention (at least in the case of TPM) is that we skip the creation of the TPM provider **only** if there is no TPM on the system.

For the same reason, I think that `can_fail_instantiation` is not a good name for the config flag. It's not about skipping or tolerating failures. It's more about a preflight check for what the platform supports. My suggested name was `conditional_on_platform_support`, although I'm not saying that's a perfect choice either. We could also have `continue_if_not_supported` or some better name. The point is that it's about _support_, not about _failure_. It's about allowing a common config file to work on both TPM and non-TPM platforms.

The simplification that I would be happy with is to only handle TPM for now. Not every provider type can do a pre-flight check in any reliable way, and I would be happy for other providers to reject the config flag (if set to `true`) as being invalid for that provider type. Again, that would be a fatal mis-configuration error - trying, for example, to set the flag to `true` for the PKCS#11 provider or the Mbed provider. (Trusted Service provider is an interesting case that we might want to think about).

For TPM, we can check the availability of the TCTI. If it's a device TCTI, then it should just be a simple file search to see whether the device path is present on the system. If it's a socket TCTI, then perhaps we can check whether the port is listening? To be honest, though, even if we only did the device path check, that might be enough.

ionut-arm:
> 
> For TPM, we can check the availability of the TCTI. If it's a device TCTI, then it should just be a simple file search to see whether the device path is present on the system. If it's a socket TCTI, then perhaps we can check whether the port is listening? To be honest, though, even if we only did the device path check, that might be enough.

But it becomes a bit difficult there, when a device TCTI is used, we simply can't tell the difference between the path not existing because the TPM isn't present and it not existing because the configuration is wrong. But I guess we can just offload that to the admins with some clever wording in our docs.

paulhowardarm:
We should work on the basis that the configuration is correct. The thing we are examining is the platform. I don't want to get into the game of trying to second-guess all possible typos that admins might have put into the config file. It is inevitable that config mistakes might lead to undesirable behaviour of all kinds, including fooling whatever heuristics we might code into the start-up. If the config gives us a TPM device path, then let's assume it is the correct path, but check the platform to see whether it is present.

hug-dev:
Thanks for the explanation, I understand the focus on platform support. I think I will add a method on the [`Tcti`](https://docs.rs/tss-esapi/5.1.0/tss_esapi/enum.Tcti.html) type, `supported`, that will do the file path/socket checks and return an error if there is a problem. I might ignore the TABRMD TCTI for now.
In the service builder, I will try to build the TCTI first and then check if the platform supports it or not.

hug-dev:
Trying the CI with the new solution, with the `skip_if_no_tpm` flag. It is currently using https://github.com/parallaxsecond/rust-tss-esapi/pull/234 on my fork

ionut-arm (approved):
👍🏻

hug-dev:
@paulhowardarm Kindly pinging you for a review, after updates based on the discussion here

paulhowardarm (approved):
Looks good to me!

hug-dev:
Thanks. I also checked the good behaviour of the test on the CI:

```
2021-07-06T17:02:47.0488228Z [INFO  parsec_service::utils::service_builder] Creating a TPM Provider.
2021-07-06T17:02:47.0500545Z ERROR:tcti:src/tss2-tcti/tcti-device.c:440:Tss2_Tcti_Device_Init() Failed to open device file /dev/tpm0: No such file or directory 
2021-07-06T17:02:47.0501747Z WARNING:tcti:src/tss2-tcti/tctildr.c:62:tcti_from_init() TCTI init for function 0x7f081c2decb0 failed with a000a 
2021-07-06T17:02:47.0502865Z WARNING:tcti:src/tss2-tcti/tctildr.c:92:tcti_from_info() Could not initialize TCTI named: tcti-device 
2021-07-06T17:02:47.0507497Z ERROR:tcti:src/tss2-tcti/tctildr-dl.c:150:tcti_from_file() Could not initialize TCTI file: device 
2021-07-06T17:02:47.0509503Z ERROR:tcti:src/tss2-tcti/tctildr.c:418:Tss2_TctiLdr_Initialize_Ex() Failed to instantiate TCTI 
2021-07-06T17:02:47.0510873Z [ERROR tss_esapi::tcti_ldr] Error when creating a TCTI context: response code not recognized
2021-07-06T17:02:47.0513603Z [ERROR parsec_service::utils::service_builder] Error creating a TCTI context; Error: response code not recognized
2021-07-06T17:02:47.0514654Z [WARN  parsec_service::utils::service_builder] Provider TPM provider is skipped.
```

PR #452, opened 2021-06-24
Add import and export support for ECC for PKCS11

ionut-arm:
This commit adds support for importing and exporting public EC keys in
the PKCS11 provider. E2E tests are also enabled, and cross-provider
tests for ECC keys have been added.


src/providers/pkcs11/key_management.rs
```diff
@@ -228,38 +266,41 @@ impl Provider {
...
+        key_data: &[u8],
+        bits: usize,
+        curve_family: EccFamily,
+        template: &mut Vec<Attribute>,
+    ) -> Result<()> {
+        let mut data_in = key_data.to_vec();
+        if data_in.len() > 255 {
+            return Err(ResponseStatus::PsaErrorInvalidArgument);
         }
+        let mut ec_point = vec![0x04, data_in.len() as u8];
```
hug-dev, line 290: The [description of the EC public key format](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_export_public_key.html#description) seems to imply that `0x04` is already included there? Also, is this valid for all curve families? Seems like it is all except Montgomery

edit: ah had another look, is it to have the same format as `[0x04, POINT_LENGTH, 0x04, X, Y]`? Which also explains the `> 255` check above.
Is it the format required by PKCS11? In the export function below you seem to say that there might be multiple.

I think a comment explaining all that would be nice here.

src/providers/pkcs11/key_management.rs
```diff
@@ -228,38 +266,41 @@ impl Provider {
...
+        ec_point.append(&mut data_in);
```
hug-dev, line 291: If `data_in` is copied here, it might be worth to zeroize after

src/providers/pkcs11/key_management.rs
```diff
@@ -305,11 +361,48 @@ impl Provider {
...
+        key_attributes: Attributes,
+    ) -> Result<Vec<u8>> {
+        let mut attributes = session
+            .get_attributes(key, &[AttributeType::EcPoint])
+            .map_err(to_response_status)?;
+
+        if attributes.len() != 1 {
+            error!("Expected to find EC point attribute in public key.");
+            return Err(ResponseStatus::PsaErrorCommunicationFailure);
+        }
```
hug-dev, line 383: Maybe this is already covered in the `else` arm of line 385?

hug-dev (commented):
Thanks 👏 I have a few questions

src/providers/pkcs11/key_management.rs
ionut-arm: Ok, yeah, I'll add a comment. From what I've read it seems those first two bytes are needed, though I tried initially without it and it worked. But I'm afraid some tokens will reject it. I should probs try this with some real hardware... 

And yes, I should probs check which curve is being imported.

src/providers/pkcs11/key_management.rs
ionut-arm: If you call `remove` on a vector and the index is out of bounds, you get a panic ☠️

src/providers/pkcs11/key_management.rs
ionut-arm: Not sure it's really needed, since this can only be a public key that we're handling here, but I'll zeroize at the end, just to be sure.

ionut-arm: Hmm, actually I'd have to start searching in the `template` vector after we're done creating the key for that `EcPoint`, since this `ec_point` is essentially moved into that one. There's similar processing with no zeroizing for the RSA keys, so I'll just let this be, it's not sensitive stuff.

src/providers/pkcs11/key_management.rs
ionut-arm: Changed, I found the proper representation, I just needed to wrap the PSA input data into an ASN.1 Octet String :)

ionut-arm:
NOTE: I've had to make some changes to `sign` and `verify` too, as it seems using the ASN.1 `DigestInfo` structure for ECDSA is a no-no, and the token just assumed that was part of the hash as well. I've also enabled more tests to verify these formatting compatibilities.

ionut-arm:
I've verified this on the Nitrokey HSM2 as well - it doesn't support verification on chip with ECDSA, but signing worked just fine.

src/providers/pkcs11/key_management.rs
```diff
@@ -228,38 +266,56 @@ impl Provider {
...
+                error!("Importing EC keys using Montgomery curves is not currently supported.");
+                return Err(ResponseStatus::PsaErrorNotSupported);
             }
+            _ => (),
         }
+        // The format expected by PKCS11 is an ASN.1 OctetString containing the
+        // data that the PSA Crypto interface specifies.
+        // See ECPoint in [SEC1](https://www.secg.org/sec1-v2.pdf). PKCS11 mandates using
+        // [ANSI X9.62 ECPoint](https://cryptsoft.com/pkcs11doc/v220/group__SEC__12__3__3__ECDSA__PUBLIC__KEY__OBJECTS.html),
+        // however SEC1 is an equivalent spec.
```
hug-dev, line 301: 👌

hug-dev (approved):
Nice and clean ⭐

src/providers/pkcs11/asym_sign.rs
```diff
@@ -32,10 +33,18 @@ impl Provider {
         let key = self.find_key(&session, key_id, KeyPairType::PrivateKey)?;
         info!("Located signing key.");
 
+        let hash = match key_attributes.key_type {
+            // For RSA signatures we need to format the hash into a DigestInfo structure
+            Type::RsaKeyPair => utils::digest_info(op.alg, op.hash.to_vec())?.into(),
```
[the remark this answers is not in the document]
ionut-arm: That should fall under https://github.com/parallaxsecond/parsec-interface-rs/issues/107 , but yes, something we need to consider at some point.

e2e_tests/tests/per_provider/normal_tests/export_public_key.rs
```diff
@@ -1,14 +1,13 @@
 // Copyright 2019 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
-#[cfg(any(feature = "mbed-crypto-provider", feature = "cryptoauthlib-provider"))]
+#![allow(unused_imports, unused)]
```
ionut-arm, line 3: I've swapped the explicit `cfg` for features tagged on inports and constants with this because it was starting to be a mess of figuring out what's used where, and trying to align across tens of tests that are supported by different providers.

hug-dev: Sure, that's totally fine for tests

PR #454, opened 2021-06-28
Make cross-compilation run on release version

ionut-arm:
This commit changes the command for cross compilation building to work
on the release version of the binary instead. This is because a recent
change in the way process liveness is done refers to the release path in
particular.


ionut-arm:
This PR isn't ready yet :( I'll post an update when it's working.

ionut-arm:
OK, seems to work now, see [here](https://github.com/parallaxsecond/parsec/runs/2952632693?check_suite_focus=true). I removed the SPIFFE authenticator from unit test checking. Should I also remove it from the coverage reporting, then?

hug-dev (approved):

PR #465, opened 2021-07-08
Added missing ingress logs to providers. #416

### LINKED ISSUE #416: Fix ingress/egress trace logs
I think we should use the convention:
- have an `ingress` log in the first line of a function
- have an `egress` log on the line just after the function was called

Moreover, [this](https://github.com/parallaxsecond/parsec/blob/ff9b6d1e7fb5899d6e89b9037daae8f2d54b53bf/src/providers/mbed_crypto/mod.rs#L215) is wrong (export_public_key instead of export_key).

I noticed as some in/egress logs are missing.

MattDavis00:
Added to core, mbed_crypto, pkcs11, tpm, and trusted_service providers.
Fixed psa_export_key function being referenced as psa_export_public_key.
Changed some egress logs from `psa_aead_encrypt_egress` format to
`psa_aead_encrypt egress` format.

Closes #416

### DIFF
--- a/src/back/backend_handler.rs
+++ b/src/back/backend_handler.rs
@@ -154,7 +154,7 @@ impl BackEndHandler {
                 let app = unwrap_or_else_return!(app.ok_or(ResponseStatus::NotAuthenticated));
                 let result =
                     unwrap_or_else_return!(self.provider.psa_export_key(app.into(), op_export_key));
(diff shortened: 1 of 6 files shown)


ionut-arm (approved):
👌🏻

hug-dev (approved):
Thank you, that will be useful!

PR #466, opened 2021-07-08
Add a allow_export flag to restrict exporting

### LINKED ISSUE #462: Add PKCS11 provider export-attributes switch
Creating keys in PKCS11 backends involves constructing templates for the keys, which limit or enhance the operations you can perform with them. Two of these attributes control the exportability of private keys - `CKA_SENSITIVE` and `CKA_EXTRACTABLE` - and thus we set these attributes according to the key policy provided by the Parsec client (based on the usage flags).

A problem arises when certain tokens (e.g. the PKCS11 implementation on NXP Layerscape boards) choose to restrict the usage of these attributes, making them read-only to never allow key exporting. Because of how read-only attributes are treated, the token is expected to fail creation if such attributes are ever set during key generation. This deviation from the spec is allowed and there is no way to check whether this is the case for the underlying platform through some discovery mechanism. This does, however, mean that currently we always fail to create asymmetric keys on such platforms. 

A solution _could_ be to remove the attributes from the templates and do the checks in Parsec alone, but this means that platforms that default those attributes to non-exportable, but would allow changes and key exporting, would also get non-exportable keys regardless of what the client specifies in the policy.

Another solution is to have another flag in the Parsec service config which effectively prohibits the service from setting `CKA_SENSITIVE` and `CKA_EXTRACTABLE` for PKCS11 keys. This issue covers the addition of this flag, along with tests that ensure everything still works with the new flag enabled.

hug-dev:
On some PKCS11 implementations, it is not allowed to create a key that
can be exported. For that reason, prevent creating exportable keys from
the configuration file so that clients can have a better return value.

This probably should be written in the book as well, as a new possible return value for generate key.

Fix #462

### DIFF
--- a/src/providers/pkcs11/key_management.rs
+++ b/src/providers/pkcs11/key_management.rs
@@ -178,8 +178,13 @@ impl Provider {
     ) -> Result<psa_import_key::Result> {
         let key_name = op.key_name;
         let key_attributes = op.attributes;
-        let key_triple = KeyTriple::new(app_name, ProviderId::Pkcs11, key_name);
 
+        if key_attributes.policy.usage_flags.export && !self.allow_export {
+            error!("The configuration of this provider does not allow it to generate keys that can be exported.");
+            return Err(ResponseStatus::PsaErrorNotPermitted);
+        }
+
+        let key_triple = KeyTriple::new(app_name, ProviderId::Pkcs11, key_name);
         self.key_info_store.does_not_exist(&key_triple)?;
 
         let session = self.new_session()?;
--- a/src/providers/pkcs11/key_management.rs
+++ b/src/providers/pkcs11/key_management.rs
@@ -95,6 +95,11 @@ impl Provider {
         let key_name = op.key_name;
         let key_attributes = op.attributes;
 
+        if key_attributes.policy.usage_flags.export && !self.allow_export {
+            error!("The configuration of this provider does not allow it to generate keys that can be exported.");
+            return Err(ResponseStatus::PsaErrorNotPermitted);
+        }
+
         let key_triple = KeyTriple::new(app_name, ProviderId::Pkcs11, key_name);
         self.key_info_store.does_not_exist(&key_triple)?;
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
       # run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec -t parsec-service-test-all /tmp/parsec/ci.sh tpm
 
   trusted-service-provider:
-    name: Integration tests using Cypto Trusted Service provider
+    name: Integration tests using Crypto Trusted Service provider
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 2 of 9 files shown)


ionut-arm:
But you're not exactly solving the problem in #462 - you're still setting the attributes for keys that have `export` as false.

hug-dev:
As per the investigation in [here](https://github.com/parallaxsecond/parsec/issues/462#issuecomment-876449201), I checked that those attributes are actually allowed (in Layerscape) but they can't be set to unsecure mode (exportable=true, sensitive=false). It would still work if they explicitely set to the secure mode. That's why I "simplified" what we planned to this.
I believe Yubico will follow the same logic (as per [this](https://github.com/Yubico/yubico-piv-tool/issues/310))

ionut-arm:
Oh, it's because we have a bug 😬 I was wondering wth is going on, because what you're saying and what Sahil reported are at odds - most keys we generate for tests [have `export` set to `false`](https://github.com/parallaxsecond/parsec/blob/main/e2e_tests/src/lib.rs#L139), so it made no sense that `create_and_destroy` would fail if what you said above were true.

However, see if you can spot something wrong with this:
https://github.com/parallaxsecond/parsec/blob/6018d25f072c674de6e6e2192bf18ba730a1baa7/src/providers/pkcs11/utils.rs#L97

wiktor-k:
Folks, if we only patch the entire world of buggy PKCS#11 drivers then this PR could be obsoleted! :partying_face:

ionut-arm:
> 
> 
> Folks, if we only patch the entire world of buggy PKCS#11 drivers then this PR could be obsoleted! 🥳

Unfortunately not, see my comment above ☹️ we set `SENSITIVE` and `EXTRACTABLE` to the same boolean value, when they should be "opposites"

wiktor-k:
Oh, right. That's totally something to improve. And what a coincidence that buggy PKCS#11 drivers exposed a bug in parsec :) Fortunately this one is easy to fix.

hug-dev:
Should I add the fix in this PR?

Would have been funny if you did not spot the [inverted logic](https://github.com/Yubico/yubico-piv-tool/pull/311#discussion_r667847299) and the double bug in Parsec and there made the key secure

ionut-arm:
> 
> 
> Should I add the fix in this PR?

I think so - it's only one line and related to what's going on in here (and it was discovered via the same conversation).

wiktor-k:
> Continuous Integration / Integration tests using Cypto Trusted Service provider

I think there's a typo in the test name that failed ^^ :)

src/providers/pkcs11/key_management.rs
```diff
@@ -95,6 +95,11 @@ impl Provider {
         let key_name = op.key_name;
         let key_attributes = op.attributes;
 
+        if key_attributes.policy.usage_flags.export && !self.allow_export {
+            error!("The configuration of this provider does not allow it to generate keys that can be exported.");
+            return Err(ResponseStatus::PsaErrorNotPermitted);
+        }
```
ionut-arm, line 101: Should this be in `psa_import_key_internal` too? We don't handle private keys for now, though.

ionut-arm (approved):
👍🏻

src/providers/pkcs11/key_management.rs
hug-dev: I think ultimately it should be yes! I did not add it now for that reason.

hug-dev: Well, added it anyway 🌞

PR #467, opened 2021-07-08
Update dependency on Trusted Services

ionut-arm:
This commit updates our dependency on the upstream Trusted Services repo
to include recent commits. The changes include major updates to the
crypto interface, including a move to PSA Crypto v1, meaning a very
different approach to managing keys is needed. Calls to OpenKey and
CloseKey have been scrapped.

Issue #341, opened 2021-02-10
Add support for other cryptographic services in the Trusted Service provider


ionut-arm:
Placeholder issue for adding support for other cryptographic services (i.e. only random number generation, for now) in the Trusted Service provider. If any major changes are needed, please fill in a design proposal below.

The new operations should be tested through E2E tests.

PR #468, opened 2021-07-09
Added psa_export_key & psa_generate_random to TS Provider

MattDavis00:
References #341

Waiting on https://github.com/parallaxsecond/parsec/pull/467 for breaking changes.

### DIFF
--- a/src/providers/trusted_service/context/generate_random.rs
+++ b/src/providers/trusted_service/context/generate_random.rs
@@ -1,0 +1,18 @@
+// Copyright 2021 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use super::error::Error;
+use super::ts_protobuf::{GenerateRandomIn, GenerateRandomOut};
+use super::Context;
+use log::info;
+use std::convert::TryInto;
+
+impl Context {
+    pub fn generate_random(&self, size: usize) -> Result<Vec<u8>, Error> {
+        info!("Handling GenerateRandom request");
+        let open_req: GenerateRandomIn = GenerateRandomIn {
+            size: size.try_into()?,
+        };
+        let result: GenerateRandomOut = self.send_request(&open_req)?;
+        Ok(result.random_bytes)
+    }
+}
--- a/src/providers/trusted_service/context/key_management.rs
+++ b/src/providers/trusted_service/context/key_management.rs
@@ -2,8 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 use super::error::Error;
 use super::ts_protobuf::{
-    DestroyKeyIn, DestroyKeyOut, ExportPublicKeyIn, GenerateKeyIn, ImportKeyIn, KeyAttributes,
-    KeyLifetime, KeyPolicy,
+    DestroyKeyIn, DestroyKeyOut, ExportKeyIn, ExportPublicKeyIn, GenerateKeyIn, ImportKeyIn,
+    KeyAttributes, KeyLifetime, KeyPolicy,
 };
 use super::Context;
 use log::info;
@@ -82,6 +82,13 @@ impl Context {
     pub fn export_public_key(&self, id: u32) -> Result<Vec<u8>, Error> {
         info!("Handling ExportPublicKey request");
         let req = ExportPublicKeyIn { id };
+        self.send_request(&req)
+    }
+
+    /// Export the key given its ID.
+    pub fn export_key(&self, id: u32) -> Result<Vec<u8>, Error> {
+        info!("Handling ExportKey request");
+        let req = ExportKeyIn { id };
         self.send_request(&req)
     }
 
--- a/src/providers/trusted_service/context/ts_protobuf.rs
+++ b/src/providers/trusted_service/context/ts_protobuf.rs
@@ -54,8 +54,16 @@ opcode_impl!(SignHashIn, SignHashOut, SignHash);
 opcode_impl!(VerifyHashIn, VerifyHashOut, VerifyHash);
 opcode_impl!(ImportKeyIn, ImportKeyOut, ImportKey);
 opcode_impl!(ExportPublicKeyIn, ExportPublicKeyOut, ExportPublicKey);
+opcode_impl!(ExportKeyIn, ExportKeyOut, ExportKey);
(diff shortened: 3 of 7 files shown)


src/providers/trusted_service/context/generate_random.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright 2020 Contributors to the Parsec project.
```
ionut-arm, line 1:
```suggestion
// Copyright 2021 Contributors to the Parsec project.
```

src/providers/trusted_service/context/generate_random.rs
```diff
@@ -0,0 +1,20 @@
...
+use super::ts_protobuf::{GenerateRandomIn, GenerateRandomOut};
+use super::Context;
+use log::info;
+use std::convert::TryInto;
+
+impl Context {
+    pub fn generate_random(&self, size: usize) -> Result<Vec<u8>, Error> {
+        info!("Handling GenerateRandom request");
+        let open_req = GenerateRandomIn {
+            size: size.try_into().unwrap(),
```
ionut-arm, line 13: Could you add a comment here about the `.unwrap()`?

src/providers/trusted_service/context/generate_random.rs
```diff
@@ -0,0 +1,20 @@
...
+        };
+        match self.send_request(&open_req) {
+            Ok(GenerateRandomOut { random_bytes }) => Ok(random_bytes),
+            Err(e) => Err(e),
+        }
```
ionut-arm, line 18: I think you could simplify this to something like:
```Rust
Ok(self.send_request(&open_req)?.random_bytes)
```
But this is just a bit of bikeshedding

src/providers/trusted_service/context/ts_protobuf.rs
```diff
@@ -79,6 +81,7 @@ set_handle_impl!(VerifyHashIn);
 set_handle_impl!(AsymmetricEncryptIn);
 set_handle_impl!(AsymmetricDecryptIn);
 set_handle_impl!(ExportPublicKeyIn);
+set_handle_impl!(ExportKeyIn);
 
 impl Drop for ImportKeyIn {
```
ionut-arm, line 86: Could you please implement `Drop` for `ExportKeyOut`, scrubbing away the `data` field? We do this for all operations that handle sensitive data such as "raw" key material. (I'm checking now if we should actually do the same for `GenerateRandomOut`, since that output could be used for sensitive stuff)

src/providers/trusted_service/generate_random.rs
```diff
@@ -0,0 +1,22 @@
+use super::Provider;
```
ionut-arm, line 1: Copyright needed ©️

src/providers/trusted_service/context/key_management.rs
```diff
@@ -91,6 +92,12 @@ impl Context {
         self.send_request_with_key(ExportPublicKeyIn::default(), id)
     }
 
+    /// Export the key given its ID.
+    pub fn export_key(&self, id: u32) -> Result<Vec<u8>, Error> {
+        info!("Handling ExportKey request");
+        self.send_request_with_key(ExportKeyIn::default(), id)
```
ionut-arm, line 98: This is quite cool, didn't realise `prost` is able to do this. Wouldn't it work for `GenerateRandomOut` as well?

ionut-arm (commented):
Thanks!! I left a few comments below

ionut-arm:
You can rebase on top of `main` now that #467 was merged 🚀

src/providers/trusted_service/context/ts_protobuf.rs
MattDavis00: Did you decide if this should also be implemented for `GenerateRandomOut`?

ionut-arm: Ah, yes - yeah, could you add `GenerateRandomOut` there as well? The output isn't really "defined" as sensitive, but it could reasonably be used for quite sensitive stuff.

ionut-arm (approved):
👍🏻

src/providers/trusted_service/context/generate_random.rs
```diff
@@ -0,0 +1,18 @@
...
+use super::ts_protobuf::{GenerateRandomIn, GenerateRandomOut};
+use super::Context;
+use log::info;
+use std::convert::TryInto;
+
+impl Context {
+    pub fn generate_random(&self, size: usize) -> Result<Vec<u8>, Error> {
+        info!("Handling GenerateRandom request");
+        let open_req: GenerateRandomIn = GenerateRandomIn {
+            size: size.try_into().unwrap(), // Can't fail usize->u64 conversion
```
hug-dev, line 13: Couldn't it fail on 32 bits platforms? Looks like `impl From<std::num::TryFromIntError> for Error` is [there](https://github.com/parallaxsecond/parsec/blob/1b6a60f8e09ff15487dac0367e9fd656f60b8191/src/providers/trusted_service/context/error.rs#L61) so you can probably just use `?` instead of `unwrap` here?

hug-dev (approved):
🙆‍♂️

src/providers/trusted_service/context/generate_random.rs
ionut-arm: Not really, if `usize` is 32 bits then extending to `u64` can't fail. You can get an error the other way.

The implementation is [here](https://doc.rust-lang.org/src/core/convert/num.rs.html#351), and that macro is implemented [here](https://doc.rust-lang.org/src/core/convert/num.rs.html#172). For some reason they went with `TryFrom` even though there isn't any possible error.......

ionut-arm: Oh, but looking at that implementation, maybe `size as u64` would be better, like in the `try_from_unbounded` implementation linked above.

hug-dev: Right that is the safe direction here. Could fail on 128 bits platforms 🤡 
I would personally prefer `try_into()?` if that works easily but I'm ok with the `as` if not. Just so that when reading the code you don't have to think about it too much

MattDavis00: Changed to `size: size.try_into()?` if that works

PR #469, opened 2021-07-12
Make KeyInfo a private type Fix #400

### LINKED ISSUE #400: Make `KeyInfo` a private type
[`KeyInfo`](https://docs.rs/parsec-service/0.7.2/parsec_service/key_info_managers/struct.KeyInfo.html)` should only be used internally by the KeyInfoManagerClient. The interface of the KeyInfoManagerClient should only expose types directly useful to the providers.

Currently, [`remove_key_info`](https://docs.rs/parsec-service/0.7.2/parsec_service/key_info_managers/struct.KeyInfoManagerClient.html#method.remove_key_info) returns a `Result<KeyInfo>`. It should either return a tuple `(T, Attributes)` or nothing at all I think.

Kakemone:
Make KeyInfo private and change remove_key_info() to return something of type Result<(Vec<u8>, Attributes)> rather than Result<KeyInfo>

Signed-off-by: Sam Davis <sam.davis@arm.com>

### DIFF
--- a/src/key_info_managers/mod.rs
+++ b/src/key_info_managers/mod.rs
@@ -43,11 +43,11 @@ impl fmt::Display for KeyTriple {
 /// Information stored about a key
 #[derive(Serialize, Deserialize, Debug, PartialEq, Clone, Zeroize)]
 #[zeroize(drop)]
-pub struct KeyInfo {
+struct KeyInfo {
     /// Reference to a key in the Provider
-    pub id: Vec<u8>,
+    id: Vec<u8>,
     /// Attributes of a key
-    pub attributes: Attributes,
+    attributes: Attributes,
 }
 
 impl KeyTriple {
@@ -221,13 +221,13 @@ impl KeyInfoManagerClient {
     pub fn remove_key_info(
         &self,
         key_triple: &KeyTriple,
-    ) -> parsec_interface::requests::Result<KeyInfo> {
+    ) -> parsec_interface::requests::Result<()> {
         let mut key_info_manager_impl = self
             .key_info_manager_impl
             .write()
             .expect("Key Info Manager lock poisoned");
         match key_info_manager_impl.remove(key_triple) {
-            Ok(Some(key_info)) => Ok(key_info),
+            Ok(Some(_key_info)) => Ok(()),
             Ok(None) => Err(ResponseStatus::PsaErrorDoesNotExist),
             Err(string) => Err(to_response_status(string)),
         }


src/key_info_managers/mod.rs
```diff
@@ -43,11 +43,23 @@ impl fmt::Display for KeyTriple {
...
 #[zeroize(drop)]
-pub struct KeyInfo {
+struct KeyInfo {
     /// Reference to a key in the Provider
-    pub id: Vec<u8>,
+    id: Vec<u8>,
     /// Attributes of a key
-    pub attributes: Attributes,
+    attributes: Attributes,
+}
```
ionut-arm, line 51: Hmm, does it matter whether the inner fields of a private type are public? I'm ambivalent on this, tbh, just curious what the argument was

Kakemone: I'm not sure if it matters I just thought that the issue wanted all the parts of the struct to be private but I may have misunderstood. I can change this if necessary and resubmit the PR.

src/key_info_managers/mod.rs
```diff
@@ -221,13 +233,13 @@ impl KeyInfoManagerClient {
     pub fn remove_key_info(
         &self,
         key_triple: &KeyTriple,
-    ) -> parsec_interface::requests::Result<KeyInfo> {
+    ) -> parsec_interface::requests::Result<(Vec<u8>, Attributes)> {
```
hug-dev, line 236: Since nothing broke in the code changing this (everything compiled successfully), I think all callers of this method do:
```rust
let _ = client.remove_key_info(key_triple)?;
```
ignoring the returned value.

For that reason I think it would be fine to make this method return nothing, ie `Result<()>`?

If we wanted to return the key info, it would be best to replace the `Vec<u8>` with the generic deserialised type (ie `T` in the method `get_key_id`). Return nothing seems like a reasonable simplification for now, since this is not used.

src/key_info_managers/mod.rs
hug-dev: I think it's good practice if they are private by default, unless needed in another module?

hug-dev (commented):
Thanks! A suggestion for `remove_key_info`

src/key_info_managers/mod.rs
ionut-arm: Well, not sure - saying this mostly because a lot of our types don't really represent some well-defined abstract entity, they're just wrappers around things that we tend to pass around together. Like here you end up splitting the contents into `(Vec<u8>, Attributes)` at the end, so `KeyInfo` isn't really more than a tuple in disguise - don't know if hiding the contents is relevant in any way. Maybe to make it impossible to mutate? But you never pass this struct outside of the module, so you could just mutate the contents before sending them out.

hug-dev: The context of making `KeyInfo` private is explained [here](https://github.com/parallaxsecond/parsec/issues/400). Basically to force people using the other methods that allow to directly get the deserialised Key ID instead of just a vector of bytes. I agree with your "tuple in disguise" point, see [this comment](https://github.com/parallaxsecond/parsec/pull/469#discussion_r667791363) where I say it would be better to remove the `Vec<u8>` or use `T` directly.

I believe even though the setters/getters are public, because `KeyInfo` is not public it means that the fields can only be accessed in this file (ie the setters/getters are not strictly needed for now).

hug-dev (approved):
It's good for me as it is, will let @ionut-arm confirm

ionut-arm (approved):
👍🏻

PR #471, opened 2021-07-12
Update usage of prost crates

ionut-arm:


ionut-arm:
It seems things don't compile under the 1.43.0 compiler: https://github.com/parallaxsecond/parsec/runs/3047525631?check_suite_focus=true

This is bad news, means I have to revert the interface change. We can do this safely because we don't use the `TimeStamp` from `prost`. Maybe there's a way to tell `cargo-audit` to acknowledge some advisories?

ionut-arm:
Closing this PR, will open a different one to fix our build through a different route.

PR #472, opened 2021-07-12
Split out the all-providers cargo check into its own CI job.

### LINKED ISSUE #447: Split the build tests on a different CI workflow
See [this specific comment](https://github.com/parallaxsecond/parsec/pull/445#discussion_r651785179) for background.

We perform a lot of build checks with different features and different versions of the Rust compiler. Those take a lot of time on the `all-providers` test. It would be wise to split them in their own workflow.

To do that, it might be needed to add a new argument to the `ci.sh` script so that only those are executed and run the script with that new argument in the Docker container.

MattDavis00:
The cargo check using the latest version of rust, and an old version, is no longer included in the `Various tests targeting a Parsec image with all providers included` job. It now has it's own dedicated GitHub Actions job.

Closes #447 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,19 @@ jobs:
       # When running the container built on the CI
       # run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec -t parsec-service-test-all /tmp/parsec/ci.sh all
 
+  build-all-providers:
+    name: Cargo check all-providers (current Rust stable & old compiler)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      # Use the following step when updating the `parsec-service-test-all` image
+      # - name: Build the container
+      #   run: pushd e2e_tests/docker_image && docker build -t parsec-service-test-all -f parsec-service-test-all.Dockerfile . && popd
+      - name: Run the container to execute the test script
+        run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec ghcr.io/parallaxsecond/parsec-service-test-all /tmp/parsec/ci.sh cargo-check
+      # When running the container built on the CI
+      # run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec -t parsec-service-test-all /tmp/parsec/ci.sh cargo-check
+
   mbed-crypto-provider:
     name: Integration tests using Mbed Crypto provider
     runs-on: ubuntu-latest
--- a/ci.sh
+++ b/ci.sh
@@ -117,13 +117,18 @@ while [ "$#" -gt 0 ]; do
         --no-stress-test )
             NO_STRESS_TEST="True"
         ;;
-        mbed-crypto | pkcs11 | tpm | trusted-service | cryptoauthlib | all)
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -16,6 +16,19 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      # Use the following step when updating the `parsec-service-test-all` image
+      # - name: Build the container
+      #   run: pushd e2e_tests/docker_image && docker build -t parsec-service-test-all -f parsec-service-test-all.Dockerfile . && popd
+      - name: Run the container to execute the test script
+        run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec ghcr.io/parallaxsecond/parsec-service-test-all /tmp/parsec/ci.sh cargo-check
+      # When running the container built on the CI
+      # run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec -t parsec-service-test-all /tmp/parsec/ci.sh build-all
```
hug-dev, line 30: The argument of `ci.sh` here is `build-all` and not `cargo-check`, maybe because you changed your mind? I personally prefer `build-all` but will leave it to you 😄

ci.sh
```diff
@@ -117,13 +117,20 @@ while [ "$#" -gt 0 ]; do
...
             PROVIDER_NAME=$1
-            cp $(pwd)/e2e_tests/provider_cfg/$1/config.toml $CONFIG_PATH
-            if [ "$PROVIDER_NAME" = "all" ]; then
+
+            # If running cargo-check, use all-providers config
+            if [ "$PROVIDER_NAME" = "cargo-check" ]; then
+                cp $(pwd)/e2e_tests/provider_cfg/all/config.toml $CONFIG_PATH
+            else
+                cp $(pwd)/e2e_tests/provider_cfg/$1/config.toml $CONFIG_PATH
+            fi
```
hug-dev, line 131: This shouldn't be needed as during the `cargo-check` tests we do not run Parsec but only build it, the config file is only useful at runtime 👌

hug-dev (commented):
Thank you, will speed up CI 💨

.github/workflows/ci.yml
ionut-arm: 😂 we actually had a discussion about that, changed because I wasn't keen on `build-all`...

ci.sh
ionut-arm: Hmm, true, I suggested this because the `trusted-service-provider` would need setting up for those tests, but I didn't realise it's disabled for now.

.github/workflows/ci.yml
hug-dev: I see 😛! Can stick to `cargo-check`

ci.sh
MattDavis00: Best to leave it in or remove it do you reckon?
If it is going to be required in the future it's an insignificant instruction in terms of test duration, so leaving it in wouldn't bring any adverse performance to the tests.

ionut-arm: I'd say remove it - there's plenty going on in this file, no reason to "maintain" things that aren't even used.

hug-dev (approved):
Thanks!

PR #474, opened 2021-07-13
Adjust TS provider linking

### LINKED ISSUE #427: Adjust linking for TS provider
The current state of `build.rs` is centred on the idea that the TS provider will be used for local testing, with in-process Trusted Services. The service is thus linked to Mbed Crypto (which provides the crypto implementation) and to `libstdc++`.

In order to prepare for a state in which the TS provider is actually used as intended, with the TS running in a TEE, we only need to dynamically link against `libts`. It's worth checking, however, if the dependency on `libstdc++` is required still. This involves making some of the dependencies conditional on, potentially, a compile-time flag or feature which indicates that the build is meant for testing - the CI should continue to work as it currently does.

ionut-arm:
This commit changes the libraries we link to for the TS provider to
work. Namely, it removes many of the previously linked libraries that we
used for testing. It seems even for test builds, there is no reason to
link the other libraries - simply linking to `libts.so` is enough.

Fixes #427

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -31,17 +31,8 @@ fn generate_ts_bindings(ts_include_dir: String) -> Result<()> {
         })?;
     let out_path = PathBuf::from(env::var("OUT_DIR").unwrap());
     bindings.write_to_file(out_path.join("ts_bindings.rs"))?;
-    // TODO: Remove once we can use the full TS stack and this isn't needed
-    println!("cargo:rustc-link-search=native=/usr/lib");
-    // libstdc++ is used on Ubuntu 18. On Ubuntu 20, libc++ (from libc++-dev and libc++abi-dev) can be used
-    println!("cargo:rustc-link-lib=dylib=stdc++");
 
-    println!("cargo:rustc-link-search=native=/usr/local/lib");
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -31,17 +31,8 @@ fn generate_ts_bindings(ts_include_dir: String) -> Result<()> {
...
-    println!("cargo:rustc-link-search=native=/usr/lib");
-    // libstdc++ is used on Ubuntu 18. On Ubuntu 20, libc++ (from libc++-dev and libc++abi-dev) can be used
-    println!("cargo:rustc-link-lib=dylib=stdc++");
 
-    println!("cargo:rustc-link-search=native=/usr/local/lib");
     println!("cargo:rustc-link-lib=dylib=ts");
-    // TODO: Remove once we can use the full TS stack and this isn't needed
-    println!("cargo:rustc-link-lib=static=mbedcrypto");
-    println!("cargo:rustc-link-lib=static=protobuf-nanopb");
-
```
hug-dev, line 44: The PRs we want 🥇

hug-dev (approved):

PR #482, opened 2021-07-19
Add TS provider to all-providers

ionut-arm:
This commit adds the TS provider feature to the all-providers feature. A
new check is also added on the CI for the `trusted-service-provider`
feature.

No tests are added so far.


ci.sh
```diff
@@ -276,9 +275,9 @@ if [ "$PROVIDER_NAME" = "cargo-check" ]; then
     RUST_BACKTRACE=1 cargo check --features="pkcs11-provider"
     RUST_BACKTRACE=1 cargo check --features="tpm-provider"
     RUST_BACKTRACE=1 cargo check --features="cryptoauthlib-provider"
-    # To be added when trusted-service is added to all-providers feature
```
hug-dev, line 279: 😎

hug-dev (approved):

ionut-arm:
NOTE: This particular change, adding the Trusted Services provider to `all-providers` and adding some cross-tests, has gone through a bit of churn to get to this form. The main problem stems from an incompatibility between having both the Trusted Service provider and the Mbed Crypto provider running at the same time. The core problem is that the "simulated", in-memory Trusted Service uses an implementation of Mbed Crypto under the hood and thus stores keys in the same way, under the same namespace as the Mbed Crypto provider. Attempting to create a key in both would then result in an `AlreadyExists` error being issued by the one that loses the data race.

To solve this, I attempted to modify the build steps for the Mbed Crypto library which we link to for testing, changing the location where the key files are stored (and leaving the TS to use the default location, thus avoiding any conflicts). However, when linking both the `libts.so` and `libmbedcrypto.so`, some of the PSA Crypto symbols get aliased. This will either lead to a segmentation fault (if the two libraries use incompatible versions), or to both providers still using the same storage space for keys.

The solution implemented here avoids all this by not using the Trusted Service provider in the main `all-providers` tests. Data format compatibility is enforced by means of a cross-signing/verification test with the PKCS11 provider. Since the PKCS11 provider is also cross-tested with the other providers for format compatibility, we can be sure that the Trusted Service provider is fully compatible with the rest.

PR #489, opened 2021-07-26
Do not login if no user pin was entered

hug-dev:
Fixes a bug where the service would still login even though no user pin
was given in the configuration file.


ionut-arm (approved):
💯 great stuff, thanks!

PR #491, opened 2021-07-27
Added provider names to the config.

### LINKED ISSUE #487: Add Provider Name Config Option
Add the following config option so that we can have multiple providers of the same type.

```
[[provider]]
name = "trusted-services-provider"
```

MattDavis00:
This is an optional field and defaults to a suitable name for each provider if it is not provided.

Two providers cannot have the same name.

Closes #487

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -101,6 +101,16 @@ manager_type = "OnDisk"
 
 # Example of an Mbed Crypto provider configuration.
 [[provider]]
+# ⚠
+# ⚠ WARNING: Provider name cannot change.
+# ⚠ WARNING: Choose a suitable naming scheme for your providers now.
+# ⚠ WARNING: Provider name defaults to "mbed-crypto-provider" if not provided, you will not be able to change
+# ⚠ the provider's name from this if you decide to use the default.
+# ⚠ WARNING: Changing provider name after use will lead to loss of existing keys.
+# ⚠
+# (Optional) The name of the provider
+name = "mbed-crypto-provider"
+
 # (Required) Type of provider.
 provider_type = "MbedCrypto"
 
@@ -114,6 +124,15 @@ key_info_manager = "on-disk-manager"
 
 # Example of a PKCS 11 provider configuration
 #[[provider]]
+# ⚠
+# ⚠ WARNING: Provider name cannot change.
+# ⚠ WARNING: Choose a suitable naming scheme for your providers now.
+# ⚠ WARNING: Provider name defaults to "pkcs11-provider" if not provided, you will not be able to change
+# ⚠ the provider's name from this if you decide to use the default.
+# ⚠ WARNING: Changing provider name after use will lead to loss of existing keys.
+# ⚠
+# (Optional) The name of the provider
+# name = "pkcs11-provider"
 #provider_type = "Pkcs11"
 #key_info_manager = "on-disk-manager"
 # (Required for this provider) Path to the location of the dynamic library loaded by this provider.
@@ -135,6 +154,15 @@ key_info_manager = "on-disk-manager"
 
 # Example of a TPM provider configuration
 #[[provider]]
+# ⚠
+# ⚠ WARNING: Provider name cannot change.
+# ⚠ WARNING: Choose a suitable naming scheme for your providers now.
+# ⚠ WARNING: Provider name defaults to "tpm-provider" if not provided, you will not be able to change
+# ⚠ the provider's name from this if you decide to use the default.
+# ⚠ WARNING: Changing provider name after use will lead to loss of existing keys.
+# ⚠
+# (Optional) The name of the provider
+# name = "tpm-provider"
 #provider_type = "Tpm"
 #key_info_manager = "on-disk-manager"
 # (Required) TPM TCTI device to use with this provider. The string can include configuration values - if no
@@ -161,6 +189,15 @@ key_info_manager = "on-disk-manager"
 # All below parameters depend on what devices, interfaces or parameters are required or supported by
 # "rust-cryptoauthlib" wrapper for cryptoauthlib and underlying hardware.
 #[[provider]]
+# ⚠
+# ⚠ WARNING: Provider name cannot change.
+# ⚠ WARNING: Choose a suitable naming scheme for your providers now.
+# ⚠ WARNING: Provider name defaults to "cryptoauthlib-provider" if not provided, you will not be able to change
+# ⚠ the provider's name from this if you decide to use the default.
+# ⚠ WARNING: Changing provider name after use will lead to loss of existing keys.
+# ⚠
+# (Optional) The name of the provider
+# name = "cryptoauthlib-provider"
 #provider_type = "CryptoAuthLib"
 #key_info_manager = "on-disk-manager"
 ##########
@@ -213,6 +250,15 @@ key_info_manager = "on-disk-manager"
 
 # Example of a Trusted Service provider configuration.
 #[[provider]]
+# ⚠
+# ⚠ WARNING: Provider name cannot change.
+# ⚠ WARNING: Choose a suitable naming scheme for your providers now.
+# ⚠ WARNING: Provider name defaults to "trusted-service-provider" if not provided, you will not be able to change
+# ⚠ the provider's name from this if you decide to use the default.
+# ⚠ WARNING: Changing provider name after use will lead to loss of existing keys.
+# ⚠
+# (Optional) The name of the provider
+# name = "trusted-service-provider"
 # (Required) Type of provider.
 #provider_type = "TrustedService"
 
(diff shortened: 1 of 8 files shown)


config.toml
```diff
@@ -101,6 +101,15 @@ manager_type = "OnDisk"
 
 # Example of an Mbed Crypto provider configuration.
 [[provider]]
+# ⚠
+# ⚠ WARNING: Provider name cannot change.
```
ionut-arm, line 105: I'd be inclined to specify why it shouldn't be changed, that it leads to existing keys being lost.

src/utils/config.rs
```diff
@@ -202,6 +212,27 @@ impl ProviderConfig {
             ProviderConfig::TrustedService { .. } => ProviderId::TrustedService,
         }
     }
+    /// Get the name of the Provider
+    /// If there is not one set, use the default.
+    pub fn provider_name(&self) -> String {
+        match *self {
+            ProviderConfig::MbedCrypto { ref name, .. } => name
+                .clone()
+                .unwrap_or_else(|| String::from("mbed-crypto-provider")),
```
ionut-arm, line 221: Could you move the `&str`s to be [associated constants](https://doc.rust-lang.org/edition-guide/rust-2018/trait-system/associated-constants.html) of the providers?

ionut-arm (commented):
Looks good, just a couple of small comments

config.toml
```diff
@@ -101,6 +101,15 @@ manager_type = "OnDisk"
...
+# ⚠ WARNING: Choose a suitable naming scheme for your providers now.
+# ⚠ WARNING: Provider name defaults to "mbed-crypto-provider" if not provided, you will not be able to change
+# ⚠ the provider's name from this if you decide to use the default.
+# ⚠
+# (Optional) The name of the provider
+name = "mbed-crypto-provider"
```
hug-dev, line 111: I am just thinking now if it's really worth it to have optional `name`? In the end most people will copy and adapt the default configuration so they won't lose time thinking about a provider name? And that would save us in some cases where people do not follow the warning notices?

Note that we can go from required to optional but can't do the reverse (for stability).

MattDavis00: Doesn't our stability guarantee that new config options will be optional? As to ensure that old config files will be valid in new versions of parsec?

hug-dev: ah yes you are right, forgot about that point 😁

ionut-arm: I guess we can add optional sections that have required fields, but not required fields to already existing sections... 😢

src/providers/cryptoauthlib/mod.rs
```diff
@@ -31,22 +31,33 @@ mod key_management;
...
 #[derive(Derivative)]
 #[derivative(Debug)]
 pub struct Provider {
     #[derivative(Debug = "ignore")]
     device: rust_cryptoauthlib::AteccDevice,
     provider_id: ProviderId,
+    // The name of the provider set in the config.
+    provider_name: String,
+    // The UUID of the provider.
+    provider_uuid: String,
```
ionut-arm, line 47: I just noticed: why is the provider UUID needed as a field here? Presumably that UUID is entirely static, so we could just use the `const` defined above.

ionut-arm (approved):
Looks good overall, just one minor comment

src/providers/cryptoauthlib/mod.rs
MattDavis00: I added it so that we can use `self.provider_uuid` when declaring `KeyIdentity`s in the submodules.
Don't know if that makes any sense though or if there is a better way. I did think it needed moving out from just being a `&str` in the `describe` function though as we'll need to use it in multiple places.

hug-dev: If you put it inside `impl Provider` (like the default name), you might be able to do `Self::PROVIDER_UUID` in submodules?

ionut-arm: Hmm, I think instead of doing that we could store the `const` value as a `Uuid` directly, using [one of the many `const` methods](https://docs.rs/uuid/0.8.2/uuid/struct.Uuid.html) implemented on Uuid. We can't use the current `parse_str`, but I think that's ok, we just need to make sure the current values are preserved. Then you can simply use that constant for `KeyIdentity`. Truth be told, I'd be inclined to not even have to pass that UUID as a parameter and simply populate it from the `KeyInfoManagerClient` like we do with the provider ID now.

ionut-arm: a `const Uuid` in the `impl` sounds like the best approach.

src/back/backend_handler.rs
```diff
@@ -33,6 +33,7 @@ pub struct BackEndHandler {
     provider: Arc<dyn Provide + Send + Sync>,
     #[derivative(Debug = "ignore")]
     converter: Box<dyn Convert + Send + Sync>,
+    provider_name: String,
```
hug-dev, line 36: Why do you need the `provider_name` here in the backend handler? For a future step of the implementation?

src/utils/service_builder.rs
```diff
@@ -207,11 +209,13 @@ fn build_backend_handlers(
...
     kim_factorys: HashMap<String, KeyInfoManagerFactory>,
-) -> Result<Vec<(ProviderId, Provider)>> {
+) -> Result<Vec<(ProviderId, Provider, String)>> {
     let mut list = Vec::new();
+    let mut provider_names = Vec::new();
     for config in configs {
+        // Check for duplicate providers.
         let provider_id = config.provider_id();
-        if list.iter().any(|(id, _)| *id == provider_id) {
+        if list.iter().any(|(id, _, _)| *id == provider_id) {
```
hug-dev, line 218: I guess this check will be updated in a future PR to check that we have unique (provider_name, provider_uuid) tuples?

src/utils/service_builder.rs
```diff
@@ -207,11 +209,13 @@ fn build_backend_handlers(
 fn build_providers(
     configs: &[ProviderConfig],
     kim_factorys: HashMap<String, KeyInfoManagerFactory>,
-) -> Result<Vec<(ProviderId, Provider)>> {
+) -> Result<Vec<(ProviderId, Provider, String)>> {
     let mut list = Vec::new();
+    let mut provider_names = Vec::new();
```
hug-dev, line 214: Nerd comment 🤓 : this could be a `HashSet` to make `contains` `O(1)` instead of `O(n)`
Note that you can ignore since `n`, the number of provider will likely not be `u32::MAX`

src/utils/service_builder.rs
```diff
@@ -220,6 +224,16 @@ fn build_providers(
             .into());
         }
 
+        // Check for duplicate provider names.
```
hug-dev, line 227: This seems to check for duplicate accross all providers as opposed to only providers of the same UUID. Are we allowing that providers of different type have the same name?
I think it's better that way (similarly for key names), that provider names are unique. That would also simplify the schema as `provider_name` alone can now be used to choose a provider.

hug-dev (commented):
Since you are doing all the work on a separate feature branch, this is totally fine to merge things little by little. I think it would be nice to have in a future PR some config tests:
- checking that the service starts with multiple providers of the same type with different names
- checking that the service fails to start with multiple providers of the same type with the default name
- creating a key with a provider of type A then restarting the service with a configuration where a new provider of the type A is added but check that the key can still be accessed

src/utils/service_builder.rs
MattDavis00: Yes, for now I've just kept the check as is until multiple providers are properly implemented.

src/utils/service_builder.rs
MattDavis00: Very true. Good spot 🤓

src/utils/service_builder.rs
MattDavis00: Yes, that's what I was going for. It's a design choice that can be up for debate though.
Personally I'm indifferent about how we scope `provider_name`'s uniqueness.

src/utils/service_builder.rs
ionut-arm: I'm genuinely curios if for cases like this (very small sets) there is actually a difference in speed. It's probably insignificant, but I guess for neatness' sake `HashSet` is better.

src/utils/service_builder.rs
MattDavis00: Fully removed this check in the end, now just the unique `provider_name` check exists.

src/back/backend_handler.rs
MattDavis00: Think I thought it might be required to link the backend handler to the provider but I can see there is already a reference to the provider object itself. Have removed this for now.

hug-dev (approved):
Thanks for the changes! We can always change things here in the future as this feature develops

ionut-arm (approved):
👍🏻

PR #493, opened 2021-07-28
Add the Class attribute when generating key pairs

hug-dev:
As per the specs this is not mandatory but could help with some PKCS11
implementations.


paulhowardarm (approved):
LGTM if the CI passes.

ionut-arm (approved):

PR #497, opened 2021-08-04
Switch imports to crates.io

ionut-arm:
This commit switches our imports from github-format to crates.io-format,
using the newly released crates.


src/providers/mbed_crypto/mod.rs
```diff
@@ -117,8 +117,9 @@ impl Provider {
                                     max_key_id = key_id;
                                 }
                             }
-                            Err(status::Error::DoesNotExist) => to_remove.push(key_triple.clone()),
+                            Err(status::Error::InvalidHandle) => to_remove.push(key_triple.clone()),
                             Err(e) => {
+                                println!("Key ID was {}", key_id);
```
hug-dev, line 122: I guess you used that to debug. Maybe that should be a `format_error` here.

ionut-arm: Yeah and nah - will just remove it, I wanted to see if the ID we were getting was actually invalid.

ci.sh
```diff
@@ -95,7 +95,7 @@ run_old_e2e_tests() {
         # /tmp/parsec/parsec.sock. This can not be created in the Dockerfile as this is where
         # the repository is checked out.
         ln -s /tmp/parsec.sock /tmp/parsec/parsec.sock
-        RUST_BACKTRACE=1 cargo test --manifest-path /tmp/old_e2e_tests/Cargo.toml normal_tests
+        RUST_BACKTRACE=1 cargo test --manifest-path /tmp/old_e2e_tests/Cargo.toml normal_tests -- --skip asym_verify_fail
```
hug-dev, line 98: Maybe add a comment explaining why we ignore this specific test?

hug-dev (approved):
Nice 🥇

ci.sh
ionut-arm: I'll add a comment, no one will check the CHANGELOG if they see this in the code at some point.

hug-dev: I meant adding the comment in this file as part of the CHANGELOG PR, to not have to run the CI once more

ionut-arm: Ah, sure!

hug-dev (approved):
The CI passed successfully [here](https://github.com/ionut-arm/parsec/actions/runs/1102081850) so I think it's all good!

Issue #486, opened 2021-07-22
Implement SQLiteKeyInfoManager


MattDavis00:
This issue will track the implementation of the `SQLiteKeyInfoManager`

### Design

Overarching design discussion #424 

### Steps

- [x] [Create design proposal](https://github.com/parallaxsecond/parsec/issues/424#issuecomment-881570975)
- [ ] Add Provider Name Config Option
- [ ] Change KeyTriple to Include Auth ID, Provider Name & Provider UUID
- [ ] Add Basic SQLiteKeyInfoManager Storage/Retrieval Functionality
- [ ] Refactor KeyInfoManager Unit Tests
- [ ] Add KeyInfoManager Migration Script

ionut-arm:
Removing the milestone from this issue as only some of the linked issues will be required for 0.9.0

PR #500, opened 2021-08-09
Added KeyIdentity to replace KeyTriple

### LINKED ISSUE #488: Change KeyTriple to Include Auth ID, Provider Name & Provider UUID
The structure of the new KeyTriple is still undecided but let the following structure provide an example.

```
KeyIdentity {
  application: ApplicationIdentity: {
    name: String,
    authenticator_id: u8
  },
  provider: ProviderIdentity {
    uuid: String,
    name: String,
  },
  key_name: String,
}
```

MattDavis00:
`KeyIdentity` contains an `ApplicationIdentity`, `ProviderIdentity` and a `key_name`.
`KeyTriple` still exists but solely for the use by the `OnDiskKeyInfoManager`.
This is preliminary work towards the new `SQLiteKeyInfoManager` as part of #486

Apologies for the large pr... 👨🏼‍💻

There may be some incorrect variable naming in places regarding `key_triple` where it should be `key_identity` instead. I will address this in a future PR as this will be a ~200 loc by itself.

Closes #488

### DIFF
--- a/src/authenticators/direct_authenticator/mod.rs
+++ b/src/authenticators/direct_authenticator/mod.rs
@@ -8,7 +8,7 @@
 //! This authenticator does not offer any security value and should only be used in environments
 //! where all the clients and the service are mutually trustworthy.
 
-use super::{AdminList, Application, Authenticate};
+use super::{AdminList, Application, ApplicationIdentity, Authenticate};
 use crate::front::listener::ConnectionMetadata;
 use crate::utils::config::Admin;
 use log::error;
@@ -61,7 +61,13 @@ impl Authenticate for DirectAuthenticator {
                 Ok(str) => {
                     let app_name = String::from(str);
                     let is_admin = self.admins.is_admin(&app_name);
-                    Ok(Application::new(app_name, is_admin))
+                    Ok(Application {
+                        identity: ApplicationIdentity {
+                            name: app_name,
+                            authenticator_id: AuthType::Direct,
+                        },
+                        is_admin,
+                    })
                 }
                 Err(_) => {
                     error!("Error parsing the authentication value as a UTF-8 string.");
@@ -76,7 +82,6 @@ impl Authenticate for DirectAuthenticator {
 mod test {
     use super::super::Authenticate;
     use super::DirectAuthenticator;
-    use crate::authenticators::ApplicationName;
     use parsec_interface::requests::request::RequestAuth;
     use parsec_interface::requests::ResponseStatus;
 
@@ -90,12 +95,12 @@ mod test {
         let req_auth = RequestAuth::new(app_name.clone().into_bytes());
         let conn_metadata = None;
 
-        let app = authenticator
+        let application = authenticator
             .authenticate(&req_auth, conn_metadata)
             .expect("Failed to authenticate");
 
-        assert_eq!(app.get_name(), &ApplicationName::from_name(app_name));
-        assert!(!app.is_admin);
+        assert_eq!(application.identity.name, app_name);
+        assert!(!application.is_admin);
     }
 
     #[test]
@@ -136,22 +141,19 @@ mod test {
         let req_auth = RequestAuth::new(app_name.clone().into_bytes());
         let conn_metadata = None;
 
-        let auth_name = authenticator
+        let application = authenticator
             .authenticate(&req_auth, conn_metadata)
             .expect("Failed to authenticate");
 
-        assert_eq!(auth_name.get_name(), &ApplicationName::from_name(app_name));
-        assert!(!auth_name.is_admin);
+        assert_eq!(application.identity.name, app_name);
+        assert!(!application.is_admin);
 
         let req_auth = RequestAuth::new(admin_name.clone().into_bytes());
-        let auth_name = authenticator
+        let application = authenticator
             .authenticate(&req_auth, conn_metadata)
             .expect("Failed to authenticate");
 
-        assert_eq!(
-            auth_name.get_name(),
-            &ApplicationName::from_name(admin_name)
-        );
-        assert!(auth_name.is_admin);
+        assert_eq!(application.identity.name, admin_name);
+        assert!(application.is_admin);
     }
 }
--- a/src/authenticators/mod.rs
+++ b/src/authenticators/mod.rs
@@ -4,7 +4,7 @@
 //!
 //! [Authenticators](https://parallaxsecond.github.io/parsec-book/parsec_service/authenticators.html)
 //! provide functionality to the service for verifying the authenticity of requests.
-//! The result of an authentication is an `ApplicationName` which is parsed by the authenticator and
+//! The result of an authentication is an `Application` which is parsed by the authenticator and
 //! used throughout the service for identifying the request initiator. The input to an authentication
 //! is the `RequestAuth` field of a request, which is parsed by the authenticator specified in the header.
 //! The authentication functionality is abstracted through an `Authenticate` trait.
@@ -29,28 +29,83 @@ use crate::front::listener::ConnectionMetadata;
 use crate::utils::config::Admin;
 use parsec_interface::operations::list_authenticators;
 use parsec_interface::requests::request::RequestAuth;
-use parsec_interface::requests::Result;
+use parsec_interface::requests::{AuthType, Result};
+use std::fmt;
 use std::ops::Deref;
 
-/// String wrapper for app names
-#[derive(Debug, Clone, Eq, PartialEq, Hash)]
-pub struct ApplicationName {
+/// A unique identifier for an application.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ApplicationIdentity {
+    /// The name of the application.
     name: String,
+    /// The id of the authenticator used to authenticate the application name.
+    authenticator_id: AuthType,
 }
 
-impl Deref for ApplicationName {
-    type Target = String;
+impl fmt::Display for ApplicationIdentity {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "ApplicationIdentity: [name=\"{}\", authenticator_id=\"{}\"]",
+            self.name, self.authenticator_id
+        )
+    }
+}
 
-    fn deref(&self) -> &Self::Target {
+impl ApplicationIdentity {
+    /// Creates a new instance of ProviderIdentity.
+    pub fn new(name: String, authenticator_id: AuthType) -> ApplicationIdentity {
+        ApplicationIdentity {
+            name,
+            authenticator_id,
+        }
+    }
+
+    /// Get the identity of the application
+    pub fn name(&self) -> &String {
         &self.name
+    }
+
+    /// Get whether the application has administrator rights
+    pub fn authenticator_id(&self) -> &AuthType {
+        &self.authenticator_id
     }
 }
 
 /// Wrapper for a Parsec application
 #[derive(Debug, Clone)]
 pub struct Application {
-    name: ApplicationName,
+    /// The identity of the Application
+    identity: ApplicationIdentity,
+    /// Whether the application has administrator rights
     is_admin: bool,
+}
+
+impl fmt::Display for Application {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "Application {{identity: {}, is_admin: {}}}",
+            self.identity, self.is_admin
+        )
+    }
+}
+
+impl Application {
+    /// Creates a new instance of ProviderIdentity.
+    pub fn new(identity: ApplicationIdentity, is_admin: bool) -> Application {
+        Application { identity, is_admin }
+    }
+
+    /// Get the identity of the application
+    pub fn identity(&self) -> &ApplicationIdentity {
+        &self.identity
+    }
+
+    /// Get whether the application has administrator rights
+    pub fn is_admin(&self) -> &bool {
+        &self.is_admin
+    }
 }
 
 /// Authentication interface
@@ -76,45 +131,6 @@ pub trait Authenticate {
         auth: &RequestAuth,
         meta: Option<ConnectionMetadata>,
     ) -> Result<Application>;
-}
-
-impl ApplicationName {
-    /// Create ApplicationName from name string only
-    pub fn from_name(name: String) -> ApplicationName {
-        ApplicationName { name }
-    }
-}
-
-impl Application {
-    /// Create a new Application structure
-    pub fn new(name: String, is_admin: bool) -> Application {
-        Application {
-            name: ApplicationName::from_name(name),
-            is_admin,
-        }
-    }
-
-    /// Check whether the application is an admin
-    pub fn is_admin(&self) -> bool {
-        self.is_admin
-    }
-
-    /// Get a reference to the inner ApplicationName string
-    pub fn get_name(&self) -> &ApplicationName {
-        &self.name
-    }
-}
-
-impl From<Application> for ApplicationName {
-    fn from(auth: Application) -> Self {
-        auth.name
-    }
-}
-
-impl std::fmt::Display for ApplicationName {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.name)
-    }
 }
 
 #[derive(Debug, Clone, Default)]
--- a/src/key_info_managers/mod.rs
+++ b/src/key_info_managers/mod.rs
@@ -6,15 +6,18 @@
 //! trait to help providers to store in a persistent manner the mapping between the name and the
 //! information of the keys they manage. Different implementors might store this mapping using different
 //! means but it has to be persistent.
-
-use crate::authenticators::ApplicationName;
+use crate::authenticators::ApplicationIdentity;
+#[allow(deprecated)]
+use crate::key_info_managers::on_disk_manager::KeyTriple;
+use crate::providers::ProviderIdentity;
 use crate::utils::config::{KeyInfoManagerConfig, KeyInfoManagerType};
 use anyhow::Result;
 use derivative::Derivative;
 use parsec_interface::operations::psa_key_attributes::Attributes;
-use parsec_interface::requests::{ProviderId, ResponseStatus};
+use parsec_interface::requests::{AuthType, ResponseStatus};
 use serde::de::DeserializeOwned;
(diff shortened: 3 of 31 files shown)


src/key_info_managers/on_disk_manager/mod.rs
```diff
@@ -303,48 +303,75 @@ impl OnDiskKeyInfoManager {
 }
 
 impl ManageKeyInfo for OnDiskKeyInfoManager {
-    fn get(&self, key_triple: &KeyTriple) -> Result<Option<&KeyInfo>, String> {
+    fn get(&self, key_identity: &KeyIdentity) -> Result<Option<&KeyInfo>, String> {
+        let key_triple = KeyTriple::try_from(key_identity.clone()).map_err(|_e| {
+            error!("Error creating KeyTriple from KeyIdentity");
+            "Error creating KeyTriple from KeyIdentity"
```
ionut-arm, line 309: Seems a bit superfluous, I think the KIM client then goes ahead and prints what it gets from this function as an `error!` as well, it might be better to just return the `&str` (?)

src/key_info_managers/mod.rs
```diff
@@ -261,24 +414,25 @@ impl KeyInfoManagerClient {
...
-        for key_triple in key_triples {
-            if !clients.contains(key_triple.app_name()) {
-                let _ = clients.push(key_triple.app_name().clone());
+        let mut clients = Vec::new();
+        for key_identity in key_identities {
+            // TODO: Check this comparison
+            if !clients.contains(&key_identity.application) {
+                let _ = clients.push(key_identity.application.clone());
             }
         }
```
ionut-arm, line 437: Could do the filtering as a `HashSet`, but don't know if it's that helpful since you'd have to convert to `Vec` afterwards.

src/providers/core/mod.rs
```diff
@@ -142,7 +150,10 @@ impl Provide for Provider {
             // Currently Parsec only stores keys, we delete all of them.
             let keys = provider
                 .list_keys(
-                    ApplicationName::from_name(client.clone()),
+                    ApplicationIdentity {
+                        name: client.clone(),
+                        authenticator_id: AuthType::NoAuth as u8,
+                    }, // TODO: Check auth type
```
ionut-arm, line 156: This is interesting. So we have to delete a client, and this has to happen for the authenticator used by the admin for this request. We have to change the interface for `delete_client`, I assume

src/key_info_managers/mod.rs
```diff
@@ -76,6 +123,108 @@ impl KeyTriple {
...
+            }
+            #[cfg(feature = "tpm-provider")]
+            crate::providers::tpm::Provider::PROVIDER_UUID => {
+                Ok::<ProviderId, Self::Error>(ProviderId::Tpm)
+            }
+            #[cfg(feature = "trusted-service-provider")]
+            crate::providers::trusted_service::Provider::PROVIDER_UUID => {
+                Ok::<ProviderId, Self::Error>(ProviderId::TrustedService)
+            }
+            _ => Err("Provider does not exist"),
```
ionut-arm, line 164: Maybe something like "Provider not recognized", as it could be that the provider "exists", it just wasn't compiled in. Also, it's worth adding the UUID in there, just to make it easier to debug.

ionut-arm (commented):
Thanks for the massive change! 💪🏻 I left a bunch of comments below, apologies if some of them seem to contradict each other - I went on a journey of self-discovery myself, reading through this.

src/key_info_managers/mod.rs
```diff
@@ -303,26 +457,32 @@ impl KeyInfoManagerClient {
...
-        let key_triples = key_info_manager_impl
-            .get_all(self.provider_id)
+        let key_identities = key_info_manager_impl
+            .get_all(self.provider_identity.clone())
             .map_err(to_response_status)?;
 
-        for key_triple in key_triples {
-            if key_triple.app_name() != app_name {
+        for key_identity in key_identities {
+            // TODO: Change this to also match on authenticator_id for SQLiteKeyInfoManager.
```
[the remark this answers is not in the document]
MattDavis00: I did originally write something along the lines of:
```
if key_identity.application.name != application_identity.name
  || key_identity.application.authenticator_id != application_identity.authenticator_id
{
    continue;
}
```
Failed CI though as when the list method is invoked; the `ApplicationIdentity` for that request has a valid `AuthType`/`authenticator_id`, whereas the keys being returned from the `OnDiskKeyInfoManager` will always return `AuthType::NoAuth`. Could be fixed by changing the default from `AuthType::NoAuth` to `AuthType::Unknown` or `AuthType::Wildcard`? Then code in a "skip this check if Auth type for key is unknown" (`key_identity.application.authenticator_id != application_identity.authenticator_id`).

src/key_info_managers/on_disk_manager/mod.rs
```diff
@@ -303,48 +303,75 @@ impl OnDiskKeyInfoManager {
...
+        Ok(key_identites)
     }
 
     fn insert(
         &mut self,
-        key_triple: KeyTriple,
+        key_identity: KeyIdentity,
         key_info: KeyInfo,
     ) -> Result<Option<KeyInfo>, String> {
+        println!("{}", key_identity);
```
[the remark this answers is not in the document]
MattDavis00: Yes 😆

src/providers/core/mod.rs
MattDavis00: Yes I would think so. Will work fine currently as the On Disk KIM doesn't rely on the AuthType. Although for the SQLite KIM it will need updating.

src/key_info_managers/mod.rs
```diff
@@ -76,6 +123,108 @@ impl KeyTriple {
...
+    }
+}
+
+impl TryFrom<KeyIdentity> for KeyTriple {
+    type Error = &'static str;
+
+    fn try_from(key_identity: KeyIdentity) -> ::std::result::Result<Self, Self::Error> {
+        let provider_id = match key_identity.provider.uuid.as_str() {
+            crate::providers::core::Provider::PROVIDER_UUID => {
+                Ok::<ProviderId, Self::Error>(ProviderId::Core)
```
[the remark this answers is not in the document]
MattDavis00: Think the compiler can. It's clippy that got angry.
Changed the error return type from `str` to `String` to allow for formatting and that seems to have fixed it 🤷🏼‍♂️

src/key_info_managers/mod.rs
```diff
@@ -76,6 +123,108 @@ impl KeyTriple {
...
+        }?;
+
+        Ok(KeyIdentity {
+            provider: ProviderIdentity {
+                name: String::from(""),
+                uuid: provider_uuid,
+            },
+            application: ApplicationIdentity {
+                name: String::from(key_triple.app_name().as_str()),
+                authenticator_id: AuthType::NoAuth as u8,
```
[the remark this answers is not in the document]
MattDavis00: Added `0xFF` for values returned from the KIM where no authentication information is available.
Added `0xFE` for values where the request does not have sufficient authentication information (`delete_client`).

src/key_info_managers/mod.rs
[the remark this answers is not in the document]
MattDavis00: I will leave this for when I implement the `SQLiteKeyInfoManager` logic. I will need to write a test to check that keys are correctly divided by authenticator_id when I do this.

src/authenticators/mod.rs
```diff
@@ -46,11 +49,65 @@ impl Deref for ApplicationName {
...
+    }
+}
+
+/// A unique identifier for an application.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ApplicationIdentity {
+    /// The name of the application.
+    pub name: String,
+    /// The id of the authenticator used to authenticate the application name.
+    pub authenticator_id: u8,
```
ionut-arm, line 81: Maybe it'd be possible to use `Option<AuthType>` here, or some wrapper type around `AuthType` so that we don't just have the special case values (like `0xFF`) floating around. Maybe an enum, something like
```rust
enum AuthenticatorIdentity {
    KnownType(AuthType),
    UnknownType,
}
```
@hug-dev - thoughts?

ionut-arm (commented):
Another few comments, should've noticed the `pub` fields yesterday :(

hug-dev (commented):
Thanks for the changes. Sorry for the big amount of review comments adding up to what you already got 😛 
It is good though since this is a big and important change!

src/authenticators/mod.rs
```diff
@@ -46,11 +49,65 @@ impl Deref for ApplicationName {
...
 /// Wrapper for a Parsec application
 #[derive(Debug, Clone)]
 pub struct Application {
-    name: ApplicationName,
-    is_admin: bool,
+    /// The identity of the Application
+    pub identity: ApplicationIdentity,
+    /// Whether the application has administrator rights
+    pub is_admin: bool,
+}
```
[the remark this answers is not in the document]
MattDavis00: > would it make sense to move those structure in the on_disk_key_info_manager module

👍🏽 I'll move the legacy structures into the OnDiskKIM.

> To make it simpler, could we flatten the struct?

I'd lean towards keeping it separated. This is as at the backend handler just the identity is plucked and passed down like so:
```
let app = unwrap_or_else_return!(app.ok_or(ResponseStatus::NotAuthenticated));
let result = unwrap_or_else_return!(self
    .provider
    .psa_generate_key(app.identity, op_generate_key));
```

src/authenticators/mod.rs
MattDavis00: We would need something like `AuthType::Unknown` adding to the parsec interface.

MattDavis00:
```
pub enum AuthType {
    /// No authentication
    NoAuth = 0,
    /// Direct authentication
    Direct = 1,
    /// JSON Web Tokens (JWT) authentication (not currently supported)
    Jwt = 2,
    /// Unix peer credentials authentication
    UnixPeerCredentials = 3,
    /// Authentication verifying a JWT SPIFFE Verifiable Identity Document
    JwtSvid = 4,
    /// The authentication of an incoming request is unknown.
    IngressAuthUnknown = 0xFE
    /// The authentication of a stored key is unknown.
    EgressAuthUnknown = 0xFF
}
```
Maybe something like this? Just a suggestion

ionut-arm: Hmmm, I'd keep the "unknown" stuff out of `AuthType`, I think Hugues was asking why we can't use `AuthType` as it is now. We can't because there are places (e.g. when reading from disk in the on-disk KIM) where we don't know what authentication was used, but we need to populate with something. The initial option was to just use `NoAuth`, but I don't think that's sound because you could end up actually confusing it with `NoAuth` as an actual authentication "mechanism" somewhere.

MattDavis00:
Test comment

src/authenticators/mod.rs
MattDavis00: Suppose in the future we update `list_keys` and the `KeyInfo` object to return our new data types. Would it not be better to have a "we don't know what authentication this key was created with" type in the interface? I do agree we should have something different to `NoAuth`

MattDavis00: After further investigation, `IngressAuthUnknown` isn't really needed as I can just update the functions where authentication information is currently missing.
```
pub enum AuthType {
    /// No authentication
    NoAuth = 0,
    /// Direct authentication
    Direct = 1,
    /// JSON Web Tokens (JWT) authentication (not currently supported)
    Jwt = 2,
    /// Unix peer credentials authentication
    UnixPeerCredentials = 3,
    /// Authentication verifying a JWT SPIFFE Verifiable Identity Document
    JwtSvid = 4,
    /// The authentication type of a stored key is unknown.
    /// In older versions of Parsec the authentication method was not stored.
    /// Therefore, a key with AuthType::Unknown *could* have been created with
    /// authentication or with NoAuth; neither state can however be guaranteed
    /// therefore AuthType::Unknown will be returned to avoid collisions.
    Unknown = 0xFF
}
```
Something like this?

MattDavis00:
Currently admin operations such as `delete_client` are not scoped to the `authenticator` of the administrator.
Example:
- Apps:
`app1`-`authenticator A`
`app2`-`authenticator B`
- Admin:
`admin`-`authenticator B`

Currently the admin can delete both `app1` and `app2`. With regards to the new KIM and admin permissions, how do we want this to work? Currently only 1 authenticator is supported and hence this isn't really an issue **yet**. Do we want an administrator authenticated with `authenticator B` to be scoped only to actions on `authenticator B` resources, or do we want them to have "global" admin privileges where they can perform admin actions on any resource regardless of authenticator?

src/authenticators/direct_authenticator/mod.rs
```diff
@@ -136,22 +141,19 @@ mod test {
...
+        let application = authenticator
             .authenticate(&req_auth, conn_metadata)
             .expect("Failed to authenticate");
 
-        assert_eq!(
-            auth_name.get_name(),
-            &ApplicationName::from_name(admin_name)
-        );
-        assert!(auth_name.is_admin);
+        assert_eq!(application.identity.name, admin_name);
```
hug-dev, line 156: Just for my curiosity: how can this compile if the fields are now private?

src/key_info_managers/mod.rs
```diff
@@ -24,18 +27,21 @@ pub mod on_disk_manager;
 /// This structure corresponds to a unique identifier of the key. It is used internally by the Key
 /// ID manager to refer to a key.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
-pub struct KeyTriple {
-    app_name: ApplicationName,
-    provider_id: ProviderId,
+pub struct KeyIdentity {
+    /// The identity of the application that created the key.
+    application: ApplicationIdentity,
```
hug-dev, line 32: Right, I see why creating `ApplicationIdentity` (and not flattening it) makes sense now 👌

src/key_info_managers/mod.rs
```diff
@@ -24,18 +27,21 @@ pub mod on_disk_manager;
...
+    /// The identity of the provider where the key is stored.
+    provider: ProviderIdentity,
```
hug-dev, line 34: What about the idea of having this as an `Option<ProviderIdentity>`, only required for the OnDisk KIM?

src/key_info_managers/mod.rs
```diff
@@ -50,29 +56,43 @@ struct KeyInfo {
...
+            application,
+            provider,
             key_name,
         }
     }
 
     /// Checks if this key belongs to a specific provider.
-    pub fn belongs_to_provider(&self, provider_id: ProviderId) -> bool {
-        self.provider_id == provider_id
+    pub fn belongs_to_provider(&self, provider_name: String) -> bool {
```
hug-dev, line 74: Probably `&str` is better here. Also do we then say that provider names are unique (I think that's fine)?

src/key_info_managers/mod.rs
```diff
@@ -96,14 +116,14 @@ trait ManageKeyInfo {
...
-    fn get(&self, key_triple: &KeyTriple) -> Result<Option<&KeyInfo>, String>;
+    fn get(&self, key_identity: &KeyIdentity) -> Result<Option<&KeyInfo>, String>;
 
     /// Returns a Vec of reference to the key triples corresponding to this provider.
     ///
     /// # Errors
     ///
     /// Returns an error as a String if there was a problem accessing the Key Info Manager.
-    fn get_all(&self, provider_id: ProviderId) -> Result<Vec<&KeyTriple>, String>;
+    fn get_all(&self, provider_identity: ProviderIdentity) -> Result<Vec<KeyIdentity>, String>;
```
hug-dev, line 126: I think passing a reference would be better here. But maybe `provider_name: &str` is enough?

src/key_info_managers/mod.rs
```diff
@@ -363,13 +395,14 @@ pub struct KeyInfoManagerFactory {
 
 impl KeyInfoManagerFactory {
     /// Create a KeyInfoManagerFactory
-    pub fn new(config: &KeyInfoManagerConfig) -> Result<Self> {
+    pub fn new(config: &KeyInfoManagerConfig, default_auth_type: AuthType) -> Result<Self> {
```
hug-dev, line 398: Maybe an `Option` is well suited here as this field can be clearly ignored for other KIM

hug-dev (approved):
Mainly fine for me. I think I should not block the implementation for too long since this is hard to work little by little here. And it's all OK since you are on your own branch.

> Do we want an administrator authenticated with authenticator B to be scoped only to actions on authenticator B resources, or do we want them to have "global" admin privileges where they can perform admin actions on any resource regardless of authenticator?

I think the scoped version is the one that we want. Having the new KIM in mind we wrote [in the book](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/delete_client.html#description):

> Only the clients using the same authentication method as this request will be deleted. It has no impact currently as only one authentication method in the service is supported but might do if the service supports multiple.

src/authenticators/direct_authenticator/mod.rs
MattDavis00: Interesting, didn't notice that. I think it's because the `DirectAuthenticator` is a descendant of `Authenticator` where `Application` is defined. Has to do with this little snippet in Rust by Example I believe:

> If an item is private, it may be accessed by the current module and its descendants.

Source: https://doc.rust-lang.org/reference/visibility-and-privacy.html

src/key_info_managers/mod.rs
MattDavis00: Yeah maybe. I'll have a look into it more seeing what issues I run into implementing the `SQLiteKeyInfoManager` itself. E.g. what information I'm missing etc

src/key_info_managers/mod.rs
MattDavis00: Chose to do it like this as it will need to follow the common API of `ManageKeyInfo`.`get_all` method, which will later be used by the `SQLiteKeyInfoManager`. And as the `SQLiteKeyInfoManager` will use this. As to whether we're just using `provider_name`s to distinguish providers is another topic. I personally think provider names should be unique 🤷🏼‍♂️

src/key_info_managers/mod.rs
MattDavis00: Yes, probably best. I'm sure I'll run into having to change it when I add the `SQLiteKeyInfoManager` to the `KeyInfoManagerFactory`

src/authenticators/direct_authenticator/mod.rs
hug-dev: Ah interesting!

PR #512, opened 2021-09-03
Added validation when database is loaded within the SQLiteKeyInfoManager.

MattDavis00:
This validation checks that:
- The database exists.
- The correct tables exist.
- The `schema_version` is correct.
- All mapping records have the correct `key_info_version`.

There is a future opportunity here to create more unit tests covering:
- Loading an empty database.
- Loading a database with either the `key_mapping` or `kim_metadata` table missing.
- Loading a database with an incorrect `schema_version`.
- Loading a database where some mappings have incorrect/incompatible `key_info_version` values.

### DIFF
--- a/src/key_info_managers/sqlite_manager/mod.rs
+++ b/src/key_info_managers/sqlite_manager/mod.rs
@@ -7,7 +7,7 @@ use super::{KeyIdentity, KeyInfo, ManageKeyInfo};
 use crate::authenticators::ApplicationIdentity;
 use crate::providers::ProviderIdentity;
 use anyhow::Result;
-use log::info;
+use log::{error, info};
 use num_traits::FromPrimitive;
 use parsec_interface::requests::AuthType;
 use rusqlite::types::Type::{Blob, Integer};
@@ -35,11 +35,6 @@ pub struct SQLiteKeyInfoManager {
     /// The file path where the SQLite database exists. This database holds
     /// key identity to key info mappings.
     database_path: PathBuf,
-}
-
-struct KeyInfoRecord {
-    key_identity: KeyIdentity,
-    key_info: KeyInfo,
 }
 
 /// TODO: Implement this until the interface TryFrom u8 to AuthType is implemented.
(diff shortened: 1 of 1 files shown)


src/key_info_managers/sqlite_manager/mod.rs
```diff
@@ -70,28 +65,171 @@ impl SQLiteKeyInfoManager {
...
+                        .into_string()
+                        .unwrap_or_else(|_| "DB_FILE_PATH_UNKNOWN".to_string()),
+                );
+                error!("{}", error_message);
+                return Err(Error::new(ErrorKind::Other, error_message).into());
+            }
+            // The correct number of tables are present, no-op
+            2 => {}
+            // Shouldn't be possible to reach this. If it does, throw an error
+            _ => {
```
ionut-arm, line 141: could you skip `1` above and just let it fall under this? Seems like the code is identical.

src/key_info_managers/sqlite_manager/mod.rs
```diff
@@ -70,28 +65,171 @@ impl SQLiteKeyInfoManager {
...
+            SELECT
+                *
+            FROM
+                key_mapping
+        ",
+        )?;
+        let mut rows = key_info_version_stmt.query(params![])?;
+        while let Some(row) = rows.next()? {
+            let version_number: u8 = row.get("key_info_version")?;
+            if version_number != CURRENT_KEY_INFO_VERSION {
```
ionut-arm, line 202: Is it not possible (and I guess this applies to the previous query as well) to just expand the query to ask for entries where the version number has this property (i.e. `!= CURRENT_KEY_INFO_VERSION`)? Would've thought SQL allows us to do that, and then you can just check the row count.

ionut-arm (commented):
Thanks for the patch! Don't worry about the build for now 😬

src/key_info_managers/sqlite_manager/mod.rs
MattDavis00: Yes 🤦🏼‍♂️ think I was just on autopilot copying from one of the previous SQL statements.

MattDavis00: Having looked at it a second time. I've added the `WHERE key_info_version != CURRENT_KEY_INFO_VERSION` clause to the SQL statement. I've however chosen not to use the length, this is as the value returned from the query only implements iterator. I would have to call `.count()` which under the hood calls `.next()` for every item.

Opted for this instead:
```
let rows = SQL_QUERY();
// If a mapping exists with the wrong key_info_version, throw an error.
if let Some(row) = rows.next()? {
    .....throw error
}
```

ionut-arm (approved):
👍🏻 Thanks!

PR #513, opened 2021-09-03
Added provider name config tests.

### LINKED ISSUE #496: Add config tests for multiple provider names

MattDavis00:
Added config tests that cover different values given to provider names within the config.
These tests cover two providers where:
- Same provider type, same name. (Throw error)
- Same provider type, different name. (Allow)
- Same provider type, same default name. (Throw error).
- Different types, same name. (Throw error). (We could allow this in the future if we wish).
- Different types, different default names. (Allow)

Closes #496

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -5,6 +5,16 @@ pub mod cli;
 pub mod config;
 mod global_config;
 mod service_builder;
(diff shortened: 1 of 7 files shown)


ionut-arm (approved):
Looks good, thanks!

PR #515, opened 2021-09-03
Update provider to use new version fo TransKeyCtx

ionut-arm:
This commit updates the TPM provider to use the updated version of the
TransientKeyContext which handles key material in the form of public and
private fields (instead of an encrypted context).

### DIFF
--- a/e2e_tests/tests/per_provider/normal_tests/import_key.rs
+++ b/e2e_tests/tests/per_provider/normal_tests/import_key.rs
@@ -365,43 +365,43 @@ fn failed_imported_key_should_be_removed() -> Result<()> {
     Ok(())
 }
 
-#[cfg(feature = "tpm-provider")]
-#[test]
-fn import_key_pair() {
-    let mut client = TestClient::new();
-    let key_name = String::from("failed_imported_key_should_be_removed");
+// #[cfg(feature = "tpm-provider")]
+// #[test]
+// fn import_key_pair() {
+//     let mut client = TestClient::new();
+//     let key_name = String::from("failed_imported_key_should_be_removed");
 
-    client
-        .import_key(
-            key_name,
-            Attributes {
-                lifetime: Lifetime::Persistent,
-                key_type: Type::RsaKeyPair,
-                bits: 1024,
-                policy: Policy {
-                    usage_flags: UsageFlags {
-                        export: false,
-                        copy: false,
-                        cache: false,
-                        encrypt: false,
-                        decrypt: false,
-                        sign_message: true,
-                        sign_hash: true,
-                        verify_message: true,
-                        verify_hash: true,
-                        derive: false,
-                    },
-                    permitted_algorithms: Algorithm::AsymmetricSignature(
-                        AsymmetricSignature::RsaPkcs1v15Sign {
-                            hash_alg: Hash::Sha256.into(),
-                        },
-                    ),
-                },
-            },
-            KEY_PAIR_DATA.to_vec(),
-        )
-        .unwrap();
-}
+//     client
+//         .import_key(
+//             key_name,
+//             Attributes {
+//                 lifetime: Lifetime::Persistent,
+//                 key_type: Type::RsaKeyPair,
+//                 bits: 1024,
+//                 policy: Policy {
+//                     usage_flags: UsageFlags {
+//                         export: false,
+//                         copy: false,
+//                         cache: false,
+//                         encrypt: false,
+//                         decrypt: false,
+//                         sign_message: true,
+//                         sign_hash: true,
+//                         verify_message: true,
+//                         verify_hash: true,
+//                         derive: false,
+//                     },
+//                     permitted_algorithms: Algorithm::AsymmetricSignature(
+//                         AsymmetricSignature::RsaPkcs1v15Sign {
+//                             hash_alg: Hash::Sha256.into(),
+//                         },
+//                     ),
+//                 },
+//             },
+//             KEY_PAIR_DATA.to_vec(),
+//         )
+//         .unwrap();
+// }
 
 #[cfg(any(feature = "mbed-crypto-provider", feature = "cryptoauthlib-provider"))]
 #[test]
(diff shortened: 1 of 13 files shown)


src/providers/tpm/key_management.rs
```diff
@@ -14,12 +14,61 @@ use parsec_interface::operations::{
...
+                            )
+                            .map_err(utils::to_response_status)?,
+                        auth_value: legacy_ctx.auth_value,
+                        context: utils::empty_tpms_ctx(),
+                    };
+
+                    // Grab key attributes and replace legacy entry with new one
+                    let attributes = self.key_info_store.get_key_attributes(key_triple)?;
+                    self.key_info_store.remove_key_info(key_triple)?;
+                    let _ = self.key_info_store.insert_key_info(
```
paulhowardarm, line 60: If we don't already, it might be worth stress-testing this pattern of removal followed by re-insertion of the same key in the on-disk KIM, because I have known this kind of thing to fail due to race conditions against the filesystem (although this is going back several years). A `patch_key_info` operation in the KIM might be preferable, but I can understand that this requires a more extensive change of the kind we're trying to avoid here. On balance, I'm happy with this as it is.

paulhowardarm (approved):
The overall approach looks good to me.

src/providers/tpm/key_management.rs
ionut-arm: Yeah, I've been thinking about this - I'll add a `replace_key_info` to the KIM client. The worst that can happen here is that a thread tries to get info about the key in the split second between when the attributes are deleted and when they're placed back, and doesn't find any so it returns "PsaErrorNotFound".

ionut-arm: Done!

src/providers/tpm/utils.rs
```diff
@@ -93,27 +95,64 @@ pub fn to_response_status(error: Error) -> ResponseStatus {
...
+    /// This value is kept for legacy purposes, to aid in the migration process
+    pub context: TpmsContext,
+    /// This value is confidential and needs to be zeroized by its new owner.
+    pub auth_value: Vec<u8>,
+    pub key_material: KeyMaterial,
+}
+
+// LegacyPasswordContext that stored key contexts only.
+#[derive(Serialize, Deserialize, Zeroize)]
+pub struct LegacyPasswordContext {
```
MattDavis00, line 107: Should this struct have a `#[deprecated(...)]` attribute?

src/providers/tpm/utils.rs
```diff
@@ -93,27 +95,64 @@ pub fn to_response_status(error: Error) -> ResponseStatus {
 // The PasswordContext is what is stored by the Key Info Manager.
 #[derive(Serialize, Deserialize, Zeroize)]
 pub struct PasswordContext {
+    /// This value is kept for legacy purposes, to aid in the migration process
+    pub context: TpmsContext,
+    /// This value is confidential and needs to be zeroized by its new owner.
+    pub auth_value: Vec<u8>,
+    pub key_material: KeyMaterial,
+}
```
MattDavis00, line 103: Could this struct be encapsulated, removing the use of `pub` properties?

src/providers/tpm/utils.rs
```diff
@@ -93,27 +95,64 @@ pub fn to_response_status(error: Error) -> ResponseStatus {
...
+                2048 => RsaKeyBits::Rsa2048,
+                3072 => RsaKeyBits::Rsa3072,
+                4096 => RsaKeyBits::Rsa4096,
+                _ => {
+                    error!(
+                        "Requested RSA key size is not supported ({})",
+                        attributes.bits
+                    );
+                    return Err(ResponseStatus::PsaErrorInvalidArgument);
+                }
```
MattDavis00, line 131: Not sure if it can make use of `try_from()` within the `tss-esapi` crate [here](https://github.com/parallaxsecond/rust-tss-esapi/blob/main/tss-esapi/src/interface_types/key_bits.rs#L134). I would say the conversion should probably lie within the `tss-esapi` crate. Maybe an implementation using `num::FromPrimitive` or just normal match arms.

MattDavis00 (commented):
Looking good so far! Just a few comments.

e2e_tests/tests/per_provider/normal_tests/import_key.rs
```diff
@@ -365,43 +365,43 @@ fn failed_imported_key_should_be_removed() -> Result<()> {
...
-                    ),
-                },
-            },
-            KEY_PAIR_DATA.to_vec(),
-        )
-        .unwrap();
-}
+// #[cfg(feature = "tpm-provider")]
+// #[test]
+// fn import_key_pair() {
```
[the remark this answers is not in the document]
ionut-arm: Yes, I've essentially removed the import key functionality (for key pairs, public keys are still allowed).

src/providers/tpm/utils.rs
ionut-arm: Yep, guess that would make it obvious if it gets used anywhere else.

src/providers/tpm/utils.rs
ionut-arm: Yeah, there is a `TryFrom<u16>` implemented already, I'll switch to using that

e2e_tests/tests/per_provider/tpm_reset.rs
```diff
@@ -0,0 +1,36 @@
+// Copyright 2021 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+
+use e2e_tests::TestClient;
+
+const RSA_KEY_NAME: &str = "tpm-reset-rsa";
+const ECC_KEY_NAME: &str = "tpm-reset-ecc";
+
+#[test]
```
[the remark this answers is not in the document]
ionut-arm: Done

src/providers/tpm/utils.rs
ionut-arm: Done

src/providers/tpm/utils.rs
ionut-arm: Done

anta5010 (approved):

MattDavis00 (approved):
LGTM 🚀

PR #520, opened 2021-09-10
Changed list_keys() to be scoped by ApplicationIdentity for the SQLiteKIM

MattDavis00:


ionut-arm:
Would it not be possible to expand the `ManageKeyInfo` trait to have a `fn key_info_manager_type()` there, and each KIM then returns its own type, a bit like the `describe` thing for providers? Asking mostly because a setup like this opens the door for inconsistencies - you _could_ end up with an incorrect `key_info_manager_type` value in the KIM client.

ionut-arm (approved):
LGTM 👍🏻

PR #522, opened 2021-09-10
Added the CanDoCrypto operation as well as fixing some of the other test scripts.

Kakemone:
Added the internal CanDoCrypto as a new capability_discovery file in the pkcs11 provider and made corresponding changes to the relevent files. Also added the end to end tests for CanDoCrypto and fixed the other tests that were still using the old syntax for the usage flags.

Signed-off-by: Sam Davis <sam.davis@arm.com>

### DIFF
--- a/e2e_tests/tests/per_provider/normal_tests/capability_discovery.rs
+++ b/e2e_tests/tests/per_provider/normal_tests/capability_discovery.rs
@@ -1,0 +1,355 @@
+// Copyright 2021 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use e2e_tests::TestClient;
+use parsec_client::core::interface::operations::psa_algorithm::*;
+use parsec_client::core::interface::requests::{Opcode, ResponseStatus};
+use parsec_client::core::interface::operations::psa_key_attributes::*;
+use parsec_client::core::interface::operations::can_do_crypto::CheckType;
+
+#[test]
+fn derive_check() {
+    let mut client = TestClient::new();
+    if !client.is_operation_supported(Opcode::CanDoCrypto) {
+        return
+    }
+
+    let mut usage_flags: UsageFlags = Default::default();
+    let _ = usage_flags.set_derive();
+    let attributes = Attributes {
+        lifetime: Lifetime::Persistent,
+        key_type: Type::RsaPublicKey,
+        bits: 1024,
+        policy: Policy {
+            usage_flags,
+            permitted_algorithms: Algorithm::None,
+        },
+    };
+
+    let status = client.can_do_crypto(CheckType::Derive, attributes);
+
+    assert_eq!(Err(ResponseStatus::PsaErrorNotSupported), status)
+}
+
+#[test]
+fn key_size_check() {
+    let mut client = TestClient::new();
+    if !client.is_operation_supported(Opcode::CanDoCrypto) {
+        return
+    }
+
+    let mut usage_flags: UsageFlags = Default::default();
+    let _ = usage_flags.set_sign_hash();
+    let attributes = Attributes {
+        lifetime: Lifetime::Persistent,
+        key_type: Type::EccPublicKey {curve_family: EccFamily::SecpR1},
+        bits: 8,
+        policy: Policy {
+            usage_flags,
+            permitted_algorithms: Algorithm::AsymmetricSignature(AsymmetricSignature::Ecdsa {
+                hash_alg: Hash::Sha256.into(),
+            }),
+        },
+    };
+
+    let status = client.can_do_crypto(CheckType::Use, attributes);
+
+    assert_eq!(Err(ResponseStatus::PsaErrorNotSupported), status);
+
+    let mut usage_flags: UsageFlags = Default::default();
+    let _ = usage_flags.set_sign_hash();
+    let attributes = Attributes {
+        lifetime: Lifetime::Persistent,
+        key_type: Type::EccPublicKey {curve_family: EccFamily::SecpR1},
+        bits: 256,
+        policy: Policy {
+            usage_flags,
+            permitted_algorithms: Algorithm::AsymmetricSignature(AsymmetricSignature::Ecdsa {
+                hash_alg: Hash::Sha256.into(),
+            }),
+        },
+    };
+
+    let status = client.can_do_crypto(CheckType::Use, attributes);
+
+    assert_eq!(Ok(()), status);
+
+    let mut usage_flags: UsageFlags = Default::default();
+    let _ = usage_flags.set_sign_hash();
+    let attributes = Attributes {
+        lifetime: Lifetime::Persistent,
+        key_type: Type::EccPublicKey {curve_family: EccFamily::SecpR1},
+        bits: 1024,
+        policy: Policy {
+            usage_flags,
+            permitted_algorithms: Algorithm::AsymmetricSignature(AsymmetricSignature::Ecdsa {
+                hash_alg: Hash::Sha256.into(),
+            }),
+        },
+    };
+
+    let status = client.can_do_crypto(CheckType::Use, attributes);
+
+    assert_eq!(Err(ResponseStatus::PsaErrorNotSupported), status)
+}
+
+#[test]
+fn use_check() {
+    let mut client = TestClient::new();
+    if !client.is_operation_supported(Opcode::CanDoCrypto) {
+        return
+    }
+
+    let usage_flags: UsageFlags = Default::default();
+    let attributes = Attributes {
+        lifetime: Lifetime::Persistent,
+        key_type: Type::RsaPublicKey,
+        bits: 256,
+        policy: Policy {
+            usage_flags,
+            permitted_algorithms: Algorithm::None,
+        },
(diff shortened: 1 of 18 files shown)


ionut-arm (commented):
The function is quite convoluted - could you break it down into multiple methods/functions, maybe one for each of `Use`, `Generate`..., and maybe have a separate function for anything that is common?

src/providers/pkcs11/capability_discovery.rs
```diff
@@ -0,0 +1,118 @@
...
+                {
+                    return Err(PsaErrorNotSupported);
+                }
+            } else {
+                if !((*mechanism_info.min_key_size() as u64) <= (attributes.bits as u64)
+                    && (attributes.bits as u64) <= (*mechanism_info.max_key_size() as u64))
+                {
+                    return Err(PsaErrorNotSupported);
+                }
+            }
```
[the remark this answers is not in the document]
Kakemone: I don't think that this will be that much better as <= and >= cannot be applied to Ulongs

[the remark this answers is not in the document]
hug-dev: There is a `From<Ulong> for CK_ULONG` implemented, that could help performing the operations on `CK_ULONG` not caring if that type is `u32` or `u64`?

src/providers/pkcs11/capability_discovery.rs
```diff
@@ -0,0 +1,123 @@
...
+            return Err(PsaErrorNotSupported);
+        } else if !(attributes.policy.usage_flags.decrypt()
+            || attributes.policy.usage_flags.encrypt()
+            || attributes.policy.usage_flags.sign_hash()
+            || attributes.policy.usage_flags.sign_message()
+            || attributes.policy.usage_flags.verify_hash()
+            || attributes.policy.usage_flags.verify_message())
+        {
+            return Err(PsaErrorNotSupported);
+        }
```
hug-dev, line 92: Not sure if we need to care about the `usage_flags` here.

src/providers/pkcs11/capability_discovery.rs
```diff
@@ -0,0 +1,123 @@
...
+                .map_err(to_response_status)?;
+            let mechanism = Mechanism::try_from(attributes.policy.permitted_algorithms)
+                .map_err(to_response_status)?;
+            let mechanism_info: MechanismInfo = self
+                .backend
+                .get_mechanism_info(self.slot_number, mechanism.mechanism_type())
+                .map_err(to_response_status)?;
+            if !(supported_mechanisms.contains(&mechanism.mechanism_type())) {
+                return Err(PsaErrorNotSupported);
+            }
```
hug-dev, line 57: That's pretty clean!
I am thinking, maybe we should put that (and the key size check below) inside the `use_check` function? And replace the content of the `if` with just a call to `use_check`?

In the `match` at the end of the function we can check for `CheckType::Use` that the algorithm is not `None`.

src/providers/pkcs11/capability_discovery.rs
```diff
@@ -0,0 +1,123 @@
...
+                curve_family: EccFamily::SecpR1,
+            } => {
+                if !(supported_ecc_family_sizes.contains(&attributes.bits)) {
+                    return Err(PsaErrorNotSupported);
+                }
+            }
+            Type::RsaKeyPair | Type::RsaPublicKey => (),
+            _ => return Err(PsaErrorNotSupported),
+        }
+        if attributes.policy.permitted_algorithms != Algorithm::None {
```
hug-dev, line 44: In the case it isn't `None`, maybe it would be good to call: `attributes.compatible_with_alg(attributes.policy.permitted_algorithms)?;` to make sure the algorithm is compatible with the key type. That will return `PsaErrorNotPermitted` which can be added as a possible return value from this operation in the contract. [See here](https://docs.rs/psa-crypto/0.9.0/psa_crypto/types/key/struct.Attributes.html#method.compatible_with_alg)

hug-dev (commented):
Thank you for this! It looks as a good start 👌 Added some comments here and there.

e2e_tests/tests/per_provider/normal_tests/capability_discovery.rs
```diff
@@ -0,0 +1,371 @@
...
+            }),
+        },
+    };
+
+    let status = client.can_do_crypto(CheckType::Use, attributes);
+
+    assert_eq!(Ok(()), status);
+
+    let mut usage_flags: UsageFlags = Default::default();
+    let _ = usage_flags.set_verify_hash();
```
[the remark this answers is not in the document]
Kakemone: I included them as they are part of the attributes and will determine whether some operations can be done. I think that they are only check for Use to see if at least one of those flags is set.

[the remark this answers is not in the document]
hug-dev: > to see if at least one of those flags is set

Ah yes that is a good idea to check for that if the algorithm is not `None`! Might be worth to remember it and note it in the contract specs as well.

e2e_tests/tests/per_provider/normal_tests/capability_discovery.rs
```diff
@@ -0,0 +1,371 @@
...
+            usage_flags,
+            permitted_algorithms: Algorithm::AsymmetricSignature(AsymmetricSignature::Ecdsa {
+                hash_alg: Hash::Sha256.into(),
+            }),
+        },
+    };
+
+    let status = client.can_do_crypto(CheckType::Use, attributes);
+
+    assert_eq!(Ok(()), status);
```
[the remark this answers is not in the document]
Kakemone: I haven't implemented any checks for which keys work with which algorithms as I didn't realise that there were any restrictions.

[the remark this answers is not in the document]
Kakemone: Could this be implemented by just using `attributes.compatible_with_alg(attributes.policy.permitted_algorithms)?;` ?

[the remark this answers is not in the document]
hug-dev: Yes I think so! Could be done directly in the implementation I think. See [here](https://github.com/parallaxsecond/parsec/pull/522#discussion_r712160621)

src/providers/pkcs11/capability_discovery.rs
```diff
@@ -0,0 +1,126 @@
...
+        if !(attributes.policy.usage_flags.decrypt()
+            || attributes.policy.usage_flags.encrypt()
+            || attributes.policy.usage_flags.sign_hash()
+            || attributes.policy.usage_flags.sign_message()
+            || attributes.policy.usage_flags.verify_hash()
+            || attributes.policy.usage_flags.verify_message())
+        {
+            return Err(PsaErrorNotSupported);
+        }
+        attributes.compatible_with_alg(attributes.policy.permitted_algorithms)?;
```
hug-dev, line 69: Can you please make a PR on the book operation saying that `NotPermitted` is returned in that case if the alg is not compatible?

src/providers/pkcs11/capability_discovery.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    fn generate_check(attributes: Attributes) -> Result<can_do_crypto::Result> {
+        if !(attributes.key_type == Type::RsaKeyPair
+            || attributes.key_type
+                == (Type::EccKeyPair {
+                    curve_family: EccFamily::SecpR1,
+                }))
+        {
+            return Err(PsaErrorNotSupported);
+        }
```
hug-dev, line 111: If `attributes.policy.permitted_algorithms` is not `None`, I think we should call `use_check` from here too. Same for `import`. I think some tests with non-compatible key type/algorithm are still passing because of that.

Kakemone: Calling use_check causes an issue as it checks if at least on of the the relevant usage flags are set which isn't applicable to import and generate so would it work to just run `attributes.compatible_with_alg(attributes.policy.permitted_algorithms)?;` to check if the type and algorithm match?

hug-dev: Reading the [contract](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/can_do_crypto.html#description) we have written now, I think it should be applicable?

For the Generate, Import and Derive check there is:

> If the `key_algorithm` is not `None`, also perform the `Use` check.

Kakemone: So are we saying that Generate and Import should have also have at least one of the relevant usage flags set? (And Derive but that is never supported by pkcs11)

hug-dev: I think so, if an Algorithm is set, it means that it should be used for at least something

hug-dev (approved):
Thanks for the change. I think since this is a development branch, work can continue on it to make things better little by little in further PRs.

It's really hard to cover everything now, I think we will improve it when adding it as a pre-requisite for our other e2e tests.

ionut-arm (approved):
Let's merge this as is and we'll make note of ways to improve the code in the original issue!

PR #524, opened 2021-09-16
Use as_ptr for TS service name

### LINKED ISSUE #501: Memory leak in TS context
When the `service_locator_query` utility function is called, the service name is constructed as a `CString` and the necessary pointer is obtained through the `into_raw` method. From [that method's documentation](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.into_raw), we should be recapturing ownership of that pointer after use and free it from Rust. However, we don't do any of that and the memory is most likely leaked. We should probably use [`as_ptr`](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.as_ptr) instead.

anta5010:

Fixes #501

### DIFF
--- a/src/providers/trusted_service/context/mod.rs
+++ b/src/providers/trusted_service/context/mod.rs
@@ -74,15 +74,10 @@ impl Context {
 
         info!("Obtaining a crypto Trusted Service context.");
         let mut status = 0;
-        let service_context = unsafe {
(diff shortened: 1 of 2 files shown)


ionut-arm (approved):
Change looks good, but I think good old `rustfmt` doesn't like it 😬

PR #536, opened 2021-09-28
[CryptoAuthLib provider] PsaAeadEncrypt and PsaAeadDecrypt implemented

TomaszPawelecGL:
Signed-off-by: tomek <tomasz.pawelec@globallogic.com>

### DIFF
--- a/e2e_tests/tests/per_provider/normal_tests/aead.rs
+++ b/e2e_tests/tests/per_provider/normal_tests/aead.rs
@@ -37,7 +37,6 @@ const CIPHERTEXT_CCM: [u8; 40] = [
     0xb0, 0x6d, 0x43, 0x13, 0xf0, 0xdb, 0x9b, 0xe9,
 ];
 
-#[cfg(feature = "cryptoauthlib-provider")]
 const CIPHERTEXT_GCM: [u8; 40] = [
     0x7e, 0x3c, 0x18, 0xe5, 0xca, 0xfb, 0xd5, 0x5c, 0x15, 0xda, 0x01, 0x84, 0xff, 0x99, 0x66, 0xc6,
     0xc6, 0x1a, 0x21, 0x7c, 0xae, 0x40, 0x0d, 0x94, 0xa9, 0x43, 0xfa, 0x0f, 0x40, 0x5d, 0xd1, 0xe1,
@@ -51,7 +50,6 @@ const EXPECTED_DECRYPT_CCM: [u8; 56] = [
     0x26, 0x83, 0x8c, 0xf4, 0x18, 0x67, 0x9, 0xf2,
 ];
 
-#[cfg(feature = "cryptoauthlib-provider")]
 const EXPECTED_DECRYPT_GCM: [u8; 56] = [
     0x45, 0x35, 0xd1, 0x2b, 0x43, 0x77, 0x92, 0x8a, 0x7c, 0xa, 0x61, 0xc9, 0xf8, 0x25, 0xa4, 0x86,
     0x71, 0xea, 0x5, 0x91, 0x7, 0x48, 0xc8, 0xef, 0xd3, 0xf1, 0x7e, 0x3c, 0x99, 0x7b, 0x6b, 0xf5,
@@ -436,7 +434,6 @@ fn aead_encrypt_ccm_encrypt_tag_length_too_long() {
     );
 }
 
-#[cfg(feature = "cryptoauthlib-provider")]
 #[test]
 fn aead_encrypt_gcm_encrypt() {
     let key_name = String::from("aead_encrypt_gcm_encrypt");
@@ -463,7 +460,6 @@ fn aead_encrypt_gcm_encrypt() {
     assert_eq!(&CIPHERTEXT_GCM[..], ciphertext.as_slice());
 }
 
-#[cfg(feature = "cryptoauthlib-provider")]
 #[test]
 fn aead_encrypt_gcm_encrypt_not_equal() {
     let key_name = String::from("aead_encrypt_gcm_encrypt_not_equal");
@@ -490,7 +486,6 @@ fn aead_encrypt_gcm_encrypt_not_equal() {
     assert_ne!(&CIPHERTEXT_GCM[..], ciphertext.as_slice());
 }
 
-#[cfg(feature = "cryptoauthlib-provider")]
 #[test]
 fn aead_encrypt_gcm_decrypt() {
     let key_name = String::from("aead_encrypt_gcm_decrypt");
@@ -517,7 +512,6 @@ fn aead_encrypt_gcm_decrypt() {
     assert_eq!(&EXPECTED_DECRYPT_GCM[..], plaintext.as_slice());
 }
 
-#[cfg(feature = "cryptoauthlib-provider")]
 #[test]
 fn aead_encrypt_gcm_decrypt_not_equal() {
     let key_name = String::from("aead_encrypt_gcm_decrypt_not_equal");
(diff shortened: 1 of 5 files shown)


src/providers/cryptoauthlib/aead.rs
```diff
@@ -0,0 +1,155 @@
...
+            aead_alg: AeadWithDefaultLengthTag::Ccm,
+            tag_length,
+        } => Some(*tag_length),
+        Aead::AeadWithShortenedTag {
+            aead_alg: AeadWithDefaultLengthTag::Gcm,
+            tag_length,
+        } => Some(*tag_length),
+        _ => None,
+    }
+}
```
hug-dev, line 34: Reminder to myself that this would be a useful method on [`Aead`](https://docs.rs/psa-crypto/0.9.0/psa_crypto/types/algorithm/enum.Aead.html)

src/providers/cryptoauthlib/aead.rs
```diff
@@ -0,0 +1,155 @@
...
+                op.validate(key_attributes)?;
+
+                let aead_param_gcm = rust_cryptoauthlib::AeadParam {
+                    nonce: op.nonce.to_vec(),
+                    tag_length: Some(tag_length as u8),
+                    additional_data: Some(op.additional_data.to_vec()),
+                    ..Default::default()
+                };
+
+                let aead_param_ccc = aead_param_gcm.clone();
```
hug-dev, line 68: Maybe this line can be put inside the `if` line 70 to only do it when required? Same for decrypt

src/providers/cryptoauthlib/aead.rs
```diff
@@ -0,0 +1,155 @@
...
+                let key_triple =
+                    KeyTriple::new(app_name, ProviderId::CryptoAuthLib, op.key_name.clone());
+                let key_id = self.key_info_store.get_key_id::<u8>(&key_triple)?;
+                let key_attributes = self.key_info_store.get_key_attributes(&key_triple)?;
+                op.validate(key_attributes)?;
+
+                let mut ciphertext: Vec<_> = op.ciphertext.to_vec();
+                let tag: Vec<_> = ciphertext
+                    .drain((ciphertext.len() - tag_length)..)
+                    .collect();
```
hug-dev, line 116: Maybe [`split_off`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.split_off) can be used in an easier way here?

e2e_tests/tests/per_provider/normal_tests/aead.rs
```diff
@@ -213,3 +258,502 @@ fn aead_encrypt_ccm_encrypt_decrypt() {
 
     assert_eq!(&PLAINTEXT[..], plaintext.as_slice());
 }
+
+#[cfg(feature = "cryptoauthlib-provider")]
```
hug-dev, line 262: The other tests in this file do not have that `cfg` gate. It would be nice to remove to make sure those tests also pass on the Mbed Crypto provider (our reference provider implementation). Do you know if that's possible? Feel free to just remove it to see if the CI pass :)

e2e_tests/tests/per_provider/normal_tests/aead.rs
```diff
@@ -213,3 +258,502 @@ fn aead_encrypt_ccm_encrypt_decrypt() {
...
+    {
+        return;
+    }
+    client
+        .import_aes_key(
+            key_name.clone(),
+            KEY_DATA.to_vec(),
+            Aead::AeadWithShortenedTag {
+                aead_alg: AeadWithDefaultLengthTag::Ccm,
+                tag_length: 16,
```
hug-dev, line 278: Isn't 16 the same as the default size?

e2e_tests/tests/per_provider/normal_tests/aead.rs
TomaszPawelecGL: Done.

e2e_tests/tests/per_provider/normal_tests/aead.rs
TomaszPawelecGL: Default tag length is applicable only for Aead::AeadWithDefaultLengthTag(). When it comes to another API Aead::AeadWithShortenedTag() then the tag length must be explicitly specified.  Your comment refer to test aead_encrypt_ccm_encrypt_decrypt_shortened_tag which tests if the second API works well with cryptoauthlib provider.

hug-dev (commented):
Thanks for the changes. Looks good to me and let's see if the CI passes.

src/providers/cryptoauthlib/aead.rs
```diff
@@ -0,0 +1,154 @@
...
+                    get_aead_algorithm(aead_param, &op.alg),
+                    key_id,
+                    &mut ciphertext,
+                ) {
+                    Ok(true) => Ok(psa_aead_decrypt::Result {
+                        plaintext: ciphertext.into(),
+                    }),
+                    Ok(false) => {
+                        error!("aead_decrypt authentication failed");
+                        Err(ResponseStatus::PsaErrorGenericError)
```
ionut-arm, line 139: This should be `PsaErrorInvalidSignature` ([reference](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_aead_decrypt.html#specific-response-status-codes))

akazimierskigl:
Hi, I'm taking over this task from Tomasz since he is leaving our team. Knowledge has been passed and I'll start to work on this task as soon as I'll be done with parsec-interface-rs PR.

ionut-arm (approved):
Looks good, one last comment/question

akazimierskigl:
Ok, so there are few differences for cryptoauthlib and mbed-crypto implementations and I've made branches inside tests to cover it.
For tests:

- aead_encrypt_gcm_encrypt_nonce_too_long
- aead_encrypt_gcm_encrypt_nonce_too_short
- aead_encrypt_gcm_encrypt_tag_length_too_short

Mbed-crypto seems not to check these arguments and allows encryption when cryptoauthlib does not.

For tests:

- aead_encrypt_gcm_encrypt_decrypt_shortened_tag
- aead_encrypt_ccm_encrypt_decrypt_shortened_tag

Mbed-crypto doesn't permit decryption with shortened tag when cryptoauthlib allows it.

In this way all tests are performed for both providers and both of them are passing them. If you have different ideas please share.

src/providers/cryptoauthlib/aead.rs
```diff
@@ -0,0 +1,177 @@
...
+                let key_triple =
+                    KeyTriple::new(app_name, ProviderId::CryptoAuthLib, op.key_name.clone());
+                let key_id = self.key_info_store.get_key_id::<u8>(&key_triple)?;
+                let key_attributes = self.key_info_store.get_key_attributes(&key_triple)?;
+                op.validate(key_attributes)?;
+
+                let aead_param = rust_cryptoauthlib::AeadParam {
+                    nonce: op.nonce.to_vec(),
+                    tag_length: Some(tag_length as u8),
+                    additional_data: Some(op.additional_data.to_vec()),
```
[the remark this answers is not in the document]
akazimierskigl: Rust-cryptoauthlib is checking it and returns error if so.

ionut-arm:
Hmm, apologies, I looked again at the tests, didn't notice the `cfg`s before. Thanks for the explanations!

I think we weren't very clear with the initial comments, that's our fault - the tests that fail on the Mbed Crypto provider should be disabled for it (or only enabled for the CAL provider) and we can raise an issue to look into why it behaves like that. The request to take `cfg` off was mostly to understand if the tests work on the Mbed Crypto provider as well, as they were configured to run against the CAL provider with no comments/documentation as to why that's the case.

akazimierskigl:
@ionut-arm 
I checked with [RFC 3610](https://datatracker.ietf.org/doc/html/rfc3610) and [RFC 5288](https://datatracker.ietf.org/doc/html/rfc5288) once again and it seems to be better to change these tests to be exclusive for CAL for now and add comments about that because mbed-crypto is not behaving correctly with incorrect tag and nonce lengths.

I don't see option to set "do not megre" label for this PR as it usually is so please don't add second approval yet :)

ionut-arm (approved):
Thanks, it looks better now! Are you ok with this being merged?

akazimierskigl:
Yes, it can be merged now.

PR #556, opened 2021-11-12
Re-factor e2e tests to use common key attributes functions

anta5010:

### DIFF
--- a/e2e_tests/tests/per_provider/normal_tests/asym_sign_verify.rs
+++ b/e2e_tests/tests/per_provider/normal_tests/asym_sign_verify.rs
@@ -1,7 +1,8 @@
 // Copyright 2019 Contributors to the Parsec project.
 // SPDX-License-Identifier: Apache-2.0
-use e2e_tests::TestClient;
 use e2e_tests::auto_test_keyname;
+use e2e_tests::TestClient;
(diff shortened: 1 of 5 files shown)


e2e_tests/tests/per_provider/normal_tests/asym_sign_verify.rs
```diff
@@ -891,33 +899,22 @@ fn verify_message_not_permitted() {
...
+        });
+
+    if client.is_operation_supported(Opcode::CanDoCrypto) {
+        assert_eq!(
+            client
+                .can_do_crypto(CheckType::Generate, attributes)
+                .unwrap_err(),
+            ResponseStatus::PsaErrorNotSupported
+        );
+    }
```
ionut-arm, line 919: I wonder if this test should be moved to the file for key generation tests.

ionut-arm (approved):
Looks good, thanks!

e2e_tests/tests/per_provider/normal_tests/asym_sign_verify.rs
anta5010: It can be moved into key_attributes.rs, but as we test the signature algorithm, there is nothing wrong to keep it where it is I think.

PR #563, opened 2021-11-30
[CryptoAuthLib provider] PsaCipherEncrypt and PsaCipherDecrypt implementation

akazimierskigl:
Implementation for PsaCipherEncrypt and PsaCipherDecrypt operations. 
Implementation of end-2-end test cases for Cipher operations.
Updated to newest parsec-client-rust which includes Cipher operations.

Signed-off-by: artur.kazimierski <artur.kazimierski@globallogic.com>

### DIFF
--- a/e2e_tests/tests/per_provider/normal_tests/cipher.rs
+++ b/e2e_tests/tests/per_provider/normal_tests/cipher.rs
@@ -6,14 +6,25 @@ use e2e_tests::TestClient;
 use parsec_client::core::interface::operations::psa_algorithm::Cipher;
 use parsec_client::core::interface::requests::{Opcode, ResponseStatus};
 
+// Test Vector from:
+// https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/AES_CFB.pdf
+// https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/AES_OFB.pdf
+// https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/AES_CTR.pdf
+// https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/AES_ECB.pdf
+// https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/AES_CBC.pdf
+
 #[cfg(feature = "cryptoauthlib-provider")]
 const KEY_DATA: [u8; 16] = [
-    0x41, 0x89, 0x35, 0x1B, 0x5C, 0xAE, 0xA3, 0x75, 0xA0, 0x29, 0x9E, 0x81, 0xC6, 0x21, 0xBF, 0x43,
+    0x2B, 0x7E, 0x15, 0x16, 0x28, 0xAE, 0xD2, 0xA6, 0xAB, 0xF7, 0x15, 0x88, 0x09, 0xCF, 0x4F, 0x3C,
 ];
 
 const IV_SIZE: usize = 16;
 const IV: [u8; IV_SIZE] = [
     0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E, 0x0F,
+];
+#[cfg(feature = "cryptoauthlib-provider")]
+const IV_CTR: [u8; IV_SIZE] = [
+    0xF0, 0xF1, 0xF2, 0xF3, 0xF4, 0xF5, 0xF6, 0xF7, 0xF8, 0xF9, 0xFA, 0xFB, 0xFC, 0xFD, 0xFE, 0xFF,
 ];
 
 const VALID_DATA_SIZE: usize = 64;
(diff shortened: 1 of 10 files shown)


ionut-arm (commented):
Thanks for the patch!

One thing that would be nice to add is a test that imports an AES key, encrypts something with a crypto crate (maybe `ring`?) and then decrypts it with Parsec (or the other way around).

akazimierskigl:
I've delivered fixes for comments to the code and now I will try figure out your proposition about decrypting something which was not encrypted by parsec.

ionut-arm:
Hi @akazimierskigl , I think the more pressing question is whether the `psa_generate_random` implementation from the CAL provider can actually return a vector of 0 bytes instead of completely random, while claiming that the operation was successful. If that's the case, I'd be inclined to move the retry and error-throwing into the `generate_random` implementation - we can hopefully include this in our 0.9.0 release.

akazimierskigl:
@ionut-arm I will clarify this with our CAL expert. Previously I completely ignored the fact that `generate_random` can check itself if vector is cryptographically secure and I added this checking and loop just in case.

Meanwhile I have a first proposition for decryption tests. Instead of encrypting message by external crate we could use already prepared and proven set of IVs, plaintexts and ciphertexts from publications. The same how unit tests were done for rust-cryptoauthlib. https://github.com/PelionIoT/rust-cryptoauthlib/blob/master/cryptoauthlib/src/unit_tests/hw_backend_aes_cipher_stream.rs

ionut-arm:
Sure, that sounds like a sensible plan. It's likely that if we expand the functionality to other providers in the future we'll add more stuff, but hardcoded values is enough for now.

akazimierskigl:
Ok so I have an answer and it is possible to generate vector of zeros by CAL without any error returned. So we need to make decision how cover this extremely rare case.

ionut-arm:
> Ok so I have an answer and it is possible to generate vector of zeros by CAL without any error returned. So we need to make decision how cover this extremely rare case.

I'd say integrating the retry logic from this PR into `psa_generate_random` and sending back a `PSA_ERROR_INSUFFICIENT_ENTROPY` if that fails would work just fine.

akazimierskigl:
@ionut-arm Good idea with only decrypt test cases. One algorithm didn't work properly so implementation of operations is changed too. Code is ready for re-review.

e2e_tests/tests/per_provider/normal_tests/cipher.rs
```diff
@@ -6,15 +6,26 @@ use e2e_tests::TestClient;
 use parsec_client::core::interface::operations::psa_algorithm::Cipher;
 use parsec_client::core::interface::requests::{Opcode, ResponseStatus};
 
+// Test Vector from:
+// https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/AES_CFB.pdf
+// https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/AES_OFB.pdf
+// https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/AES_CTR.pdf
+// https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/AES_ECB.pdf
+// https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Standards-and-Guidelines/documents/examples/AES_CBC.pdf
```
ionut-arm, line 14: This is really nice, thank you for that!

src/providers/cryptoauthlib/generate_random.rs
```diff
@@ -10,29 +10,45 @@ impl Provider {
...
+                    break Ok(psa_generate_random::Result {
+                        random_bytes: random_bytes.into(),
+                    })
+                }
+                false => {
+                    loop_count += 1;
+                    if loop_count < 3 {
+                        continue;
+                    } else {
+                        return Err(ResponseStatus::PsaErrorInsufficientEntropy);
```
ionut-arm, line 48: Could you also put an `error!` message here as well?

src/providers/cryptoauthlib/generate_random.rs
```diff
@@ -10,29 +10,45 @@ impl Provider {
...
+        let mut loop_count = 0u8;
+        // external loop to retry generation if vector is not secure
+        loop {
+            // calculate internal loop count
+            let call_count = (op.size + rust_cryptoauthlib::ATCA_RANDOM_BUFFER_SIZE - 1)
+                / rust_cryptoauthlib::ATCA_RANDOM_BUFFER_SIZE;
+            // internal loop for vector size greater than buffer size
+            for _i in 0..call_count {
+                let mut buffer = Vec::with_capacity(rust_cryptoauthlib::ATCA_RANDOM_BUFFER_SIZE);
+                let err = self.device.random(&mut buffer);
```
ionut-arm, line 23: Hmm, I just realised something - since a bigger request (i.e. one asking for more than `ATCA_RANDOM_BUFFER_SIZE`) can end up calling this `random` more than once, is it possible that one individual call returns all zeroes, but the others return something else? I'm just wondering if the loop checking for zeroes should instead be after each `random` call, instead of the whole combined thing.

ionut-arm (commented):
Thanks for the update, and glad to hear a bug has been squashed before it was even born! 🐛

src/providers/cryptoauthlib/generate_random.rs
akazimierskigl: "is it possible that one individual call returns all zeroes, but the others return something else?"

Yes, every other call of `random` should return different result so it is near impossible to get vector `ATCA_RANDOM_BUFFER_SIZE` * 2 of zeros. This is why I think checking only one time at the end if returned vector is all zeros is needed. Also I think when returned vector has at least one non zero value then is no longer considered "unsafe".

paulhowardarm: Hi @akazimierskigl - I am substituting for @ionut-arm here because he is now on leave until the new year. I am slightly lacking some context, so forgive me if I'm misunderstanding. To resolve this question I think I need to understand the underlying cause of a vector of zeroes from the back-end HRNG. Is it purely a probabilistic thing? Is a sequence of 32 zeroes as likely as any other result? If so, then I'm not sure why it needs any special handling, because you might just as well worry about about any other specific set of values. However, if the source of the zeroes is due to some sort of known transient failure or unready state in the HRNG, then I would agree with Ionut that we really should be checking it on every call to the hardware, because otherwise it is compromising the overall entropy of the result. I hope this makes sense.

akazimierskigl: "Is it purely a probabilistic thing? Is a sequence of 32 zeroes as likely as any other result?"
Yes.
"However, if the source of the zeroes is due to some sort of known transient failure or unready state in the HRNG"
It's not.
I've added handling in a first place only because it was required by cipher algorithm to use random non-zero initialization vector.

paulhowardarm: Okay, thanks. In that case it strikes me that we're worrying over nothing, really. It's a vanishingly-unlikely case. A single check on the collected final result is probably good enough, and for all practical purposes that check would never fail anyway.

ionut-arm: Thanks @paulhowardarm, @akazimierskigl ! I misunderstood what the initial cause for the checks for 0 was, that's a lesson in communication for me :) I thought they'd result from an error in the CAL chip. Shall review once again, apologies for the massive delay this has caused.

ionut-arm (approved):
👍🏻 thanks!

Issue #531, opened 2021-09-22
Making the SQLiteKIM the default


MattDavis00:
This issue serves to track the progress made on making the SQLiteKIM the default.
As part of this, testing should be moving towards using the SQLiteKIM as we want to be testing the most-used and supported configurations.

- [ ] Change the all-providers E2E tests to use the SQLiteKIM. This is current implemented for the normal E2E tests but not for multitenancy etc.
- [ ] Change the [example config](https://github.com/parallaxsecond/parsec/blob/main/config.toml) files to use the SQLiteKIM.

PR #566, opened 2021-12-03
Feature sqlite kim

ionut-arm:
I've taken the `feature-sqlite-kim` branch from the main repo, rebased it on top of our current `main` and this is the result. The PR is fairly large, but luckily it is broken up in a number of commits, that should help in reviewing.

During the rebase I had to make a number of changes - most notably, I had to update the implementations of `CanDoCrypto`, `PrepareKeyAttestation`, and `AttestKey` to use the new KIM client interface, and to use `ApplicationIdentity` instead of `ApplicationName`. The changes have been folded into the original commits.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -125,7 +125,7 @@ while [ "$#" -gt 0 ]; do
             fi
             PROVIDER_NAME=$1
 
-            # If running anything but cargo-check, copy config
+            # Copy provider specific config, unless CI is running `cargo-check` or `sqlite-kim` CI
             if [ "$PROVIDER_NAME" != "cargo-check" ] && [ "$PROVIDER_NAME" != "sqlite-kim" ]; then
                 cp $(pwd)/e2e_tests/provider_cfg/$1/config.toml $CONFIG_PATH
             elif [ "$PROVIDER_NAME" = "sqlite-kim" ]; then
--- a/ci.sh
+++ b/ci.sh
@@ -42,6 +42,7 @@ where PROVIDER_NAME can be one of:
     - cryptoauthlib
     - all
     - coverage
+    - sqlite-kim
 "
 }
 
@@ -118,18 +119,20 @@ while [ "$#" -gt 0 ]; do
         --no-stress-test )
             NO_STRESS_TEST="True"
         ;;
-        mbed-crypto | pkcs11 | tpm | trusted-service | cryptoauthlib | all | cargo-check)
+        mbed-crypto | pkcs11 | tpm | trusted-service | cryptoauthlib | all | cargo-check | sqlite-kim)
             if [ -n "$PROVIDER_NAME" ]; then
                 error_msg "Only one provider name must be given"
             fi
             PROVIDER_NAME=$1
 
             # If running anything but cargo-check, copy config
-            if [ "$PROVIDER_NAME" != "cargo-check" ]; then
+            if [ "$PROVIDER_NAME" != "cargo-check" ] && [ "$PROVIDER_NAME" != "sqlite-kim" ]; then
                 cp $(pwd)/e2e_tests/provider_cfg/$1/config.toml $CONFIG_PATH
+            elif [ "$PROVIDER_NAME" = "sqlite-kim" ]; then
+                cp $(pwd)/e2e_tests/provider_cfg/all/sqlite-kim-all-providers.toml $CONFIG_PATH
             fi
 
-            if [ "$PROVIDER_NAME" = "all" ] || [ "$PROVIDER_NAME" = "cargo-check" ]; then
+            if [ "$PROVIDER_NAME" = "all" ] || [ "$PROVIDER_NAME" = "cargo-check" ] || [ "$PROVIDER_NAME" = "sqlite-kim" ]; then
                 FEATURES="--features=all-providers,all-authenticators"
                 TEST_FEATURES="--features=all-providers"
             else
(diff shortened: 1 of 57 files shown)


ci.sh
```diff
@@ -118,18 +119,20 @@ while [ "$#" -gt 0 ]; do
...
             NO_STRESS_TEST="True"
         ;;
-        mbed-crypto | pkcs11 | tpm | trusted-service | cryptoauthlib | all | cargo-check)
+        mbed-crypto | pkcs11 | tpm | trusted-service | cryptoauthlib | all | cargo-check | sqlite-kim)
             if [ -n "$PROVIDER_NAME" ]; then
                 error_msg "Only one provider name must be given"
             fi
             PROVIDER_NAME=$1
 
             # If running anything but cargo-check, copy config
```
MattDavis00, line 128:
```suggestion
            # Copy provider specific config, unless CI is running cargo-check or sqlite-kim CI
```

e2e_tests/provider_cfg/all/sqlite-kim-all-providers.toml
```diff
@@ -0,0 +1,53 @@
...
+library_path = "/usr/local/lib/softhsm/libsofthsm2.so"
+user_pin = "123456"
+# The slot_number mandatory field is going to replace the following line with a valid number
+# slot_number
+
+[[provider]]
+provider_type = "CryptoAuthLib"
+key_info_manager = "sqlite-manager"
+device_type = "always-success"
+iface_type = "test-interface"
```
MattDavis00, line 47: Don't know about the discussions around the `trusted-services` provider, but was that added into the `all-providers` group in the end? If so, should it be added into here?

MattDavis00 (commented):
I've partially looked through this PR. I'll continue to work through it as and when I get time 👨🏽‍💻

ionut-arm:
> I've partially looked through this PR. I'll continue to work through it as and when I get time 👨🏽‍💻

Thanks for taking the time to review!!

e2e_tests/provider_cfg/all/sqlite-kim-all-providers.toml
ionut-arm: Yeah, will add it

ionut-arm: Ah, I remember now what the problem was: the Trusted Services build we use for testing on the CI transforms the whole TS stack into a library that you link into the executable. Because the crypto Trusted Service is built on top of an implementation of Mbed Crypto, if you try to link both that and the implementation used for the Mbed Crypto provider you end up with:
1) some sort of symbol aliasing, so both the TS provider and the Mbed Crypto provider call the same methods (which we don't want, because they use different versions of the Mbed library, so you get bugs/crashes when stuff gets called with wrong params)
2) because of the way Mbed Crypto stores keys, they end up sharing all their keys on disk. This is a problem because both providers assume that namespace is "free", so they both try to start with a key with ID `1`, and then throw errors when it's not possible

There might've been more problems involved, but for now we're just keeping the Mbed Crypto provider enabled here.

MattDavis00 (changes requested):
Overall looks good to me 👍🏽. Just a few minor comments.

Edit: think I fudged the "request changes" button in VSCode instead of "comment". Just ignore that 👀

src/key_info_managers/mod.rs
```diff
@@ -299,19 +346,19 @@ impl KeyInfoManagerClient {
     /// # Errors
```
MattDavis00, line 346: Line 344 should read `Returns a Vec<ApplicationIdentity> of clients that have keys in this provider`

anta5010 (commented):
Two future questions:
1. Have you re-based this PR on the latest version of the main branch? I can see changes in capability_discovery.rs only for TPM and PKCS11 providers but not for mbed-crypto and TS ones.

2. Have we tested that this new version correctly works with keys created by the old Parsec version if "on-disk-manager" is used?

ionut-arm:
To answer the two outstanding questions:

1. No, I haven't rebased since that one was merged, but will do that after the holiday - shouldn't be a massive change. 
2. Yes, we have stability tests for the on-disk KIM - there's an issue open to create that sort of tests once we release the new KIM too.

e2e_tests/provider_cfg/all/sqlite-kim-all-providers.toml
```diff
@@ -0,0 +1,53 @@
...
+socket_path = "/tmp/parsec.sock"
+
+[authenticator]
+auth_type = "Direct"
+admins = [ { name = "list_clients test" }, { name = "1000" }, { name = "client1" }, { name = "spiffe://example.org/parsec-client-1" } ]
+#workload_endpoint="unix:///tmp/agent.sock"
+
+[[key_manager]]
+name = "sqlite-manager"
+manager_type = "SQLite"
```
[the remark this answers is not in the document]
ionut-arm: Yes, that work's covered under a separate issue: https://github.com/parallaxsecond/parsec/issues/531

ci.sh
```diff
@@ -249,6 +252,22 @@ if [ "$PROVIDER_NAME" = "all" ]; then
     popd
 fi
 
+# Test the SQLite KIM
```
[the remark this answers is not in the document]
ionut-arm: Good point, I think instead of having extra logic we could just change the path for the config file used, but apart from that simply run as if `all` was requested.

paulhowardarm (approved):
Great to see this coming into mainline! It generally looks good. I haven't done a detailed review at the implementation level, but the overall shape of the change appears to match my expectations based on all of the discussions.

ci.sh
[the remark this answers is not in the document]
ionut-arm: Oh, actually no, because the `all` flow does some tests (like the ones I mentioned in reply to your no. 2 question above) which aren't inteded for the SQLite KIM, so we need it to be separate. I could maybe just set a different flag and branch around the stuff that we don't want to run from `all`, but it'd be just as messy.

ionut-arm:
@MattDavis00 @anta5010 all feedback items have been resolved!

MattDavis00 (approved):
Thanks for fixing all of the merge conflicts and updates from `main`. LGTM! 🚀

PR #578, opened 2022-01-18
[CryptoAuthLib provider] PsaRawKeyAgreement operation implementation

akazimierskigl:
PsaRawKeyAgreement operation implementation.
Adapted implementation for key agreement e2e tests.
Fixed error type for PsaImportKey operation.

Signed-off-by: artur.kazimierski <artur.kazimierski@globallogic.com>

### DIFF
--- a/e2e_tests/tests/per_provider/normal_tests/key_agreement.rs
+++ b/e2e_tests/tests/per_provider/normal_tests/key_agreement.rs
@@ -23,11 +23,13 @@ const EXPECTED_OUTPUT_SECPR1: [u8; 32] = [
     0x2f, 0xef, 0x8e, 0x9e, 0xce, 0x7d, 0xce, 0x03, 0x81, 0x24, 0x64, 0xd0, 0x4b, 0x94, 0x42, 0xde,
 ];
 
+#[cfg(not(feature = "cryptoauthlib-provider"))]
 const OUR_KEY_DATA_BRAINPOOL_R1: [u8; 32] = [
     0x81, 0xdb, 0x1e, 0xe1, 0x00, 0x15, 0x0f, 0xf2, 0xea, 0x33, 0x8d, 0x70, 0x82, 0x71, 0xbe, 0x38,
     0x30, 0x0c, 0xb5, 0x42, 0x41, 0xd7, 0x99, 0x50, 0xf7, 0x7b, 0x06, 0x30, 0x39, 0x80, 0x4f, 0x1d,
 ];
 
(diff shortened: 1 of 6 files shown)


e2e_tests/tests/per_provider/normal_tests/key_agreement.rs
```diff
@@ -84,6 +87,13 @@ fn raw_key_agreement_secpr1() {
         return;
     }
 
+    // Section needed for crytpoauthlib as first ECC slot has unusual access behavior
+    // and user need to be aware of that.
```
ionut-arm, line 91: Could you give more details (not necessarily in the code - a comment here will do) about what that unusual access behavior is? Is this something documented for CAL?

ionut-arm (commented):
Hi, thanks for the patch! Just one question, waiting for the PR to go beyond draft to do a proper review (though it looks alright at first read).

e2e_tests/tests/per_provider/normal_tests/key_agreement.rs
akazimierskigl: I have expanded this comment section a bit as I have learned that it depends on chip's configuration - which can be changed - and our current local testing configuration requires this code section. Parsec's universality is not helping in this case because we can't pass as parameter slot number or what type of ECC slot user would like to select.

I will try to figure out if we can implement some function to handle this case.

ionut-arm:
A more generic question: we're planning to release version 1.0.0 of the Parsec service soon, would you like/need this patch to go in before that release? If not, it would go out in 1.1.0, though I'm not certain when that would be, as the roadmap for that release has not been drawn.

akazimierskigl:
> A more generic question: we're planning to release version 1.0.0 of the Parsec service soon, would you like/need this patch to go in before that release? If not, it would go out in 1.1.0, though I'm not certain when that would be, as the roadmap for that release has not been drawn.

No, it's not needed for release. Full functionality is already there. I need to fix this CI issue only and it is ready for release.

akazimierskigl:
@ionut-arm Hi, I just want to let you know that our team is ending this project assignment with the end of January so if you want any changes to the code from us please make comments before Monday.

ionut-arm:
> @ionut-arm Hi, I just want to let you know that our team is ending this project assignment with the end of January so if you want any changes to the code from us please make comments before Monday.

Hi, thanks for the heads-up, I've looked through the code once, will review again today and see if there's anything sticking out, otherwise I'll merge.

ionut-arm (approved):
LGTM, thanks for all the patches you've contributed!!

PR #580, opened 2022-01-28
Add asymmetric encryption to TS provider

ionut-arm:
Adding support for asymmetric encryption to the Trusted Services
provider. However, the version of the TS implementation used for testing
has a bug which prevents us from properly testing using some of the
algorithms.

This is based on changes initially crated in #406


anta5010:
Some CI tests also need to be changed from PsaErrorNotSupported to OK

anta5010 (approved):

PR #584, opened 2022-02-21
Updates for Release Candidate 2

ionut-arm:
This commit makes a few cargo-audit-related changes:
* updates the lock file with newer, un-yanked versions of
futures-... crates
* Updates the cargo audit config file to ignore the
`chrono` and `time` issues.
* Updates the security policy to lay out our approach to
handling vulnerability reports.
* Remove the `cryptoauthlib-provider` from `all-providers`
* Add cross-compilation testing for the TS provider (for `aarch64-linux-gnu`)

### DIFF
--- a/e2e_tests/docker_image/parsec-service-test-cross-compile.Dockerfile
+++ b/e2e_tests/docker_image/parsec-service-test-cross-compile.Dockerfile
@@ -2,6 +2,26 @@
 # SPDX-License-Identifier: Apache-2.0
 FROM ghcr.io/parallaxsecond/parsec-service-test-all
 
+# Install aarch64-none-linux-gnu cross compilation toolchain
+RUN wget https://developer.arm.com/-/media/Files/downloads/gnu-a/9.2-2019.12/binrel/gcc-arm-9.2-2019.12-x86_64-aarch64-none-linux-gnu.tar.xz?revision=61c3be5d-5175-4db6-9030-b565aae9f766 -O aarch64-gcc.tar.xz
+RUN tar --strip-components=1 -C /usr/ -xvf aarch64-gcc.tar.xz
+RUN rm aarch64-gcc.tar.xz
+
+# Install Trusted Services lib compiled for aarch64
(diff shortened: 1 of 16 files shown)


anta5010 (approved):

e2e_tests/docker_image/parsec-service-test-cross-compile.Dockerfile
```diff
@@ -2,6 +2,26 @@
...
+# Install mock Trusted Services
+# Setup git config for patching dependencies
+RUN git config --global user.email "some@email.com"
+RUN git config --global user.name "Parsec Team"
+RUN git clone https://git.trustedfirmware.org/TS/trusted-services.git --branch integration \
+    && cd trusted-services \
+    && git reset --hard 389b50624f25dae860bbbf8b16f75b32f1589c8d
+# Install correct python dependencies
+RUN pip3 install -r trusted-services/requirements.txt
+RUN cd trusted-services/deployments/libts/arm-linux/ \
```
anta5010, line 19: I'm not sure I understand what you're trying to archive here. The `arm-linux` deployment doesn't contain a mock TS like `linux-pc` does. So, it can't be used for TS functional testing. What is the point of building it then? 
Is it just to build libts.so which is required for building Parsec TS provider and no any functional tests will be run against it? Then the comment in the line 10 should be updated.

ionut-arm: Yeah, the cross-compilation "testing" we do is just that - to cross-compile the service, nothing more. I'll change that comment to reflect what we're doing.

anta5010 (approved):

paulhowardarm (approved):
I am assuming that the CAL provider **can** still be explicitly built as a feature and used as normal, despite the removal from `all-providers`? That's what it looks like to me, but I want to double check because I don't think we should be deactivating the provider completely in such a significant milestone release.

We should aim to get this back into `all-providers` ASAP, but I see we have an issue to track that, so that's good.

Issue #593, opened 2022-03-22
Implement get_random in TPM provider


anta5010:
Parsec GreenGrass plugin requires get_random functionality from Parsec. TPM provider is missing it.

ionut-arm:
Done in #595

PR #595, opened 2022-03-25
Add generate random support into TPM provider

anta5010:
Implements #593

### DIFF
--- a/src/providers/tpm/generate_random.rs
+++ b/src/providers/tpm/generate_random.rs
@@ -1,0 +1,30 @@
+// Copyright 2022 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+use super::Provider;
+use parsec_interface::operations::psa_generate_random;
+use parsec_interface::requests::Result;
+
+impl Provider {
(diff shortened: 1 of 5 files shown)


src/providers/tpm/generate_random.rs
```diff
@@ -0,0 +1,30 @@
...
+
+        let mut esapi_context = self
+            .esapi_context
+            .lock()
+            .expect("ESAPI Context lock poisoned");
+
+        let random_bytes = esapi_context
+            .as_mut()
+            .execute_without_session(|esapi_context| esapi_context.get_random(size))
+            .expect("Failed to get random bytes")
```
ionut-arm, line 22: Probably better to convert this error and send it back up instead of panicking here.

ionut-arm (commented):
Looks good, only got one request

src/providers/tpm/generate_random.rs
anta5010: updated

Kakemone (approved):

ionut-arm:
cc @GowthamSureshKumar @MohamedAsaker

PR #608, opened 2022-05-13
Recognise a PKCS11 hardware token with its serial number instead of slot number

mohamedasaker-arm:
Signed-off-by: mohoma01 <mohamed.omarasaker@arm.com>


config.toml
```diff
@@ -150,7 +150,10 @@ key_info_manager = "sqlite-manager"
 # (Required for this provider) Path to the location of the dynamic library loaded by this provider.
 # For the PKCS 11 provider, this library implements the PKCS 11 API on the target platform.
 #library_path = "/usr/local/lib/softhsm/libsofthsm2.so"
-# (Optional) PKCS 11 slot that will be used by Parsec.
+# (Optional) PKCS 11 serial number of the token that will be used by Parsec.
+# If the token serial number is entered, the slot has the token with the provided serial number will be used, and a warning should be issued in case of a mismatch
```
ionut-arm, line 154: Trying to make this a bit clearer:
```suggestion
# If the token serial number is entered, then the slot that has the provided serial number will be used. Otherwise, if both `serial_number` and `slot_number` are given but do not match, a warning is issued and serial number takes precedence.
```

src/providers/pkcs11/mod.rs
```diff
@@ -479,27 +487,75 @@ impl ProviderBuilder {
...
+                "Failed retrieving a valid slot with an initialized token",
+                e
+            );
+            Error::new(
+                ErrorKind::InvalidData,
+                "Failed retrieving a valid slot with an initialized token",
+            )
+        })?;
+        let slot_number = match (self.serial_number, self.slot_number) {
+            (Some(serial_number), _given_slot) => {
```
ionut-arm, line 501: The `_given_slot` name shouldn't start with `_` if you actually use it below, just `given_slot` should do.

src/providers/pkcs11/mod.rs
```diff
@@ -479,27 +487,75 @@ impl ProviderBuilder {
...
+                        format_error!("Failed retrieving token info", e);
+                        Error::new(ErrorKind::InvalidData, "Failed retrieving token info")
+                    })?;
+                    let sn =
+                        String::from_utf8(current_token.serialNumber.to_vec()).map_err(|e| {
+                            format_error!("Failed retrieving token serial number", e);
+                            Error::new(
+                                ErrorKind::InvalidData,
+                                "Failed retrieving token serial number",
+                            )
```
ionut-arm, line 514:
```suggestion
                            format_error!("Failed parsing token serial number", e);
                            Error::new(
                                ErrorKind::InvalidData,
                                "Failed parsing token serial number",
                            )
```

Though this will go away once combined with the `cryptoki` update.

src/providers/pkcs11/mod.rs
```diff
@@ -479,27 +487,75 @@ impl ProviderBuilder {
...
+                let slot = Slot::try_from(slot_number).or_else(|_| {
                     Err(Error::new(
                         ErrorKind::InvalidData,
                         "cannot convert slot value",
                     ))
                 })?;
+                if !slots.contains(&slot) {
+                    return Err(Error::new(
+                        ErrorKind::InvalidInput,
+                        "No availble slot with the given number",
```
ionut-arm, line 549:
```suggestion
                        "No available slot with the given number",
```

ionut-arm (commented):
Thanks, looks good overall.

Could you also add some config tests: you're already covering the "both serial and slot numbers given" option in the main TOML for PKCS11, and there's already a test there where no slot or serial number is given, you'll need one test for "just a slot number", one for "just a serial number", and one for "serial and slot numbers, but they mismatch".

Also, we've not decided - are we merging mine or yours first? The 2nd to be merged will have to make some additional changes to their PR - I'm happy to do that, but I'll let you decide.

mohamedasaker-arm:
Thanks for the Review @ionut-arm!
Regarding which PR should go first, I am happy either way, as I believe the conflicts will be minor.
I think the ready one should go first 😄!

anta5010 (approved):

PR #609, opened 2022-05-25
Fix `cargo-audit` TOML config

ionut-arm:
For the motivation to make this change, see [this](https://github.com/parallaxsecond/parsec/actions/runs/2374951502) run.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -22,9 +22,6 @@ show_tree = true # Show inverse dependency trees along with advisories
 [target]
 os = "linux" # Ignore advisories for operating systems other than this one
 
-[packages]
-source = "all" # "all", "public" or "local"
-
 [yanked]
 enabled = true # Warn for yanked crates in Cargo.lock
 update_index = true # Auto-update the crates.io index 


paulhowardarm (commented):
Could you spell out why we're doing this, rather than just linking to some logs? The link is good for illustration, but doesn't really explain what happened. Do I deduce correctly that `packages` is a config setting that is no longer supported by the tool?

ionut-arm:
> Could you spell out why we're doing this, rather than just linking to some logs? The link is good for illustration, but doesn't really explain what happened. Do I deduce correctly that `packages` is a config setting that is no longer supported by the tool?

To answer the last question first: yes, the tool got rid of the `packages` setting in a [recent PR](https://github.com/rustsec/rustsec/pull/541). Because we always install `cargo-audit` in the nightly run the most recent version is used, so now our config file is not compatible with the tool! ☹️

paulhowardarm (approved):
Cool, thanks. LGTM.

mohamedasaker-arm (approved):

gowthamsk-arm (approved):

Issue #463, opened 2021-07-07
Change default socket path for E2E tests


ionut-arm:
Our end-to-end test framework uses the Rust client to handle requests and thus follows the model used by that library for deciding where to look for the service socket. By default the normal Rust client expects it to be in `/run/parsec/parsec.sock`, but in most test scenarios we use `/tmp/parsec.sock` for ease of use.

On the CI this is accomplished by setting the `PARSEC_SERVICE_ENDPOINT` environment variable to `unix:/tmp/parsec.sock` in the Docker images. However, for the sake of local testing, it would be better to simply set that variable to `unix:/tmp/parsec.sock` if it isn't already set to something by default. This would remove the need for the user to set it by hand or otherwise face a lot of `InvalidSocketAddress`.

The `RawRequestClient` path should also be more versatile - at the moment it simply uses the path above, but hardcoded. Preferably, the environment variable should be checked before.

PR #610, opened 2022-05-26
Change default socket path for E2E tests

gowthamsk-arm:
Fixes: #463 

This MR will allow the users to run the e2e tests without mandatorily setting the socket path environment variable. If unset a default path /tmp/parsec.sock will be used. 

### DIFF
--- a/e2e_tests/src/lib.rs
+++ b/e2e_tests/src/lib.rs
@@ -54,6 +54,10 @@ use parsec_client::core::interface::operations::psa_key_attributes::{
 use parsec_client::core::interface::requests::{Opcode, ProviderId, ResponseStatus, Result};
 use parsec_client::error::Error;
 use std::collections::HashSet;
+use std::env;
+use std::sync::Once;
+
+static INIT: Once = Once::new();
 
 /// Client structure automatically choosing a provider and high-level operation functions.
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


gowthamsk-arm:
Will fix the CI issue and push new commits.

e2e_tests/src/lib.rs
```diff
@@ -82,6 +83,11 @@ impl TestClient {
             env_logger::try_init();
         }
 
+        //Check if the environment variable is set, if not use default path
+        if Err(env::VarError::NotPresent) == env::var("PARSEC_SERVICE_ENDPOINT") {
+            env::set_var("PARSEC_SERVICE_ENDPOINT", "unix:/tmp/parsec.sock");
+        }
```
ionut-arm, line 89: I think this approach might be dangerous in terms of threading, see the docs [here](https://doc.rust-lang.org/std/env/fn.set_var.html). Perhaps you can use something like [`static_init`](https://crates.io/crates/static_init) to make sure it only gets called once?

ionut-arm (commented):
Looks good overall, I think there's just some complaints from the linter there.

e2e_tests/src/lib.rs
gowthamsk-arm: Thanks for the comment Ionut. Will fix this :)

gowthamsk-arm:
Rebased on top of main.

ionut-arm (approved):
👍🏻 Thanks

Issue #534, opened 2021-09-28
Create script for Quickstart package


ionut-arm:
Official [Github releases](https://github.com/parallaxsecond/parsec/releases) of the service contain a tar gzip file used for quickstart testing. The contents and steps are described in the [Quickstart guide](https://parallaxsecond.github.io/parsec-book/getting_started/linux_x86.html). The process of generating the `.tar.gz` is currently manual, but a script could be included for easier processing in the future.

If we're confident enough, we could even make use of Github Actions to completely automate it.

anta5010:
parsec-cli-tests.sh can be included into the archive and be mentioned in the Quickstart guide

mohamedasaker-arm:
@anta5010 I wonder if we can modify the packaged parsec-cli-tests.sh. more precisely this line 

`PARSEC_SERVICE_ENDPOINT="${PARSEC_SERVICE_ENDPOINT:-unix:/run/parsec/parsec.sock}"`

The reason is, that in the release package parsec uses the socket path ./parsec.sock and I think unifying the default socket path would enhance the OOB experience.
Please share your insights

anta5010:
> @anta5010 I wonder if we can modify the packaged parsec-cli-tests.sh. more precisely this line
> 
> `PARSEC_SERVICE_ENDPOINT="${PARSEC_SERVICE_ENDPOINT:-unix:/run/parsec/parsec.sock}"`
> 
> The reason is, that in the release package parsec uses the socket path ./parsec.sock and I think unifying the default socket path would enhance the OOB experience. Please share your insights

Sorry, I don't understand what you mean by "the release package parsec uses the socket path ./parsec.sock".

Parsec service built from sources uses `/run/parsec/parsec.sock` by default. See https://github.com/parallaxsecond/parsec/blob/87852b7723ffffc63b6266a601a99579e13ae768/src/front/domain_socket.rs#L23

And the same parameter value is used by default in the default Parsec config file. See https://github.com/parallaxsecond/parsec/blob/87852b7723ffffc63b6266a601a99579e13ae768/config.toml#L52

So, most likely you're using a Parsec config file where `socket_path` is set to a non-default value for whatever reason. I think instead of changing  parsec-cli-tests.sh you can either:
- fix the config file you're using to use the default value or
- set PARSEC_SERVICE_ENDPOINT env variable to a non-default value you require. For example `PARSEC_SERVICE_ENDPOINT='./parsec.sock' parsec-cli-tests.sh`

[...]

PR #612, opened 2022-06-07
Add a script to create the Quickstart package

mohamedasaker-arm:
Fixes: #534

### DIFF
--- a/packaging_assets/package.sh
+++ b/packaging_assets/package.sh
@@ -13,7 +13,7 @@ JOBS_NUMBER=1
 SCRIPT_PATH=$(realpath $0)
 ASSETS_DIR=$(dirname $SCRIPT_PATH)
 PARSEC_DIR=$(dirname $ASSETS_DIR)
-VERSION=$(cargo metadata --format-version=1 --no-deps --offline | python3 -c "import sys, json; data=json.load(sys.stdin); print([pkg['version'] for pkg in data['packages'] if pkg['name'] == 'parsec-service'][0])")
+VERSION=$(cargo metadata --format-version=1 --no-deps --offline | jq -r '.packages[0].version')
 
 # Usage
 USAGE_STR=\
@@ -34,24 +34,24 @@ do
     esac
 done
 
-check_mandatory_flags() {
-    echo "Checking mandatoy flags..."
+check_version() {
+    echo "Checking version"
     if [ -z "$VERSION" ];then
-            echo 'Missing the version please run with -h for help' >&2
+            echo "Couldn't extract the version!" >&2
             exit 1
     fi
 }
 
 check_release_tag() {
-    CURRENT_TAG=$(git name-rev --tags HEAD | cut -d " " -f 2)
-    LATTEST_TAG=$(git tag | tail -n 1)
+    CURRENT_TAG=$(git name-rev --tags HEAD | cut -d "/" -f 2)
+    LATTEST_TAG=$(git tag --sort=committerdate | tail -1)
     if [ -z "$LATTEST_TAG" ];then
         echo "Warning:No tags"
     fi
     if [ "$LATTEST_TAG" == "$CURRENT_TAG" ]; then
         echo "Packaging release tag: $LATTEST_TAG"
     else
-        echo "Warning: The current HEAD isn't tagged"
+        echo "Warning: The current HEAD does't match the latest tagged"
         echo "Warning: Please checkout the latest tag : $LATTEST_TAG"
         read  -n 1 -p "Do you want to continue anyway [y/n]?" choice
         if [ "$choice" != "y" ]; then
@@ -63,16 +63,12 @@ check_release_tag() {
 cleanup()
 {
     echo "Clean up"
-    rm -rf $WORK_DIR/$PACKAGE_DIR
     rm -rf $WORK_DIR
 }
 
 pre_package() {
     # Construct package name
     PACKAGE_DIR=quickstart-$VERSION-${OS}_$ARCH
-
-    # Create a temp work directory
-    mkdir -p $WORK_DIR
 
     # Create a temp work directory for parsec service
     mkdir -p $WORK_DIR/parsec
@@ -93,6 +89,11 @@ build_parsec_tool() {
     echo "Building Parsec-tool"
 
     git clone https://github.com/parallaxsecond/parsec-tool $WORK_DIR/parsec-tool
+
+    cd $WORK_DIR/parsec-tool
+    git checkout $(git tag --sort=committerdate | tail -1)
+    cd -
+
     cargo build -j $JOBS_NUMBER --release --manifest-path $WORK_DIR/parsec-tool/Cargo.toml
 }
 
@@ -112,7 +113,7 @@ echo "Packaging started..."
 
 trap cleanup EXIT
 
-check_mandatory_flags
+check_version
 check_release_tag
 cleanup
 pre_package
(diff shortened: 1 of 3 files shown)


packaging_assets/package.sh
```diff
@@ -0,0 +1,127 @@
...
+    rm -rf $WORK_DIR/$PACKAGE_DIR
+    rm -rf $WORK_DIR
+}
+
+pre_package() {
+    # Construct package name
+    PACKAGE_DIR=quickstart-$VERSION-${OS}_$ARCH
+
+    # Create a temp work directory
+    mkdir -p $WORK_DIR
```
anta5010, line 75: This is not required because `mkdir -p $WORK_DIR/parsec` below will create $WORK_DIR

packaging_assets/package.sh
```diff
@@ -0,0 +1,127 @@
...
+        if [ "$choice" != "y" ]; then
+            exit 1
+        fi
+    fi
+}
+
+cleanup()
+{
+    echo "Clean up"
+    rm -rf $WORK_DIR/$PACKAGE_DIR
```
anta5010, line 66: this line is not required, because the next rm will delete it anyway

packaging_assets/package.sh
```diff
@@ -0,0 +1,127 @@
...
+check_release_tag() {
+    CURRENT_TAG=$(git name-rev --tags HEAD | cut -d " " -f 2)
+    LATTEST_TAG=$(git tag | tail -n 1)
+    if [ -z "$LATTEST_TAG" ];then
+        echo "Warning:No tags"
+    fi
+    if [ "$LATTEST_TAG" == "$CURRENT_TAG" ]; then
+        echo "Packaging release tag: $LATTEST_TAG"
+    else
+        echo "Warning: The current HEAD isn't tagged"
```
anta5010, line 54: this comment might be incorrect if the current tag is defined but different to the latest tag.

packaging_assets/package.sh
```diff
@@ -0,0 +1,127 @@
...
+    echo "Checking mandatoy flags..."
+    if [ -z "$VERSION" ];then
+            echo 'Missing the version please run with -h for help' >&2
+            exit 1
+    fi
+}
+
+check_release_tag() {
+    CURRENT_TAG=$(git name-rev --tags HEAD | cut -d " " -f 2)
+    LATTEST_TAG=$(git tag | tail -n 1)
```
anta5010, line 47: This command won't always give you the latest tag. The current Parsec latest tag is `1.0.0` but this command prints different:
```
$ git tag | tail -n 1
1.0.0-rc3
```

packaging_assets/package.sh
```diff
@@ -0,0 +1,127 @@
...
+check_mandatory_flags() {
+    echo "Checking mandatoy flags..."
+    if [ -z "$VERSION" ];then
+            echo 'Missing the version please run with -h for help' >&2
+            exit 1
+    fi
+}
+
+check_release_tag() {
+    CURRENT_TAG=$(git name-rev --tags HEAD | cut -d " " -f 2)
```
anta5010, line 46: these commands will print `tags/\<tag`>` and 'git tag' command print tags without `tags/` prefix

packaging_assets/package.sh
```diff
@@ -0,0 +1,127 @@
...
+
+check_mandatory_flags() {
+    echo "Checking mandatoy flags..."
+    if [ -z "$VERSION" ];then
+            echo 'Missing the version please run with -h for help' >&2
+            exit 1
+    fi
+}
+
+check_release_tag() {
```
anta5010, line 45: I don't quite understand we try to archive with this function.
For the package name we use the version number we get from metadata anyway.
So, instead of checking for latest tags I think we need to either:
1. Check that the current tag is equal to the version from metadata or
2. ignore tags and include the current commit hash into the package name in addition to the version.

anta5010 (changes requested):

packaging_assets/quickstart_config.toml
```diff
@@ -0,0 +1,23 @@
+# See https://parallaxsecond.github.io/parsec-book/parsec_service/configuration.html for a full
+# example.
+
+[core_settings]
+log_level = "info"
+allow_root = true
```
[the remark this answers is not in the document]
mohamedasaker-arm: Not sure why it's used here in the first place, I've copied the config file in the recent quickstart package.
one reason I can think of is that the quickstart package would be the first thing a user may use.
so we don't want to make any assumptions unless documented.

packaging_assets/package.sh
anta5010: you don't need two `cut`. one will do the trick:
`git name-rev --tags HEAD | cut -d "/" -f 2`

packaging_assets/package.sh
mohamedasaker-arm: Since it's a manual step, then by mistake a package might upload which isn't from the latest release.

packaging_assets/package.sh
anta5010: Use:
`git describe --tags --abbrev=0 `

anta5010: or `git tag --sort=committerdate | tail -1`

packaging_assets/quickstart_config.toml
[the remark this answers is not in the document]
ionut-arm: I don't think it makes much of a difference. Perhaps some people will want to run it with a root account, don't know how likely that is. We can either keep or remove this - the quickstart bundle isn't meant to be a secure deployment anyway.

packaging_assets/package.sh
```diff
@@ -0,0 +1,123 @@
...
+    echo "Building Parsec"
+
+    CARGO_TARGET_DIR=$WORK_DIR/parsec/ cargo build -j $JOBS_NUMBER --release --features mbed-crypto-provider --manifest-path $PARSEC_DIR/Cargo.toml
+}
+
+build_parsec_tool() {
+    # Package Parsec-tool
+    echo "Building Parsec-tool"
+
+    git clone https://github.com/parallaxsecond/parsec-tool $WORK_DIR/parsec-tool
```
anta5010, line 91: do we want to check out the latest tag for `parsec-tool` demo as well?

mohamedasaker-arm: Good point, I think we need to agree on how parsec and parsec-tool synced together.
Taking the latest release makes sense, but when they are not synced, which version should we pick?
Should it be (the version) provided in case the latest release of the parsec-tool is not synced?

mohamedasaker-arm: As discussed, Latest tag of parsec-tool will be used

anta5010 (approved):

packaging_assets/package.sh
```diff
@@ -0,0 +1,128 @@
...
+    cd $WORK_DIR/parsec-tool
+    git checkout $(git tag --sort=committerdate | tail -1)
+    cd -
+
+    cargo build -j $JOBS_NUMBER --release --manifest-path $WORK_DIR/parsec-tool/Cargo.toml
+}
+
+collect() {
+    # Include Parsec service
+    cp $WORK_DIR/parsec/release/parsec $WORK_DIR/$PACKAGE_DIR/
```
ionut-arm, line 102: Is this supposed to include the Parsec service release-build binary? Would've thought you need `./target/release` here for that.

packaging_assets/package.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
```
ionut-arm, line 1: I'd recommend setting some flags for the script to make sure nothing fails silently:
```bash
set -euf -o pipefail
```

ionut-arm (commented):
Thanks for the change! Got a few comments, but perhaps some of them are just me misunderstanding stuff

Issue #594, opened 2022-03-22
Implement get_random in the PKCS11 provider


ionut-arm:
There's a need for get_random from the GreenGrass integration. This should be implemented in the PKCS11 provider.

ionut-arm:
Note: this involves a bump in our dependency on `cryptoki`, since the version we use now does not offer support for RNG. The actual dependency bump will involve more work than adding the new operation will, most likely, as the API has changed quite a lot.

PR #613, opened 2022-06-09
Implement get_random in the PKCS11 provider

gowthamsk-arm:
Fixes: [594](https://github.com/parallaxsecond/parsec/issues/594)

### DIFF
--- a/e2e_tests/tests/all_providers/normal.rs
+++ b/e2e_tests/tests/all_providers/normal.rs
@@ -88,6 +88,8 @@ fn list_opcodes() {
         Opcode::PsaExportPublicKey,
         Opcode::PsaAsymmetricDecrypt,
         Opcode::PsaAsymmetricEncrypt,
+        Opcode::CanDoCrypto,
+        Opcode::PsaGenerateRandom,
     ];
     let mut mbed_crypto_opcodes = vec![
         Opcode::CanDoCrypto,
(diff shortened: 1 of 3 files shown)


src/providers/pkcs11/generate_random.rs
```diff
@@ -0,0 +1,29 @@
...
+use parsec_interface::requests::{ResponseStatus, Result};
+use std::convert::TryFrom;
+
+impl Provider {
+    pub(super) fn psa_generate_random_internal(
+        &self,
+        op: psa_generate_random::Operation,
+    ) -> Result<psa_generate_random::Result> {
+        let length = u32::try_from(op.size).or_else(|_| {
+            let error = ResponseStatus::ResponseTooLarge;
```
ionut-arm, line 15: I don't think this is the appropriate response status here - `ResponseTooLarge` is about the size of the encoded response payload being larger than the limit set by the admin (see [reference doc](https://parallaxsecond.github.io/parsec-book/parsec_client/status_codes.html)). This should probably be `PsaGenericError`, looking at the other error codes defined by the PSA spec I can't really associate either of them with this case.

src/providers/pkcs11/generate_random.rs
```diff
@@ -0,0 +1,29 @@
...
+            format_error!("Generate random status", error);
```
ionut-arm, line 16: Given that we can't be very specific with the error code above, this should probably just be something like
```
error!("Requested size is too large");
```

src/providers/pkcs11/generate_random.rs
ionut-arm: Ah, I see something similar is happening in the Mbed Crypto provider, just checking it against the `GlobalConfig::buffer_size_limit()` - I'm wondering now if we should implement that check at `BackendHandler` level, because the `BackendHandler` knows what the actual size of the reponse payload is, and should send back `ResponseTooLarge` when the size is exceeded. I'll raise an issue for this.

ionut-arm (commented):
Thanks for the change! 👍🏻

src/providers/pkcs11/generate_random.rs
gowthamsk-arm: Will update it with PsaGenericError return value.

src/providers/pkcs11/generate_random.rs
gowthamsk-arm: Will update the print log.

ionut-arm (approved):
🙏🏻

e2e_tests/tests/all_providers/normal.rs
```diff
@@ -116,6 +116,7 @@ fn list_opcodes() {
 
     let mut crypto_providers_hsm = HashSet::from_iter(common_opcodes);
     let _ = crypto_providers_hsm.insert(Opcode::CanDoCrypto);
+    let _ = crypto_providers_hsm.insert(Opcode::PsaGenerateRandom);
```
anta5010, line 119: All providers support genRandom now. So, it can be moved into `common_opcodes`. And `CanDoCrypto` as well.

anta5010 (approved):
LGTM. A small comment can be ignored if you wish.

e2e_tests/tests/all_providers/normal.rs
gowthamsk-arm: Okay. Will include them in common opcodes.

PR #620, opened 2022-06-25
Send back PsaErrorInvalidPadding when needed

### LINKED ISSUE #619: RSA padding oracle issue
I've discovered a bug that can turn out to be a security vulnerability for Parsec clients that use Parsec in a very specific way.

TL;DR: **If any Parsec client uses the service for TLS 1.2 pre-master secret decryption with RSA PKCS#1 v1.5, they're vulnerable to Bleichenbacher's padding oracle attack.**

## The bug

The Parsec service, following the lead of the PSA Crypto, supports two algorithms for asymmetric encryption/decryption: RSA OAEP, and RSA PKCS#1 v1.5. While both algorithms deal with padding, the current issue only applies to RSA PKCS#1 v1.5, so this issue touches on the use of that one only.

Whenever we complete an `psa_asymmetric_decrypt` operation, we perform a conversion from the provider-specific error code to the Parsec-specific error code (i.e., the code that will be sent to the Parsec client). Given that these error codes are not exactly "standardised" across the providers, we decided (on first implementation) to properly map only the codes that seem most relevant and that have an equivalent in the PSA Crypto API, and for the rest to return a generic error. The bug concerns these mappings (or lack thereof) - we do not properly map the error codes signifying an invalid padding to `PsaErrorInvalidPadding` for the TPM and PKCS11 providers. This means that Parsec clients would instead get a generic error code instead when a padding is found to be invalid.

## The security issue

> Ok, so why is an invalid padding important in the case of RSA PKCS#1 v1.5 encryption?

RSA PKCS#1 v1.5 encryption is vulnerable to a type of attack that was discovered by Bleichenbacher back in '98 (see a summary [here](https://cryptosense.com/blog/why-pkcs1v1-5-encryption-should-be-put-out-of-our-misery)). The attacker leverages knowledge of when a ciphertext they sent produced an invalid padding when decrypted (i.e., a padding oracle) to break the system and recover the private RSA key of the peer.

> Well then it's good that we don't reveal information about when the padding is incorrect, right?

The padding oracle problem is relevant in the context of TLS - TLS 1.2 allows the client to unilaterally produce a Pre-Master Secret (PMS), encrypt it with the server's public key, and continue to use that PMS to secure the channel post-handshake. Thus, if PKCS#1 v1.5 is used, the server must take some precautions to never reveal that the PMS decryption resulted in an invalid pad. Instead, the server [should just choose a random value R as a placeholder and pretend that nothing bad happened](https://datatracker.ietf.org/doc/html/rfc5246#section-7.4.7.1). (This will result in an error later on in the handshake when the PMS is actually supposed to be used.) The problem with our bug is: the server cannot perform this mitigation if it doesn't know the error was caused by an invalid padding. Generic errors in our backends are most likely very infrequent, so the confusion between "padding error" and "generic error" can easily be used as a padding oracle.

## The fix

All we need to do is to make sure we always report back when a padding error has just happened. We should also test for this.

ionut-arm:
This fixes #619

When we get a response code indicating an invalid padding from either a
TPM or a PKCS11 backend, we now convert that to PsaErrorInvalidPadding
so that clients can perform any mitigations they wish.

### DIFF
--- a/e2e_tests/tests/per_provider/normal_tests/asym_encryption.rs
+++ b/e2e_tests/tests/per_provider/normal_tests/asym_encryption.rs
@@ -266,8 +266,31 @@ fn asym_encrypt_decrypt_rsa_pkcs_different_keys() {
     let ciphertext = client
         .asymmetric_encrypt_message_with_rsapkcs1v15(key_name_1, PLAINTEXT_MESSAGE.to_vec())
         .unwrap();
-    let res = client
+    let _res = client
         .asymmetric_decrypt_message_with_rsapkcs1v15(key_name_2, ciphertext)
+        .unwrap_err();
+}
+
+#[test]
+fn asym_decrypt_wrong_padding() {
+    let key_name = auto_test_keyname!();
+    let mut client = TestClient::new();
+
+    if !client.is_operation_supported(Opcode::PsaAsymmetricEncrypt)
+        || !client.is_operation_supported(Opcode::PsaAsymmetricDecrypt)
+    {
+        return;
+    }
+
+    client
+        .generate_rsa_encryption_keys_rsapkcs1v15crypt(key_name.clone())
+        .unwrap();
+    let mut ciphertext = client
+        .asymmetric_encrypt_message_with_rsapkcs1v15(key_name.clone(), PLAINTEXT_MESSAGE.to_vec())
+        .unwrap();
+    ciphertext[20] ^= 0x1;
+    let res = client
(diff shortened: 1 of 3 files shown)


mohamedasaker-arm (approved):
Good catch, Thanks!
LGTM!

paulhowardarm (approved):
Thanks, Ionut - looks good!

anta5010 (approved):

PR #624, opened 2022-07-13
Fix problem reported by Clippy (rust 1.62)

mohamedasaker-arm:

### DIFF
--- a/e2e_tests/tests/per_provider/normal_tests/asym_encryption.rs
+++ b/e2e_tests/tests/per_provider/normal_tests/asym_encryption.rs
@@ -213,7 +213,7 @@ fn asym_encrypt_wrong_algorithm() {
         return;
     }
 
-    let _key_id = client
+    client
         .generate_rsa_encryption_keys_rsaoaep_sha256(key_name.clone())
         .unwrap();
     let status = client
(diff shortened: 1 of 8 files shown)


anta5010:
Is this change compatible with old version of rust. Like 1.58.1 we use in Yocto?

ionut-arm:
> Is this change compatible with old version of rust. Like 1.58.1 we use in Yocto?

We do have CI checks for that sort of thing, so it should be caught if there's a regression.

mohamedasaker-arm:
Hi @anta5010 good point!
I've just used 1.58.1 and run the following commands as test
`cargo clippy --all-targets --features=all-providers -- -D clippy::all -D clippy::cargo`
`cargo clippy --all-targets --features=pkcs11-provider --manifest-path e2e_tests/Cargo.toml -- -D clippy::all -D clippy::cargo`
and I see no complaints on these changes. Do you have another check in mind?

ionut-arm:
Just to confirm: the CI checks for [1.57](https://github.com/parallaxsecond/parsec/blob/main/ci.sh#L223) as MSRV, so anything past that should work fine.

ionut-arm (approved):
LGTM

Issue #585, opened 2022-02-24
Re-enable CAL provider in `all-providers`


ionut-arm:
Following [this](https://github.com/parallaxsecond/parsec/discussions/538#discussioncomment-2242843) report of a build issue for armv7 targets within the `rust-cryptoauthlib` crate, and given that the crate is currently unmaintained, we're removing the CAL provider from the `all-providers` list and recommending it not to be included when packaging the Parsec service v1.0.0.

Once the project is placed under maintenance again, the provider can be added back and integrated in future builds of the service for distros. This should include not just the `Cargo.toml` change itself, but also enabling all the related e2e `all-providers` tests.

mohamedasaker-arm:
In #625 `run_key_mapping_tests` is disabled for CAL provider
Work of this ticket should include re-enabling `run_key_mapping_tests` for CAL in the `ci.sh`

PR #625, opened 2022-07-20
Testing/568 add key persistence tests for ts provider

mohamedasaker-arm:


e2e_tests/docker_image/parsec-service-test-all.Dockerfile
```diff
@@ -84,19 +84,20 @@ RUN cd nanopb-0.4.4-linux-x86 \
...
-	&& git reset --hard 389b50624f25dae860bbbf8b16f75b32f1589c8d
+	&& git reset --hard 35c6d643b5f0c0387702e22bf742dd4878ca5ddd
 # Install correct python dependencies
 RUN pip3 install -r trusted-services/requirements.txt
 RUN cd trusted-services/deployments/libts/linux-pc/ \
 	&& cmake . \
 	&& make \
-	&& cp libts.so* nanopb_install/lib/libprotobuf-nanopb.a mbedtls_install/lib/libmbedcrypto.a /usr/local/lib/
+	&& cp libts.so* nanopb_install/lib/libprotobuf-nanopb.a mbedcrypto_install/lib/libmbedcrypto.a /usr/local/lib/
 RUN rm -rf trusted-services
```
ionut-arm, line 101: I'd suggest moving this to the `generate-keys.sh` script as well, no point having it twice here and it makes it somewhat confusing as to what's actually running.

.github/workflows/ci.yml
```diff
@@ -27,10 +27,10 @@ jobs:
       # - name: Build the container
       #   run: pushd e2e_tests/docker_image && docker build -t parsec-service-test-all -f parsec-service-test-all.Dockerfile . && popd
```
ionut-arm, line 28: You have to uncomment this part everywhere as well.

e2e_tests/tests/per_provider/key_mappings.rs
```diff
@@ -27,6 +27,7 @@ fn use_and_check() -> Result<()> {
     client.set_default_auth(Some(String::from("parsec-tool")));
 
     let keys = client.list_keys()?;
+    assert!(!keys.is_empty());
```
ionut-arm, line 30: This might fail for providers that don't have persistent keys installed, like the CAL provider.

ionut-arm (commented):
Hey, thanks for the change, looks pretty good, just a few minor comments

e2e_tests/tests/per_provider/key_mappings.rs
mohamedasaker-arm: Yea, I didn't see CAL provider in action. Also the motive for this change is test will pass even for the providers that should have persistent keys but they don't exist.
So do you have something in mind to avoid this false positive?

mohamedasaker-arm: Hi @ionut-arm, sorry I'll be active on this PR again. So any suggestion to avoid false positives?
And by `I didn't see CAL provider in action`, it's vague, I admit. I meant I didn't see keys generated for the CAL provider.
I wonder if we don't run the key_mapping test for the CAL provider until the TODO about the CAL provider in the generate_keys script is done.

ionut-arm: Yeah, the CAL provider is in somewhat of a limbo, I think it's ok to not run the tests for now. Perhaps add a comment about it in [this](https://github.com/parallaxsecond/parsec/issues/585) issue.

ionut-arm (approved):
Thanks!!

e2e_tests/tests/per_provider/key_mappings.rs
mohamedasaker-arm: Done!

ionut-arm (approved):

PR #627, opened 2022-08-02
Kim file permissions

### LINKED ISSUE #598: Wrong permissions on KIM files
Both the on-disk and the SQLite KIMs store files on disk, be it individual entries for each key for the on-disk KIM, or one single file representing all the KIM data for the SQLite KIM. The permissions on these files have not been properly locked down, though - at the moment, all such files allow any user capable of reaching the files to read them. This is mitigated by having the files under `/var/lib/parsec`, which does not give any permissions to "other" users, thus preventing anyone other than the `parsec` user and group to access anything within.

Ideally these files (and their host directories) would have similar permissions, though: read and write (and execute, for directories) for the `parsec` user and group.

If possible, a unit test should be implemented to verify the permissions.

gowthamsk-arm:
Closes #598. 

The current key info manager files are not having any permissions set on them. This MR adds suitable permissions to directories and files of the KIM and also adds unit tests to verify if the permissions set are correct.

### DIFF
--- a/src/key_info_managers/on_disk_manager/mod.rs
+++ b/src/key_info_managers/on_disk_manager/mod.rs
@@ -33,14 +33,24 @@ use parsec_interface::requests::{AuthType, ProviderId};
 use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::ffi::OsStr;
+use std::fs::Permissions;
 use std::fs::{DirEntry, File};
 use std::io::{Error, ErrorKind, Read, Write};
 use std::ops::Deref;
+use std::os::unix::fs::PermissionsExt;
 use std::path::{Path, PathBuf};
 use std::{fmt, fs};
 
 /// Default path where the mapping files will be stored on disk
 pub const DEFAULT_MAPPINGS_PATH: &str = "/var/lib/parsec/mappings";
+
+///Permissions for all directories under database directory
+///Should only be visible to parsec user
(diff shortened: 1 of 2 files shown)


gowthamsk-arm:
The discussion about the KIM file permissions for parsec user and group is still ongoing. This MR can be merged once we agree on a particular approach.
https://github.com/parallaxsecond/parsec/issues/598

src/key_info_managers/on_disk_manager/mod.rs
```diff
@@ -33,15 +33,21 @@ use parsec_interface::requests::{AuthType, ProviderId};
...
+use std::os::unix::fs::PermissionsExt;
 use std::path::{Path, PathBuf};
 use std::{fmt, fs};
 
 /// Default path where the mapping files will be stored on disk
 pub const DEFAULT_MAPPINGS_PATH: &str = "/var/lib/parsec/mappings";
 
+///File permissions for all files/directories under database directory
+///Should only be visible to parsec user
+pub const FILE_PERMISSION: u32 = 0o770;
```
anta5010, line 49: why would we need +x on files? Shouldn't it be 770 for dirs and 660 for files?

gowthamsk-arm: You are right. I've updated the permissions now.

anta5010 (approved):

ionut-arm (approved):
Thanks! 👍🏻

PR #628, opened 2022-08-03
build and share docker image across jobs

mohamedasaker-arm:
The motive for this change is:
Whenever the docker image needs to be updated, test these changes on the CI
Then we enable building the docker image and run it for each job, as jobs in the workflow don't share the local docker registry.
This change is to build the docker and export it to be used across the workflow jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-      # Use the following step when updating the `parsec-service-test-all` image
     - name: Build the docker container
       run: pushd e2e_tests/docker_image && docker build -t parsec-service-test-all -f parsec-service-test-all.Dockerfile . && popd
     - name: Export the docker container
@@ -23,13 +22,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Download artifact
-        uses: actions/download-artifact@v3
+      - name: Load Docker
+        uses: ./.github/actions/load_docker
         with:
-          name: parsec-service-test-all
-          path: /tmp
-      - name: Load docker image
-        run:  docker load --input /tmp/parsec-service-test-all.tar
+          image-name: "parsec-service-test-all"
+          image-path: "/tmp"
       - name: Run the container to execute the test script
         # run:
         #   docker run -v $(pwd):/tmp/parsec -w /tmp/parsec
@@ -42,13 +39,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Download artifact
-        uses: actions/download-artifact@v3
+      - name: Load Docker
+        uses: ./.github/actions/load_docker
         with:
-          name: parsec-service-test-all
-          path: /tmp
-      - name: Load docker image
-        run:  docker load --input /tmp/parsec-service-test-all.tar
+          image-name: "parsec-service-test-all"
+          image-path: "/tmp"
       - name: Run the container to execute the test script
         # run:
         #   docker run -v $(pwd):/tmp/parsec -w /tmp/parsec ghcr.io/parallaxsecond/parsec-service-test-all /tmp/parsec/ci.sh cargo-check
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -2,145 +2,186 @@ name: Continuous Integration
...
+      uses: actions/upload-artifact@v3
+      with:
+        name: parsec-service-test-all
+        path: /tmp/parsec-service-test-all.tar
+
   all-providers:
     name: Various tests targeting a Parsec image with all providers included
     runs-on: ubuntu-latest
+    if: ${{ always() }}
+    needs: [build-and-export-test-all-docker]
```
ionut-arm, line 33: Does this mean we'll always build the image on the CI?

mohamedasaker-arm: it's the other way around
it will always run even if the `build-and-export-test-all-docker` is skipped

.github/workflows/ci.yml
```diff
@@ -2,145 +2,186 @@ name: Continuous Integration
...
+    - uses: actions/checkout@v3
+    - name: Build the docker container
+      run: pushd e2e_tests/docker_image && docker build -t parsec-service-test-all -f parsec-service-test-all.Dockerfile . && popd
+    - name: Export the docker container
+      run: docker save parsec-service-test-all > /tmp/parsec-service-test-all.tar
+    - name: Upload artifact
+      uses: actions/upload-artifact@v3
+      with:
+        name: parsec-service-test-all
+        path: /tmp/parsec-service-test-all.tar
```
ionut-arm, line 27: So what's the plan with this job? It seems to be useful when some change is made to the Docker image, but less than ideal if it runs all the time.

mohamedasaker-arm: Yes, that's why there is a condition on the execution. Unfortunately, I couldn't use the environment variable in this context.
So whenever we have docker changes, we switch only two lines
- the name of the image `parsec-service-test-all`
- this job condition to `true`
default will be `ghcr.io/parallaxsecond/parsec-service-test-all`
and condition will be `false`

ionut-arm: Cool, thanks, that's what I was hoping for!

ionut-arm:
Are we leaving this until after the 1.1.0 release?

mohamedasaker-arm:
@ionut-arm  We can merge if it's in good shape (I think it is). Anyhow It's not a part of the service itself.

ionut-arm (approved):
Looks good

.github/workflows/ci.yml
```diff
@@ -2,145 +2,186 @@ name: Continuous Integration
...
-          docker run -v $(pwd):/tmp/parsec -w /tmp/parsec
-          ghcr.io/parallaxsecond/parsec-service-test-all /tmp/parsec/ci.sh all
-      # When running the container built on the CI
-      # run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec -t parsec-service-test-all /tmp/parsec/ci.sh all
+        run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec -t ${{ env.TEST_ALL_DOCKER_IMAGE }} /tmp/parsec/ci.sh all
 
   build-all-providers:
     name: Cargo check all-providers (current Rust stable & old compiler)
     runs-on: ubuntu-latest
+    if: ${{ always() }}
```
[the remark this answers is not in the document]
mohamedasaker-arm: Yes, otherwise, this job won't be executed if the `build-and-export-test-all-docker` is skipped.
The only reason for the `need` list is to preserve order.

gowthamsk-arm:
LGTM

gowthamsk-arm (approved):

.github/workflows/ci.yml
```diff
@@ -2,145 +2,186 @@ name: Continuous Integration
...
+  # TEST_ALL_DOCKER_IMAGE: 'parsec-service-test-all'
+  TEST_ALL_DOCKER_IMAGE: 'ghcr.io/parallaxsecond/parsec-service-test-all'
+
 jobs:
+  build-and-export-test-all-docker:
+    runs-on: ubuntu-latest
+    # If TEST_ALL_DOCKER_IMAGE is 'parsec-service-test-all' or any local image,
+    # the following condition must evaluate true to execute this job
+    # Else it must evaluate false to NOT execute this job
+    # Unfortunately, env.TEST_ALL_DOCKER_IMAGE cannot be used here as the `env` context is not recognize at this level.
```
gowthamsk-arm, line 15: nit: recognized

Issue #603, opened 2022-05-04
Allow binary PIN values for PKCS11 providers


ionut-arm:
As mentioned [here](https://github.com/parallaxsecond/rust-cryptoki/pull/90) and [here](https://github.com/mheese/rust-pkcs11/pull/8), some PKCS11 tokens allow (or even require) PINs that do not conform to the spec, e.g., containing null bytes. The functionality has already been added (but not released, as of writing this) in the Cryptoki crate to allow such PINs, however this needs to be plumbed up into Parsec.

The authentication value config options for the TPM provider can stand as examples of how this could be implemented, namely by allowing the passed value to take the form `hex:...`, and to default to treating it as a string otherwise.

Any implementation should be accompanied by configuration tests.

ionut-arm:
> However, there is an unlikely corner case when the pin of the token is, for example:
"str:hex:normal_string".
> The pin is a normal string that has str:hex: as a prefix

I don't think there's a problem there, the value you'd put in would be `str:str:hex:...` - basically, everything after the first `str:` is considered as a string PIN. Am I getting this right?

mohamedasaker-arm:
Yea true, I mean it might get confusing a bit in this case. but indeed it's not a problem with the implementation

PR #631, opened 2022-08-24
Feature/603 allow binary pin values for pkcs11

mohamedasaker-arm:
This change is to accept different formats for pkcs11 user pin
like the TPM approach, there are three accepted formats

- Normal string: "this string treated as it is."
- Hex string: "hex:AABBCCDD" this passed as [0xAA 0xBB 0xCC 0xDD]
- Explicitly stated string: "str:NormalString" this passed as
"NormalString"

Fixes: #603

### DIFF
--- a/e2e_tests/tests/all_providers/config/mod.rs
+++ b/e2e_tests/tests/all_providers/config/mod.rs
@@ -465,7 +465,6 @@ fn init_pkcs11_token(lib: &str, so_pin: &str, pin: &str) -> String {
     pkcs11.initialize(CInitializeArgs::OsThreads).unwrap();
     let slot = pkcs11.get_slots_with_token().unwrap().pop().unwrap();
     pkcs11.init_token(slot, so_pin, "Test Token").unwrap();
-    println!("slots {:?}", slot);
     // set flags
     let mut flags = SessionFlags::new();
     let _ = flags.set_rw_session(true).set_serial_session(true);
--- a/e2e_tests/tests/all_providers/config/mod.rs
+++ b/e2e_tests/tests/all_providers/config/mod.rs
@@ -21,19 +21,44 @@ use std::time::Duration;
 const CONFIG_TOMLS_FOLDER: &str = "tests/all_providers/config/tomls";
 const SERVICE_CONFIG_PATH: &str = "provider_cfg/tmp_config.toml";
 
-fn set_config(filename: &str) {
-    info!("Changing service configuration file to {}", filename);
-    let config_path = PathBuf::from(SERVICE_CONFIG_PATH);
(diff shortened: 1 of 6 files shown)


e2e_tests/tests/all_providers/config/mod.rs
```diff
@@ -427,3 +452,82 @@ fn activate_cred_no_auth() {
...
+    use cryptoki::session::SessionFlags;
+    use cryptoki::session::UserType;
+    use std::path::Path;
+
+    let pkcs11 = Pkcs11::new(Path::new(lib)).unwrap();
+    // // initialize the library
+    pkcs11.initialize(CInitializeArgs::OsThreads).unwrap();
+    let slot = pkcs11.get_slots_with_token().unwrap().pop().unwrap();
+    pkcs11.init_token(slot, so_pin, "Test Token").unwrap();
+    println!("slots {:?}", slot);
```
ionut-arm, line 468: Is this still needed? 👀

src/providers/pkcs11/mod.rs
```diff
@@ -217,11 +220,17 @@ impl Provider {
...
 
         if self.user_pin.is_some() {
+            let mut pin = self.user_pin.as_ref().unwrap().expose_secret().clone();
+            if pin.starts_with(PIN_HEX_PREFIX) {
+                if let Ok(raw_pin) = hex::decode(pin.split_off(PIN_HEX_PREFIX.len())) {
+                    pin = String::from_utf8_lossy(&raw_pin.as_slice()).to_string();
+                }
+            } else if pin.starts_with(PIN_STRING_PREFIX) {
+                pin = pin.split_off(PIN_STRING_PREFIX.len()).into();
+            }
```
ionut-arm, line 230: Since you're taking the pin out of the secret wrapper, you'll have to zeroize (after last use) both `pin` and `raw_pin`.

ionut-arm (commented):
Thanks for putting the effort in the test framework!

src/providers/pkcs11/mod.rs
mohamedasaker-arm: Thanks for catching this ...
so for the `pin`, it should be `zeroize` after the login call, right?

e2e_tests/tests/all_providers/config/mod.rs
mohamedasaker-arm: nope 🥲
I'll remove it 😸!

src/providers/pkcs11/mod.rs
ionut-arm: Yes, once the login finishes, you can clear it up.

ionut-arm (approved):
👍🏻 LGTM

PR #633, opened 2022-09-03
Feature/119 implement configurable exclusion of deprecated primitives

### LINKED ISSUE #119: Implement configurable exclusion of deprecated primitives
As noted in the soon-to-be-merged protobuf contracts, some (combinations of) key attributes that Parsec supports are considered deprecated and should only be used by clients with a clear understanding of the security implications.

Thomas Fossati's suggestion is that the deprecated primitives are rejected by default, but can be enabled through the runtime configuration by admins.

There is still a question about where these checks should be enforced. My suggestions:
* Create and maintain a list of deprecated primitives in `parsec-interface-rs`, along with means of checking whether a request contains any such primitives
* Implement a feature in `parsec-interface-rs` that allows deprecated algorithms and which is not enabled by default; otherwise return an error when encountering one
* Enable the feature in `parsec`, but enforce the check through some other means, somewhere along the request pipeline (most likely in the `BackEndHandler`, before sending the request to the provider)

mohamedasaker-arm:
This change introduces configuration to reject the deprecated primitives.

Fixes #119

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -36,8 +36,9 @@
 # for buffers larger than this threshold will be rejected. Defaults to 1MB.
 #buffer_size_limit = 1048576
 
-# Decide whether deprecated algorithms and key types are allowed when generating keys or not
-# The default value is true
+# Decide whether deprecated algorithms and key types are allowed when generating keys or not.
+# Note: While importing a deprecated key, only a warning is needed.
+# The default behaviour is to allow the deprecated primitives. Hence, the default value is true.
 #allow_deprecated = true
 
 # (Required) Configuration for the service IPC listener component.
--- a/config.toml
+++ b/config.toml
@@ -38,8 +38,8 @@
 
 # Decide whether deprecated algorithms and key types are allowed when generating keys or not.
 # Note: While importing a deprecated key, only a warning is needed.
-# The default behaviour is to allow the deprecated primitives. Hence, the default value is true.
-#allow_deprecated = true
+# The default behaviour is to reject the deprecated primitives. Hence, the default value is false.
+#allow_deprecated = false
 
 # (Required) Configuration for the service IPC listener component.
 [listener]
--- a/config.toml
+++ b/config.toml
@@ -36,6 +36,10 @@
 # for buffers larger than this threshold will be rejected. Defaults to 1MB.
 #buffer_size_limit = 1048576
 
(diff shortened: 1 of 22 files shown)


config.toml
```diff
@@ -36,6 +36,11 @@
 # for buffers larger than this threshold will be rejected. Defaults to 1MB.
 #buffer_size_limit = 1048576
 
+# Decide whether deprecated algorithms and key types are allowed when generating keys or not.
+# Note: While importing a deprecated key, only a warning is needed.
```
ionut-arm, line 40:
```suggestion
# Note: While importing a deprecated key, only a warning log is generated.
```

src/providers/cryptoauthlib/key_management.rs
```diff
@@ -18,6 +19,13 @@ impl Provider {
         application_identity: &ApplicationIdentity,
         op: psa_generate_key::Operation,
     ) -> Result<psa_generate_key::Result> {
+        if let Err(ResponseStatus::DeprecatedPrimitive) = op.check_deprecated() {
+            warn!("The key requested to generate is deprecated");
+            if !crate::utils::GlobalConfig::allow_deprecated() {
+                return Err(ResponseStatus::DeprecatedPrimitive);
+            }
+        }
```
ionut-arm, line 27: Given that this bit of code is literally the same across all providers, would it be possible to do something a la [`format_error`](https://github.com/parallaxsecond/parsec/blob/main/src/lib.rs#L40), and only call that macro everywhere?

ionut-arm (commented):
Hey, thanks for the patch! Left some comments, let me know what you think.

src/providers/cryptoauthlib/key_management.rs
mohamedasaker-arm: Yea, good idea. Where do you think placing this macro would be the best? `src/providers/mod.rs`?
Also, do you think `warn_on_deprecated` or `return_on_deprecated` is okay for the macro names?
I am open to suggestions 😸

ionut-arm: Yeah, that sounds like a suitable location. In terms of name those sound good as well, or `handle_deprecated` or anything like that, doesn't matter that much since it's "internal"

ionut-arm (approved):
Cheers!

gowthamsk-arm (approved):

PR #651, opened 2022-11-22
Bump sd-notify to 0.3.0

stevecapperarm:
In order to package Parsec in Debian we need to package up the dependencies. One dependency is sd-notify which, when updated, has a breaking API change.

Specifically, listen_fds() now returns an iterator of fds rather than a count of fds. Also SD_LISTEN_FDS_START is no longer exposed publicly and is instead the first value in the iterator returned by listen_fds().

Signed-off-by: Steve Capper <steve.capper@arm.com>


mohamedasaker-arm:
Sorry for keeping the PR so long 😅.
Thanks @stevecapperarm, for the change. It looks pretty good!
Only a tiny thing...
Could you rebase the PR? (to avoid new Clippy warnings)
Also, Can you downgrade sd-notify 0.3.0? To keep support for rust 1.53
Please commit any changes in the lock file after downgrading the crate. (just run `cargo check` to see the updates in the lock file)

stevecapperarm:
Thanks @mohamedasaker-arm !, I've just force-pushed a change to 0.3.0 (and added the relevant Cargo.lock change too). Yeah 0.3.0 has the same breaking API change so this commit is still helpful for us packaging this in Debian.

mohamedasaker-arm:
The PR is perfect now!
Nit: Could edit the name of the change (0.3.0 instead of 0.4.1)?

stevecapperarm:
Ahhh, just found the "edit" button :-), sorted thanks @mohamedasaker-arm

gowthamsk-arm:
Thanks for the PR. LGTM :)

gowthamsk-arm (approved):

mohamedasaker-arm (approved):

ionut-arm (approved):
Thanks, looks good to me! 👍🏻

PR #657, opened 2023-01-16
Remove dependency on crate "version"

ema:
The crate called "version" takes the environment variable CARGO_PKG_VERSION and returns the values of major, minor, and patch as integers. Do that ourselves instead of using an external dependency just for this. The main reason for this change is making the Debian packaging efforts easier: in Debian, each dependency must be packaged individually, and each individual package must provide a non-trivial amount of functionality. Splitting a string and converting the results to integer is clearly simple enough to be done without an additional dependency.

### DIFF
--- a/src/providers/core/mod.rs
+++ b/src/providers/core/mod.rs
@@ -19,10 +19,9 @@ use parsec_interface::operations::{
 use parsec_interface::requests::{Opcode, ProviderId, ResponseStatus, Result};
 use std::collections::{HashMap, HashSet};
 use std::io::{Error, ErrorKind};
-use std::str::FromStr;
+use std::num::ParseIntError;
 use std::sync::Arc;
 use uuid::Uuid;
-use version::{version, Version};
 
 const SUPPORTED_OPCODES: [Opcode; 5] = [
     Opcode::ListProviders,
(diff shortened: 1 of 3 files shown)


src/providers/core/mod.rs
```diff
@@ -326,4 +340,17 @@ mod tests {
             provider.wire_protocol_version_min
         );
     }
+
+    #[test]
+    fn test_build() {
+        let provider_builder = ProviderBuilder::new().with_wire_protocol_version(42, 12);
+        let provider = provider_builder.build().unwrap();
+        for provider_info in provider.provider_info {
+            assert!(
```
adamjohnstonarm, line 349: I can't find anything to say that version 0.0.0 is illegal (even if it is unlikely) so this might not be a valid check. I'm not sure we need to do anything beyond checking the version can be parsed into 3 integers.

ema: Would it be sufficient to call `provider_builder.build().unwrap()` or did you have any specific assertions in mind?

src/providers/core/mod.rs
```diff
@@ -263,13 +262,28 @@ impl ProviderBuilder {
...
         }
 
-        let crate_version: Version = Version::from_str(version!()).map_err(|e| {
-            format_error!("Error parsing the crate version", e);
+        let crate_version: std::result::Result<Vec<u32>, ParseIntError> = env!("CARGO_PKG_VERSION")
+            .split('.')
+            .map(|v| v.parse())
+            .collect();
+
+        let crate_version = crate_version.map_err(|_| {
```
gowthamsk-arm, line 270: Can you also include `format_error` and print out the error message?

src/providers/core/mod.rs
gowthamsk-arm: Just check the format should be sufficient enough I guess.

xueliang-zhong-arm (approved):

src/providers/core/mod.rs
ema: Done!

gowthamsk-arm (approved):

src/providers/core/mod.rs
```diff
@@ -263,13 +262,29 @@ impl ProviderBuilder {
             provider_info_vec.push(provider_info);
         }
 
-        let crate_version: Version = Version::from_str(version!()).map_err(|e| {
-            format_error!("Error parsing the crate version", e);
+        let crate_version: std::result::Result<Vec<u32>, ParseIntError> = env!("CARGO_PKG_VERSION")
+            .split('.')
+            .map(|v| v.parse())
+            .collect();
```
ionut-arm, line 268: This is mostly fine, but in the case of release candidates (for example [this](https://github.com/parallaxsecond/parsec/releases/tag/1.1.0-rc2)) it (probably?) won't work, IIUC. I'd prefer avoiding regular expressions, maybe discard everything after the first `-` if it exists? That would also take care of any alpha/beta versioning as well.

ionut-arm: Ah, nevermind, it's parsing the version in `Cargo.toml`, and that one is always in the correct format (I'd hope), disregard this.

ionut-arm (approved):

adamjohnstonarm (approved):
LGTM

PR #660, opened 2023-02-02
Allow uninlined format arguments clippy error

gowthamsk-arm:
The new rust 1.67 release adds a new Clippy error and this causes CI failure.
https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args

This MR allows the Clippy warning.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,4 +1,5 @@
 #![allow(clippy::multiple_crate_versions, unused)]
+#![allow(clippy::uninlined_format_args)]
 use std::env;
 use std::fs::read_dir;
 use std::io::{Error, ErrorKind, Result};
--- a/e2e_tests/tests/mod.rs
+++ b/e2e_tests/tests/mod.rs
@@ -28,6 +28,7 @@
 )]
 // This one is hard to avoid.
 #![allow(clippy::multiple_crate_versions)]
+#![allow(clippy::uninlined_format_args)]
 
 #[cfg(feature = "all-providers")]
 mod all_providers;
--- a/src/authenticators/mod.rs
+++ b/src/authenticators/mod.rs
@@ -9,6 +9,8 @@
 //! is the `RequestAuth` field of a request, which is parsed by the authenticator specified in the header.
 //! The authentication functionality is abstracted through an `Authenticate` trait.
 
+#![allow(clippy::uninlined_format_args)]
+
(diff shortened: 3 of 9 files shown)


ionut-arm (approved):
LGTM, thanks for taking care of this. Something to note is that perhaps we'll be able to get rid of these if we can bump the MSRV enough for the next release

gowthamsk-arm:
Sorry I didn't get it @ionut-arm 
This issue comes up for 1.67. How does MSRV update fix it? :)

ionut-arm:
I think support for inline args was introduced earlier, but it only became a clippy warning to _not_ use it recently, in 1.67. So if we're lucky, our new MSRV will allow inline arguments and we won't have to `#[allow]` uninlined ones. :)

mohamedasaker-arm (approved):
LGTM, Thanks!

mohamedasaker-arm:
@ionut-arm 
Reading this " So if we're lucky, our new MSRV will allow inline arguments and we won't have to #[allow] uninlined ones." from your previous comment makes me think about how this will solve the issue when we use 1.67 or any latest version

adamjohnstonarm:
Indeed. Would it make more sense to change the MSRV now and migrate to named format arguments? Which min versions do we need to support for the distros? How many errors are thrown in 1.67?

> from your previous comment makes me think about how this will solve the issue when we use 1.67 or any latest version

gowthamsk-arm:
Most of the `format!` statements we have, use uninlined format so there are lots of errors. Rust version 1.53 is the current MSRV as it is used in OpenSUSE 15.3 (which has reached EOL).

gowthamsk-arm:
This latest CI run shows that it's not needed.
https://github.com/parallaxsecond/parsec/actions/runs/4065029371
I will close the MR if nobody has any concerns.

gowthamsk-arm:
https://blog.rust-lang.org/2023/02/09/Rust-1.67.1.html

nyurik:
`uninlined_format_args` respects MSRV specified in cargo.toml - if it is before the 1.58(?), it will not even show. In the future, it will be re-enabled, and you can easily apply it with `cargo clippy --fix -- -A clippy::all -W clippy::uninlined_format_args`

adamjohnstonarm (approved):
LGTM

PR #661, opened 2023-02-14
Update MSRV to 1.58

gowthamsk-arm:
Rust version 1.53 was supported in OpenSUSE Leap 15.3 which has reached EOL. The new MSRV 1.58 is decided after looking into the distributions that are of interest.


ionut-arm (approved):

mohamedasaker-arm (approved):
Thanks for the change!

PR #683, opened 2023-07-14
Replace unmaintained 'users' crate with 'nix' crate

tgonzalezorlandoarm:
This PR opens a discussion thread to fix the issue of the 'users' crate currently being flagged as 'unmaintained' by our CI (which is currently failing):
See:

https://rustsec.org/advisories/RUSTSEC-2023-0040.html

Currently we only use the users crate to the method [get_current_id()](https://docs.rs/users/0.11.0/users/fn.get_current_uid.html), which makes an unsafe call to libc's getuid(). parsec-client-rust and parsec-tool repos are also affected by this, all of them using the same function call from the users crate and no other call that I can tell.
There are several options to fix this issue.

 - Mark this as okay in the CI and leave it "as is", as the only function that we are using is [get_current_id()](https://docs.rs/users/0.11.0/users/fn.get_current_uid.html) which makes a call to getuid() and nothing else, so the fact that the crate is unmaintained would not affect us. The [users crate dependencies](https://crates.io/crates/users/0.11.0/dependencies) are only libc (the log one is optional).
 - Use libc unsafe calls directly instead.
 - Change the users crate to a different crate.

    - The suggested solution by [RustSec](https://rustsec.org/advisories/RUSTSEC-2023-0040.html) is to use [sysinfo](https://crates.io/crates/sysinfo), but this one has a lot of [dependencies](https://crates.io/crates/sysinfo/0.29.4/dependencies) that have their own dependencies and from which we don't need most of them. This may lead to potential problems in the future as [RustSec](https://rustsec.org/) may flag these as a problem and we may have more versioning problems as well.
   - I found the [nix](https://crates.io/crates/nix/0.26.2) crate that has [dependencies](https://crates.io/crates/nix/0.26.2/dependencies) that are only macro related and have no nested dependencies (so there is a very limited potential for dependency problems in the future). If we take into account the ones that are already being used by parsec as dependencies from other crates, [static_assertions](https://crates.io/crates/static_assertions/1.1.0) would be the only dependency that is added by nix. Most of the dependencies and used code is controlled by enabling only the needed features. This solution would remove the CI issue and has very limited potential for future problems that may arise by dependencies.

This is a patch to replace the users crate with the nix crate that currently has no issues of that kind and implements the same needed functionality (calling libc::getuid() ). This shows that this may be implemented if needed, and this is applicable to the rest of the crates as well.

I will submit a different MR with another one of the solutions implemented as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -833,6 +833,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e5ce46fe64a9d73be07dcbe690a38ce1b293be448fd8ce1e6c1b8062c9f72c6a"
 
 [[package]]
+name = "nix"
+version = "0.26.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bfdda3d196821d6af13126e40375cdf7da646a96114af134d5f417a9a1dc8e1a"
+dependencies = [
+ "bitflags",
+ "cfg-if",
+ "libc",
+ "static_assertions",
+]
+
+[[package]]
 name = "nom"
 version = "5.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1041,6 +1053,7 @@ dependencies = [
  "hex",
  "libc",
  "log",
+ "nix",
  "num-traits",
  "parsec-interface",
  "picky-asn1",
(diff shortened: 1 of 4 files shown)


ionut-arm:
I would prefer the other fix as well, happy to close this if you agree

tgonzalezorlandoarm:
Agreed. This was discussed with @gowthamsk-arm as well and he suggested having the other fix instead of this one.

PR #684, opened 2023-07-14
Remove unmaintained 'users' crate

tgonzalezorlandoarm:
The 'users' crate is currrently being flagged as unmaintained in https://rustsec.org/advisories/RUSTSEC-2023-0040.html . It is currently only being used to get the uid of the user of a running process.

 * Replace all users::get_current_uid() calls with libc::getuid() calls.
 * Remove the users crate from Cargo.toml and Cargo.lock

Please refer to #683 for more information on this issue and the corresponding fix options

### DIFF
--- a/src/authenticators/unix_peer_credentials_authenticator/mod.rs
+++ b/src/authenticators/unix_peer_credentials_authenticator/mod.rs
@@ -109,11 +109,11 @@ mod test {
     use super::UnixPeerCredentialsAuthenticator;
     use crate::front::domain_socket::peer_credentials;
     use crate::front::listener::ConnectionMetadata;
+    use libc::{getuid, uid_t};
     use parsec_interface::requests::request::RequestAuth;
     use parsec_interface::requests::ResponseStatus;
     use rand::Rng;
     use std::os::unix::net::UnixStream;
-    use users::get_current_uid;
 
     #[test]
     fn successful_authentication() {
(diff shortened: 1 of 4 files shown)


src/authenticators/unix_peer_credentials_authenticator/mod.rs
```diff
@@ -143,7 +143,8 @@ mod test {
             .authenticate(&req_auth, conn_metadata)
             .expect("Failed to authenticate");
 
-        assert_eq!(application.identity.name, get_current_uid().to_string());
+        let current_uid: uid_t = unsafe { getuid() };
```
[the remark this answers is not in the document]
tgonzalezorlandoarm: This is because we are calling the libc system call getuid. As you can see in https://man7.org/linux/man-pages/man2/geteuid.2.html , this function is always successful and does not modify errno. 

Under the hood, the replaced crate was doing this when calling get_current_uid, as you can see here:
https://docs.rs/users/0.11.0/src/users/base.rs.html#552-557

ionut-arm (approved):

src/authenticators/unix_peer_credentials_authenticator/mod.rs
```diff
@@ -247,7 +249,8 @@ mod test {
...
```
gowthamsk-arm, line 252: Can we reuse the `current_uid` variable used before?

tgonzalezorlandoarm: Done

PR #691, opened 2023-08-21
Bump ASN1 crates dependencies

anta5010:
picky-asn1 0.7.2 contains a fix for occasional incorrect ASN1 Integer builds: https://github.com/Devolutions/picky-rs/pull/209

### DIFF
--- a/fuzz/Cargo.lock
+++ b/fuzz/Cargo.lock
@@ -32,15 +32,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
- "winapi",
-]
-
-[[package]]
 name = "anyhow"
 version = "1.0.64"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -79,7 +70,7 @@ checksum = "db8b7511298d5b7784b40b092d9e9dcd3a627a5707e4b5e507931ab0d44eeebf"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 1.0.99",
  "synstructure",
 ]
 
@@ -91,7 +82,7 @@ checksum = "2777730b2039ac0f95f093556e61b6d26cebed5393ca6f152717777cec3a42ed"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 1.0.99",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ionut-arm (approved):

tgonzalezorlandoarm (approved):

fuzz/Cargo.lock
```diff
@@ -589,18 +624,19 @@ dependencies = [
...
 [[package]]
 name = "hashbrown"
-version = "0.12.3"
+version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a9ee70c43aaf417c914396645a0fa852624801b24ebb7ae78fe8272889ac888"
+checksum = "2c6201b9ff9fd90a5a3bac2e56a830d0caa509576f0e503818ee82c181b3437a"
 dependencies = [
  "ahash",
+ "allocator-api2",
```
tgonzalezorlandoarm, line 632: This crate is optional according to https://crates.io/crates/hashbrown/0.14.0/dependencies , is it possible to avoid adding it?

fuzz/Cargo.lock
```diff
@@ -1317,27 +1364,39 @@ dependencies = [
...
 
 [[package]]
 name = "regex"
-version = "1.6.0"
+version = "1.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4c4eb3267174b8c6c2f654116623910a0fef09c4753f8dd83db29c48a0df988b"
+checksum = "81bc1d4caf89fac26a70747fe603c130093b53c773888797a6329091246d651a"
+dependencies = [
+ "aho-corasick",
```
tgonzalezorlandoarm, line 1389: Do we need this one? It's listed as optional in here https://crates.io/crates/regex/1.9.3/dependencies

fuzz/Cargo.lock
```diff
@@ -1317,27 +1364,39 @@ dependencies = [
...
+ "memchr",
```
tgonzalezorlandoarm, line 1390: Same here.

tgonzalezorlandoarm (changes requested):

tgonzalezorlandoarm:
In general I would like to avoid adding as many dependencies as possible because this would mean a lot more packages to manage and would save us some issues down the line

fuzz/Cargo.lock
anta5010: I didn't update it manually. That what `cargo update` produced. I don't see any command line options to exclude optional dependencies. Maybe we have to define required features for all dependency crates in parsec-fuzz cargo.toml or filter Cargo.lock manually.

anta5010: I can run `cargo update --package picky-asn1 --package picky-asn1-der --package picky-asn1-x509` instead of `cargo update` for this PR and leave running `cargo update` for the release process if you prefer

tgonzalezorlandoarm: I think it's better if we only update the necessary here and we update the rest for the release, as it would be better to avoid the issue of adding unneeded dependencies.

tgonzalezorlandoarm (approved):
LGTM

PR #694, opened 2023-08-31
Fix coverage builds for different providers

tgonzalezorlandoarm:
Currently, when iterating among the different providers, the variable UNIT_TEST_FEATURES is being appended a provider in each iteration. This makes the any iteration of the coverage build to build for not only the provider of the current iteration but also for the previous providers.

 * Only build for the provider of the current iteration in the coverage build.


ionut-arm (approved):
Thanks for catching this!

PR #695, opened 2023-08-31
Make wrong_permitted_algorithm test use a non-deprecated Hash

tgonzalezorlandoarm:
Currently, the wrong_permitted_algorithm test checks if, creating a key that uses sha1, that key is rejected when calling sign_with_rsa_sha256. However, as sha1 is deprecated, there is currently an error when creating the key for the test.

 * Change the key generation to use sha512 instead.


ionut-arm (approved):
Thanks!

PR #697, opened 2023-09-01
Update cryptoki and cryptoki-sys crates

tgonzalezorlandoarm:
* Update the cryptoki and cryptoki-sys crates.
 * Remove the psa-crypto-conversions feature from the cryptoki dependency in the Cargo.toml file. The psa-crypto crate is already being brought up in the Cargo.toml file and this would lead to a crate conflict when upgrading.


tgonzalezorlandoarm:
The last commit of the PR is breaking the tests (the tests pass for the previous commit) . What I'm thinking is causing the issue is the introduction of PkcsOaepSource::new and maybe how I'm using it?
This is hypothesis is based on the particular test that is failing.
@ionut-arm do you have any suggestions by any chance?

gowthamsk-arm (approved):

PR #701, opened 2023-09-21
Fix cargo-tarpaulin version to 0.26.1

tgonzalezorlandoarm:
* Create an environment variable to define the MSRV so that only
       one change is needed to update it.
     * Fix the version of cargo-tarpaulin to 0.26.1

Signed-off-by: Tomás González <tomasagustin.gonzalezorlando@arm.com>


ionut-arm (approved):
👍🏻

PR #727, opened 2023-10-25
ci/coverage: Fix cargo-tarpaulin to its locked version

tgonzalezorlandoarm:
cargo-tarpaulin's dependency toml_datetime has made a patch update to v0.6.5 which requires rust version 1.67 or newer. As our MSRV is currently 1.66.0:
 * use the '--locked' option to build cargo-tarpaulin instead.


gowthamsk-arm (approved):
LGTM

PR #728, opened 2023-10-25
e2e_tests/mangled_ping: Fix socket path

tgonzalezorlandoarm:
Use the environment variable PARSEC_SERVICE_ENDPOINT to define the socket path for the test.
This came up during HW testing (Trusted Service Provider). PARSEC_SERVICE_ENDPOINT may be defined to something different than /tmp/parsec.sock so we should be taking into account that case


gowthamsk-arm:
nit: In the commit description can you also add what prompted this change?

gowthamsk-arm (approved):

PR #732, opened 2023-11-06
Track and test dependencies' 'next' branch

tgonzalezorlandoarm:
parsec depends on several repositories under the parallaxsecond organization.

Currently, when testing the main branch of parsec, fixed versions of the latest released crates of those repositories are being used.
A problem may arise when there is a change introduced in the main/'next' branch of those repositories that would break parsec when updating/incorporating that change. This is currently not being tested in our CI.
As parsec release is dependent on having published the dependencies' crates, there are issues that will not be caught until the publishing of said crates.

* Add nightly jobs that use ci.sh --next-branch-tracking to test the
  next release branches of parallaxsecond repositories and spot
  issues before release. These jobs make use of previously defined
  re-usable actions and utils/release_tracking.py

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,15 +36,11 @@ jobs:
     if: ${{ always() }}
     needs: [build-and-export-test-all-docker]
     steps:
-      - uses: actions/checkout@v2
-      - name: Load Docker
-        uses: ./.github/actions/load_docker
-        if: ${{ env.TEST_ALL_DOCKER_IMAGE == 'parsec-service-test-all' }}
-        with:
-          image-name: "${{ env.TEST_ALL_DOCKER_IMAGE }}"
-          image-path: "/tmp"
+      - uses: actions/checkout@v3
       - name: Run the container to execute the test script
-        run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec -t ${{ env.TEST_ALL_DOCKER_IMAGE }} /tmp/parsec/ci.sh all
+        uses: ./.github/actions/ci_script
+        with:
+          ci-flags: "all"
 
   build-all-providers:
     name: Cargo check all-providers (current Rust stable & old compiler)
@@ -52,15 +48,11 @@ jobs:
     if: ${{ always() }}
     needs: [build-and-export-test-all-docker]
     steps:
-      - uses: actions/checkout@v2
-      - name: Load Docker
-        uses: ./.github/actions/load_docker
-        if: ${{ env.TEST_ALL_DOCKER_IMAGE == 'parsec-service-test-all' }}
-        with:
-          image-name: "${{ env.TEST_ALL_DOCKER_IMAGE }}"
-          image-path: "/tmp"
+      - uses: actions/checkout@v3
       - name: Run the container to execute the test script
-        run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec -t ${{ env.TEST_ALL_DOCKER_IMAGE }} /tmp/parsec/ci.sh cargo-check
+        uses: ./.github/actions/ci_script
+        with:
+          ci-flags: "cargo-check"
 
   mbed-crypto-provider:
     name: Integration tests using Mbed Crypto provider
@@ -68,15 +60,11 @@ jobs:
     if: ${{ always() }}
     needs: [build-and-export-test-all-docker]
     steps:
-      - uses: actions/checkout@v2
-      - name: Load Docker
-        uses: ./.github/actions/load_docker
-        if: ${{ env.TEST_ALL_DOCKER_IMAGE == 'parsec-service-test-all' }}
-        with:
-          image-name: "${{ env.TEST_ALL_DOCKER_IMAGE }}"
-          image-path: "/tmp"
+      - uses: actions/checkout@v3
       - name: Run the container to execute the test script
-        run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec -t ${{ env.TEST_ALL_DOCKER_IMAGE }} /tmp/parsec/ci.sh mbed-crypto
+        uses: ./.github/actions/ci_script
+        with:
+          ci-flags: "mbed-crypto"
 
   pkcs11-provider:
     name: Integration tests using PKCS 11 provider
@@ -84,15 +72,11 @@ jobs:
     if: ${{ always() }}
     needs: [build-and-export-test-all-docker]
     steps:
-      - uses: actions/checkout@v2
-      - name: Load Docker
-        uses: ./.github/actions/load_docker
-        if: ${{ env.TEST_ALL_DOCKER_IMAGE == 'parsec-service-test-all' }}
-        with:
-          image-name: "${{ env.TEST_ALL_DOCKER_IMAGE }}"
-          image-path: "/tmp"
+      - uses: actions/checkout@v3
       - name: Run the container to execute the test script
-        run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec -t ${{ env.TEST_ALL_DOCKER_IMAGE }} /tmp/parsec/ci.sh pkcs11 --no-stress-test
+        uses: ./.github/actions/ci_script
+        with:
+          ci-flags: "pkcs11 --no-stress-test"
 
   tpm-provider:
     name: Integration tests using TPM provider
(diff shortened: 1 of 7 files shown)


tgonzalezorlandoarm:
Please check the added tests in my fork:

tgonzalezorlandoarm:
Will generate a follow up PR creating a `utils/cargo-commands.py` after this , as well as simplifying the re-usable actions

utils/release_tracking.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def git_toml_deps(toml_path, updatable_deps, deps_repos):
+    lines = None
+    with open(toml_path, 'r') as f:
+        lines = f.readlines()
+
+    for i in range(len(lines)):
+        line = lines[i]
+        for dep in updatable_deps.keys():
```
minosgalanakis, line 31: We could instead do something like that:

    for dep, dep_name in updatable_deps.items():
    ....

utils/release_tracking.py
```diff
@@ -0,0 +1,97 @@
...
+                # All occurences
+                line = line.replace("'", '"')
+                if "{" not in line:
+                    # First occurence
+                    line = line.replace('"', '{ version = "', 1)
+                    line = line.rstrip() + ' }\n'
+
+                if 'path' not in line:
+                    line = re.sub(r'version = "[0-9\.]+"', f'path = "{deps_repos[dep]}"', line)
+                lines[i] = line
```
minosgalanakis, line 42: Modifying an list while we are iterating through it is pretty risky. 

    output_lines = []

    for line in lines():
        output_lines.append(lines)

utils/release_tracking.py
```diff
@@ -0,0 +1,97 @@
+import argparse
+import re
+import os
+import subprocess
+import sys
+
+
+def run_cargo_build(path):
+    print(f"cargo build, path: {path}")
+    command = f'cargo build'
```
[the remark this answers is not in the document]
tgonzalezorlandoarm: That's why I put the `command.split()`
I first make a string literal for the command so that it's easier to read for us developers, and then I split it into a list for `check_output`. It's easier to read this way :)

utils/release_tracking.py
```diff
@@ -0,0 +1,97 @@
...
+    subprocess.check_output(command.split(), cwd=path)
```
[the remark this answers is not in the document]
tgonzalezorlandoarm: `check_output` throws an exception if the return code is incorrect, which helps us catch errors earlier. It's also what projects like MbedTLS use in their testing scripts.
We can indeed return the output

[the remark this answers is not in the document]
tgonzalezorlandoarm: I will do this last bit

utils/release_tracking.py
```diff
@@ -0,0 +1,97 @@
...
+        command = f"git submodule update --init"
+        subprocess.check_output(command.split(), cwd=future_repo_dir)
+
+
+def git_toml_deps(toml_path, updatable_deps, deps_repos):
+    lines = None
+    with open(toml_path, 'r') as f:
+        lines = f.readlines()
+
+    for i in range(len(lines)):
```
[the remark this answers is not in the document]
tgonzalezorlandoarm: I am editting the lines on the fly `line[i]=new_line` so this is better for that

utils/release_tracking.py
```diff
@@ -0,0 +1,97 @@
...
+                                                 'main branches of parallaxsecond dependencies in '
+                                                 'preparation for their publishing and release')
+    parser.add_argument('--clone_dir',
+                        required=True,
+                        help='Existing directory into which repositories should be cloned')
+    parser.add_argument('paths', nargs='+', help='Paths to Cargo.toml files to be modified')
+    args = parser.parse_args()
+
+    # The order is important!
+    parallaxsecond_deps = {
```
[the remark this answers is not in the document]
tgonzalezorlandoarm: As of Python 3.7 dictionaries are ordered

utils/release_tracking.py
tgonzalezorlandoarm: I only needed the keys of the dictionary, that's why I didn't use `items()`

utils/release_tracking.py
```diff
@@ -0,0 +1,97 @@
...
+    command = f'cargo build'
+    subprocess.check_output(command.split(), cwd=path)
+
+
+def clone_repo(clone_dir, repo_name, branch):
+    git_repo = f"https://github.com/parallaxsecond/{repo_name}.git"
+    future_repo_dir = os.path.join(clone_dir, repo_name)
+    if not os.path.isdir(future_repo_dir):
+        command = f"git clone {git_repo} -b {branch} {future_repo_dir}"
+        subprocess.check_output(command.split())
```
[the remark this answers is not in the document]
tgonzalezorlandoarm: `check_output` requires a list, not a string.
As explained in a different comment, I first make a string literal for the command so that it's easier to read for us developers, and then I split it into a list for check_output. It's easier to read this way :)
I don't see how this is a security risk if all of the strings are hardcoded?

utils/release_tracking.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def run_cargo_build(path):
+    print(f"cargo build, path: {path}")
+    command = f'cargo build'
+    subprocess.check_output(command.split(), cwd=path)
+
+
+def clone_repo(clone_dir, repo_name, branch):
+    git_repo = f"https://github.com/parallaxsecond/{repo_name}.git"
```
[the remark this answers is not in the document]
tgonzalezorlandoarm: `repo_branches` `parallaxsecond_deps` or `repo_paths` are defined in main, so I think better not to hardcode them as a file global but just defined in main and in case we want to scale things then put them as a global thing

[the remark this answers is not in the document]
tgonzalezorlandoarm: I will use your suggestion and pass the git_repo as an argument

utils/release_tracking.py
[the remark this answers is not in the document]
tgonzalezorlandoarm: Done

utils/release_tracking.py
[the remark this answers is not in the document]
tgonzalezorlandoarm: Done

utils/release_tracking.py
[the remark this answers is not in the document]
tgonzalezorlandoarm: After our discussion, I changed to shell=True

utils/release_tracking.py
tgonzalezorlandoarm: Done

tgonzalezorlandoarm:
https://github.com/tgonzalezorlandoarm/parsec/actions/runs/6786709261 new test on my fork after these changes

tgonzalezorlandoarm:
https://github.com/tgonzalezorlandoarm/parsec/actions/runs/6786705099

utils/release_tracking.py
```diff
@@ -0,0 +1,99 @@
...
+    repo_paths = { repo_name: f'{args.clone_dir}/{repo_folder}/{repo_name}' \
+                   for repo_name, repo_folder in parallaxsecond_deps.items() }
+    repo_paths['parsec-interface'] = f'{args.clone_dir}/parsec-interface-rs'
+    repo_paths['parsec-client'] = f'{args.clone_dir}/parsec-client-rust'
+
+    repo_branches = { repo_name: 'main' for repo_name in parallaxsecond_deps.keys() }
+    repo_branches['tss-esapi-sys'] = '7.x.y'
+    repo_branches['tss-esapi'] = '7.x.y'
+
+    for repo_name, repo_folder in parallaxsecond_deps.items():
```
gowthamsk-arm, line 82: We could instead use [patch](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section) in the toml file. 

```
[patch.crates-io]
psa-crypto = { git = 'https://github.com/parallaxsecond/rust-psa-crypto.git', branch = "main" }
```

.github/workflows/ci.yml
```diff
@@ -33,14 +33,10 @@ jobs:
     needs: [build-and-export-test-all-docker]
     steps:
       - uses: actions/checkout@v2
-      - name: Load Docker
```
gowthamsk-arm, line 36: Can you verify if the PR changes work as expected by triggering a pipeline with `build-and-export-test-all-docker` generating the docker image?

utils/release_tracking.py
minosgalanakis: That is a very nice feature that I didn't know about.  It would remove the need for the `git_toml_deps` method?

tgonzalezorlandoarm:
https://github.com/tgonzalezorlandoarm/parsec/actions/runs/6800846830 New CI run

utils/release_tracking.py
tgonzalezorlandoarm: For future reference:

Adding the `cargo update` step here was crucial to make it work as otherwise we would be using the Cargo.lock file version instead of the patch one

.github/workflows/ci.yml
tgonzalezorlandoarm: To verify this, I ran
https://github.com/tgonzalezorlandoarm/parsec/actions/runs/6800961661

That is based on this commit:
https://github.com/tgonzalezorlandoarm/parsec/commit/9639bc9845962578240223b998834b35900377e2

gowthamsk-arm: Thanks

.github/actions/all_providers/action.yml
```diff
@@ -0,0 +1,20 @@
...
+description: "Various tests targeting a Parsec image with all providers included"
+inputs:
+    ci-flags-all-providers:
+      required: true
+      description: "Flags with which to run the all providers ci tests"
+# ...name, description and inputs as above
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/checkout@v2
```
gowthamsk-arm, line 11: Just checking, is it required here again?

gowthamsk-arm: The ci.yml still has the checkout action.

tgonzalezorlandoarm:
https://github.com/tgonzalezorlandoarm/parsec/actions/runs/6853736203

minosgalanakis (approved):
LGTM

gowthamsk-arm (approved):
LGTM. Thanks.

PR #737, opened 2023-11-21
cargo-check: Run with both stable and MSRV Compilers

tgonzalezorlandoarm:
- Introduce a RUST_TOOLCHAIN_VERSION environment variable that gets used in ci.sh to decide on the compiler to use.
- Use this variable when running variable on nightly (coverage) or build checks on each pr.
- cargo check will now cover all feature cases with both MSRV and stable, which it was not doing. Fix the cargo check build with MSRV for certain options that were not being covered by fixing versions on the Cargo.lock file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,8 +42,21 @@ jobs:
         with:
           ci-flags: "all"
 
-  build-all-providers:
-    name: Cargo check all-providers (current Rust stable & old compiler)
+  build-all-providers-msrv:
+    name: MSRV - Cargo check all-providers
+    runs-on: ubuntu-latest
+    if: ${{ always() }}
+    needs: [build-and-export-test-all-docker]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run the container to execute the test script
+        uses: ./.github/actions/ci_script
+        with:
+          ci-flags: "cargo-check"
+          rs-version: "1.66.0"
+
+  build-all-providers-stable:
+    name: Cargo check all-providers (current Rust stable)
     runs-on: ubuntu-latest
     if: ${{ always() }}
     needs: [build-and-export-test-all-docker]
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -52,7 +52,7 @@ jobs:
         with:
           ref: "${{ github.event.inputs.rev }}"
       - name: Run the container to execute the coverage script
-        run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec --security-opt seccomp=unconfined ghcr.io/parallaxsecond/parsec-service-test-all /tmp/parsec/ci.sh coverage
+        run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec --security-opt seccomp=unconfined --env RUST_TOOLCHAIN_VERSION=1.66.0 ghcr.io/parallaxsecond/parsec-service-test-all /tmp/parsec/ci.sh coverage
       - name: Collect coverage results
         run: bash <(curl -s https://codecov.io/bash)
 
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -42,8 +42,21 @@ jobs:
...
+    runs-on: ubuntu-latest
+    if: ${{ always() }}
+    needs: [build-and-export-test-all-docker]
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run the container to execute the test script
+        uses: ./.github/actions/ci_script
+        with:
+          ci-flags: "cargo-check"
+          rs-version: "1.66.0"
```
minosgalanakis, line 56: Why not set this up as an user configurable input with 1.66.0 as the default version? That would allow custom jobs to test future versions or older versions.

.github/workflows/nightly.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         with:
           ref: "${{ github.event.inputs.rev }}"
       - name: Run the container to execute the coverage script
-        run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec --security-opt seccomp=unconfined ghcr.io/parallaxsecond/parsec-service-test-all /tmp/parsec/ci.sh coverage
+        run: docker run -v $(pwd):/tmp/parsec -w /tmp/parsec --security-opt seccomp=unconfined --env RUST_TOOLCHAIN_VERSION=1.66.0 ghcr.io/parallaxsecond/parsec-service-test-all /tmp/parsec/ci.sh coverage
```
minosgalanakis, line 55: As above, we could have this as an input/jenkins parameter instead of hardcoded

.github/workflows/ci.yml
tgonzalezorlandoarm: The default is and should be stable, as seen here: https://github.com/parallaxsecond/parsec/pull/737/files#diff-6b8e64fe03097f8751f9b85361f2037d351d24d2ed2994b3d23333ba4964e208R9.

So after this PR gets merged, every new PR will be tested with both stable and MSRV.

Is there any particular reason why we would want to add a new feature in the CI to allow the user to run with a specific compiler?

.github/workflows/nightly.yml
tgonzalezorlandoarm: Here we are fine with just running with MSRV, I don't think there's much benefit on being able to run it with a different compiler.
To avoid hardcoding certain parts of the command, we could use https://github.com/parallaxsecond/parsec/blob/main/.github/actions/ci_script/action.yml, and add an input for docker flags in there. This could be for a different PR as I don't think it's relevant for the topic of this one.

Cargo.lock
```diff
@@ -1800,11 +1794,10 @@ dependencies = [
 
 [[package]]
 name = "time"
-version = "0.3.29"
```
gowthamsk-arm, line 1803: Does it not build with `1.66`? Was the cargo check with MSRV not detecting it?

ci.sh
```diff
@@ -360,12 +360,12 @@ if [ "$PROVIDER_NAME" = "cargo-check" ]; then
     # - openSUSE Tumbleweed
     # - openSUSE Leap 15.4
 
-    rustup toolchain install ${MSRV}
+    rustup default ${MSRV}
     # TODO: The "jwt-svid-authenticator" is currently not being used.
     RUST_BACKTRACE=1 cargo +${MSRV} check --release --features=all-providers,direct-authenticator,unix-peer-credentials-authenticator
```
gowthamsk-arm: We use `cargo +${MSRV} check ...` or `cargo +stable check ..` to consider particular toolchain. So do we need this?

gowthamsk-arm (approved):

minosgalanakis (approved):

Issue #738, opened 2023-11-22
e2e_tests/stress.rs: Signature Verification fails sporadically with PsaErrorInvalidArgument


tgonzalezorlandoarm:
stress tests are sporadically failing:

```
[ERROR parsec_service::providers::mbed_crypto::asym_sign] Verify status: ; Error: the signature, MAC or hash is incorrect
test per_provider::stress_test::stress_test ... FAILED

failures:

---- per_provider::stress_test::stress_test stdout ----
thread '<unnamed>' panicked at src/stress.rs:236:21:
An invalid signature or a tampering detection should be the only reasons of the verification failing. Status returned: PsaErrorInvalidArgument.
stack backtrace:
   0: rust_begin_unwind
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:595:5
   1: core::panicking::panic_fmt
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:67:14
   2: e2e_tests::stress::StressTestWorker::execute_request
             at ./src/stress.rs:236:21
   3: e2e_tests::stress::StressTestWorker::run_test
             at ./src/stress.rs:181:13
   4: e2e_tests::stress::StressClient::execute::{{closure}}
             at ./src/stress.rs:90:17
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
thread 'per_provider::stress_test::stress_test' panicked at src/stress.rs:101:27:
Test thread panicked: Any { .. }
stack backtrace:
   0: rust_begin_unwind
             at /rustc/cc66ad[4689](https://github.com/tgonzalezorlandoarm/parsec/actions/runs/6510795985/job/17685225654#step:4:4690)55717ab92600c770da8c1601a4ff33/library/std/src/panicking.rs:595:5
   1: core::panicking::panic_fmt
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/panicking.rs:67:14
   2: core::result::unwrap_failed
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/result.rs:1652:5
   3: core::result::Result<T,E>::expect
             at /rustc/cc66ad468955717ab92600c770da8c1601a4ff33/library/core/src/result.rs:1034:23
   4: e2e_tests::stress::StressClient::execute
             at ./src/stress.rs:101:13
[...]

PR #739, opened 2023-11-22
e2e_tests/stress.rs: Add a workaround for spurious test failures

tgonzalezorlandoarm:
stress_tests fail sporadically with the error:

 'An invalid signature or a tampering detection should be the only
  reasons of the verification failing. Status returned:
  PsaErrorInvalidArgument.'

The verification signature should not be failing with PsaErrorInvalidArgument for invalid signatures. So there may be a problem with either threading or a bug in Mbed Crypto. This bug is causing several sporadic test failures in our CI, as the invalid signatures are random.

 * Accept the PsaErrorInvalidArgument error as a workaround until the issue is resolved.

Please see opened issue #738

### DIFF
--- a/e2e_tests/src/stress.rs
+++ b/e2e_tests/src/stress.rs
@@ -231,6 +231,7 @@ impl StressTestWorker {
                     )
                     .expect_err("Verification should fail.");
                 if !(status == ResponseStatus::PsaErrorInvalidSignature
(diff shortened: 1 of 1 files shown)


gowthamsk-arm (approved):
Can you create an issue and link it to this one please?

PR #746, opened 2023-12-05
.cargo/config.toml: remove

billatarm:
When compiling on an aarch64 host, the override of the linker to the cross "compiler" (uses the gcc frontend) causes a build failure. Theirs no reason to set these globally for all developers as their build host may be different as well as their toolchain preference and naming conventions.

Instead override the target specific linker variable in a --config option string in the cross compilation tests.


billatarm:
This hasn't been tested yet, I didn't look into firing up the CI. But it looks like the `.cargo/config.toml` file was only needed for the cross compilation CI tests and can be nuked.

This fixes building on an aarch64 host.

billatarm:
Ahh OK, I looked through the docker hub testing locally now will upload a fix.

billatarm:
> Ahh OK, I looked through the docker hub testing locally now will upload a fix.

Had the key wrong and some minor TOML formatting issues. I have working locally now, lets see what the CI thinks.

gowthamsk-arm (approved):
LGTM. Thank you :)

PR #747, opened 2024-01-03
Use infallible conversion into instead of try_into

gowthamsk-arm:
A new clippy lint https://rust-lang.github.io/rust-clippy/master/index.html#/unnecessary_fallible_conversions has been added to rust 1.75.0 that is causing the CI failure. The use of try_into when into is available will be reported by the lint now.

### DIFF
--- a/src/providers/trusted_service/context/asym_encryption.rs
+++ b/src/providers/trusted_service/context/asym_encryption.rs
@@ -6,22 +6,16 @@ use super::ts_protobuf::{
 use super::Context;
 use parsec_interface::operations::psa_algorithm::AsymmetricEncryption;
 use parsec_interface::requests::ResponseStatus;
-use std::convert::TryInto;
-use zeroize::Zeroize;
 
 impl Context {
     pub fn asym_encrypt(
         &self,
         key_id: u32,
         alg: AsymmetricEncryption,
-        mut plaintext: Vec<u8>,
-        mut salt: Vec<u8>,
+        plaintext: Vec<u8>,
+        salt: Vec<u8>,
     ) -> Result<Vec<u8>, ResponseStatus> {
-        let alg = alg.try_into().map_err(|e| {
-            plaintext.zeroize();
-            salt.zeroize();
-            e
-        })?;
+        let alg = alg.into();
         let req = AsymmetricEncryptIn {
             id: key_id,
             alg,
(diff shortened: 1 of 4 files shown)


gowthamsk-arm:
Nightly CI failure: https://github.com/parallaxsecond/parsec/actions/runs/7391951129

src/providers/trusted_service/context/asym_encryption.rs
```diff
@@ -6,22 +6,16 @@ use super::ts_protobuf::{
...
 use parsec_interface::requests::ResponseStatus;
-use std::convert::TryInto;
-use zeroize::Zeroize;
 
 impl Context {
     pub fn asym_encrypt(
         &self,
         key_id: u32,
         alg: AsymmetricEncryption,
-        mut plaintext: Vec<u8>,
```
minosgalanakis, line 17: I am assuming zeroisation is no longer required?

minosgalanakis (approved):
LGTM

src/providers/trusted_service/context/asym_encryption.rs
ionut-arm: Hmm, what I find interesting is that we didn't zeroize it after sending the request, even though we don't pass ownership of the plaintext data, just a reference to it. Maybe we should do it before returning, regardless of whether it's an exception or not. Or maybe we didn't do it for some reason and I forgot the reason 🤔

Would it be possible for you to quickly check if zeroizing at the end breaks anything?

gowthamsk-arm: I thought all the structs from ts_protobuf have Drop trail implemented [here](https://github.com/parallaxsecond/parsec/blob/main/src/providers/trusted_service/context/ts_protobuf.rs#L62). But I can only see Drop being implemented for `ImportKeyIn`, `ExportKeyOut`, `SignHashIn`, `VerifyHashIn`. For these structs the drop method takes care of the zeroise part. 

Not sure why `AsymmetricEncryptIn` and `AsymmetricDecryptIn` don't have them.

gowthamsk-arm: The test [patch](https://github.com/parallaxsecond/parsec/pull/747/commits/87b6e7206ab9212aaaa287ff72ee82191ce1a489) seems to work fine and there are no failures reported in the CI jobs for trusted services backend.

tgonzalezorlandoarm: I would be more inclined for the Drop method of the involved types to handle the zeroization. Could you check if this is possible and easy to do? I'm approving in case it's not possible/easy.

tgonzalezorlandoarm (dismissed):
One check to do, but LGTM other than that

gowthamsk-arm:
CI run failure for the Cargo deny job is expected and @tgonzalezorlandoarm has a fix for it.

tgonzalezorlandoarm (approved):
LGTM

ionut-arm (approved):
👍🏻

PR #750, opened 2024-03-14
Cargo.toml: Bump tss-esapi to 7.5.0

tgonzalezorlandoarm:
The new version of tss-esapi contains a required security fix.

### DIFF
--- a/src/utils/cli.rs
+++ b/src/utils/cli.rs
@@ -5,6 +5,7 @@
 // WARNING: This file should be only updated in a non-breaking way. CLI flags should not be
 // removed, new flags should be tested.
 // See https://github.com/parallaxsecond/parsec/issues/392 for details.
+#![allow(unused_qualifications)]
 
 use structopt::StructOpt;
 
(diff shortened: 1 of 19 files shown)


src/utils/cli.rs
```diff
@@ -5,6 +5,7 @@
 // WARNING: This file should be only updated in a non-breaking way. CLI flags should not be
 // removed, new flags should be tested.
 // See https://github.com/parallaxsecond/parsec/issues/392 for details.
+#![allow(unused_qualifications)]
```
gowthamsk-arm, line 8: Why allow here?

tgonzalezorlandoarm: I'm getting this error here. From what I understand, this is getting mixed up with unused_qualifications in the definition of structopt, so I disabled it as it's a very small file. Please let me know if this is incorrect and I should change it.

tgonzalezorlandoarm: https://github.com/parallaxsecond/parsec/actions/runs/8283203581/job/22665825642#step:5:1281

PR #751, opened 2024-03-14
TPM Provider: Check root key's name

tgonzalezorlandoarm:
Check the name of the TPM's root key, comparing it with a previously stored version


gowthamsk-arm (approved):
LGTM. Thanks:)

PR #755, opened 2024-03-28
tpm/tests: Ignore test_root_key_check case

tgonzalezorlandoarm:
Current situation:

 1. The test depends on a TPM being available in the system.
 2. The connection settings to the TPM have to be changed to be in line with the TPM available in the system. Currently, the hardcoded settings are set to test with the software TPM enabled by the CI.

Therefore:

 * Ignore the test by default. Each developer can run their tests according to the TPM configuration if it's needed.
 * Explicitly run the test in CI to make sure this is tested at least with the software TPM.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -396,6 +396,11 @@ fi
 echo "Unit, doc and integration tests"
 RUST_BACKTRACE=1 cargo test $FEATURES
 
+# Ignored unit tests that we want to manually run should be run here!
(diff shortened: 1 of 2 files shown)


tgonzalezorlandoarm:
https://github.com/parallaxsecond/parsec/actions/runs/8468506178/job/23201701003#step:3:1080

gowthamsk-arm (approved):
LGTM

PR #778, opened 2025-02-23
ci: update gh actions

reubenmiller:
Update github actions to use latest versions and switch to a maintained rust toolchain action (`actions-rs/toolchain` => `actions-rust-lang/setup-rust-toolchain`)

### DIFF
--- a/.github/actions/build_export_docker/action.yml
+++ b/.github/actions/build_export_docker/action.yml
@@ -15,7 +15,7 @@ runs:
       run: docker save ${{ inputs.image-name }} > /tmp/${{ inputs.image-name }}.tar
       shell: bash
     - name: Upload artifact
-      uses: actions/upload-artifact@v3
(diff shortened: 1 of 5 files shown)


reubenmiller:
Note: I switch from my previously suggested `dtolnay/rust-toolchain` to `actions-rust-lang/setup-rust-toolchain` after looking at the other parallaxsecond repos which already make use of `actions-rust-lang/setup-rust-toolchain`.

hug-dev (approved):
Thanks!

reubenmiller:
> Thanks!

Though there are a few dependency issues which will make the multiple PRs that I created difficult to merge (as PR1 requires the Rust clippy fix, but that PR fails due to workflow dependency changes).

@hug-dev Let me know if you would like me to groups this PRs or not, of if you are happy to force merge one of the block PRs to get paste this situation.

PR #780, opened 2025-02-23
chore: fix pr check issues

reubenmiller:
Fix various issues with current Rust pipeline. The different changes are still split in different commits, however all of the commits related to making the PR checks pass again (which should make it also easier to merge the PR without having to bypass any checks).

### Fix clippy error on Rust stable.

```sh
error: useless conversion to the same type: `char`
  --> src/stress.rs:61:20
   |
61 |            .take(size)
   |   ____________________^
   |  |____________________|
62 | ||         .map(char::from)
   | ||________________________^
63 | |          .collect()
   | |_________- help: consider removing
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
   = note: `-D clippy::useless-conversion` implied by `-D clippy::all`
   = help: to override `-D clippy::all` add `#[allow(clippy::useless_conversion)]`
```

### ~Update spiffe dependency~

~The spiffe dependency was outdated and causing an error in the cargo-deny.~

### cargo-deny

cargo-deny has gone through some changes and many of the properties have been deprecated. Comments were removed from the deny.toml as the comments were introducing too much noise which made the file harder to read (imho).  Two new advisories were added to the ignore list (with reasons) as one was due to the old clap version (due to current MSRV), and one due to using the unmaintained derivation crate. The latter will require a decision before a replacement can be applied to all related parsec projects.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,107 +1,29 @@
-# This template contains all of the possible sections and their default values
-
-# Note that all fields that take a lint level have these possible values:
-# * deny - An error will be produced and the check will fail
-# * warn - A warning will be produced, but the check will not fail
-# * allow - No warning or error will be produced, though in some cases a note
-# will be
-
-# The values provided in this template are the default values that will be used
-# when any section or field is not specified in your own configuration
-
-# Root options
-
-# If 1 or more target triples (and optionally, target_features) are specified,
-# only the specified targets will be checked when running `cargo deny check`.
-# This means, if a particular package is only ever used as a target specific
-# dependency, such as, for example, the `nix` crate only being used via the
-# `target_family = "unix"` configuration, that only having windows targets in
-# this list would mean the nix crate, as well as any of its exclusive
-# dependencies not shared by any other crates, would be ignored, as the target
-# list here is effectively saying which targets you are building for.
-targets = [
-    # The triple can be any string, but only the target triples built in to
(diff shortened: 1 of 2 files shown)


hug-dev (dismissed):
Thanks a lot for your fixes here and in other repos!

reubenmiller:
@hug-dev I've extended the scope of this PR to also include some of the additional changes to allow the PR checks to be able to pass. There were multiple issues which were blocking this (cargo-deny config, out of date libraries which caused cargo-deny to fail, clippy errors).

Once this is merged, then the PRs should be a lot slimmer.

hug-dev (dismissed):

reubenmiller:
Ok, I was too optimistic. Seems there are some changes due to the spiffe upgrade. I'll have to spend some time getting a use-able dev setup so I can test this locally before pushing a fix.

reubenmiller:
After looking into spiffe a bit deeper and reading the [CHANGELOG.md](https://github.com/maxlambrecht/rust-spiffe/blob/main/spiffe/CHANGELOG.md) it seems that it is not going to be a simple upgrade as the from 0.3.x it requires tokio and some other breaking changes.

Instead I've added the spiffe errors to the cargo-deny config to ignore them for now until a decision can be made on how to move forward.

